PR #4, opened 2020-09-23
Add validation for rancher rbac objects

dramich:
Problem:
Adding the role restricted-admin in rancher requires validation of
objects through diferent means to validate the user doing the action

Solution:
Add webhook validation to GRBs PRTBs and CRTBs.
GRBs: Validation is done to ensure the GRB being created by the user will
not cause escalated privileges so the user creating the GRB must have at
minimum the permissions they are attempting to grant through the GRB

PRTBs and CRTBs: The same check is done for both. First validation is
only done on the local cluster. Then a subject access review is done for
the requester checking that they have `**` access in the local cluster.

Issue: https://github.com/rancher/rancher/issues/28977


pkg/admission/cluster.go
```diff
@@ -23,6 +23,11 @@ type clusterValidator struct {
 }
 
 func (c *clusterValidator) Admit(response *webhook.Response, request *webhook.Request) error {
+	if request.Operation == admissionv1.Delete {
```
ibuildthecloud, line 26: This webhook should only be called for create/update.  Can we just define a second ValidatingWebhook instead of combining the two?

pkg/admission/clusterrtb.go
```diff
@@ -0,0 +1,83 @@
+package admission
```
ibuildthecloud, line 1: For code structure we might want to move the handlers to separate packages.  All the rbac ones can go in the same pkg, but it seems like we are just going to dump everthing in pkg/admission.  So maybe pkg/cluster and pkg/rbac

pkg/admission/cluster.go
dramich: Separated the two and left cluster as-is.

pkg/admission/clusterrtb.go
dramich: Done

pkg/auth/globarolebinding.go
```diff
@@ -0,0 +1,119 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apiserver/pkg/authentication/user"
+	k8srequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubernetes/pkg/registry/rbac/validation"
+	rbacregistryvalidation "k8s.io/kubernetes/pkg/registry/rbac/validation"
+	"k8s.io/utils/trace"
+)
+
+func NewGRBValidator(cfg *rest.Config) (webhook.Handler, error) {
```
ibuildthecloud, line 28: nit: it would be cleaner to not pass in rest.Config but instead the objects we need from it

pkg/auth/globarolebinding.go
```diff
@@ -0,0 +1,119 @@
...
+	listTrace := trace.New("globalRoleBindingValidator Admit", trace.Field{Key: "user", Value: request.UserInfo.Username})
+	defer listTrace.LogIfLong(1 * time.Second)
+
+	newGRB, err := grbObject(request)
+	if err != nil {
+		return err
+	}
+
+	// Pull the global role to get the rules
+	globalRole, err := grbv.globalRoleClient.Get(newGRB.GlobalRoleName, metav1.GetOptions{})
```
ibuildthecloud, line 67: Do we need to use a cache? You know better than me how many times this might be called.  Are we creating a lot of grbs?

pkg/auth/globarolebinding.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	userInfo := &user.DefaultInfo{
+		Name:   request.UserInfo.Username,
+		UID:    request.UserInfo.UID,
+		Groups: request.UserInfo.Groups,
+		Extra:  toExtraString(request.UserInfo.Extra),
+	}
+
+	if err := grbv.ConfirmNoEscalation(globalRole.Rules, userInfo); err != nil {
```
ibuildthecloud, line 79: Well that was simple :)

pkg/auth/projectrtb.go
```diff
@@ -0,0 +1,88 @@
...
+
+	prtb, err := prtbObject(request)
+	if err != nil {
+		return err
+	}
+
+	clusterID := clusterFromProject(prtb.ProjectName)
+
+	if clusterID != "local" {
+		response.Allowed = true
```
ibuildthecloud, line 40: Do you need to return here?

pkg/auth/clusterrtb.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return err
+	}
+
+	if crtb.ClusterName != "local" {
+		response.Allowed = true
+		return nil
+	}
+
+	resp, err := c.sar.Create(request.Context, &v1.SubjectAccessReview{
```
ibuildthecloud, line 44: Should crtb and prtb use a common method for this?

pkg/auth/globarolebinding.go
dramich: Per offline discussion, this only applies around user creation and wouldn't affect cluster/project creation so leaving it as a live call

PR #7, opened 2020-09-30
Treat PRTB and CRTBs as rancher currently does

dramich:


pkg/auth/projectrtb.go
```diff
@@ -52,7 +69,7 @@ func prtbObject(request *webhook.Request) (*rancherv3.ProjectRoleTemplateBinding
 	return prtb.(*rancherv3.ProjectRoleTemplateBinding), err
 }
 
-func clusterFromProject(project string) string {
+func clusterFromProject(project string) (string, string) {
 	pieces := strings.Split(project, ":")
-	return pieces[0]
+	return pieces[0], pieces[1]
```
ibuildthecloud, line 74: There's a util function in wrangler    `first, second := kv.Split(str, ":")` to do this safely. FYI.

PR #14, opened 2020-11-13
Add fleet mutating webhook

StrongMonkey:


pkg/server/server.go
```diff
@@ -161,9 +167,50 @@ func listenAndServe(ctx context.Context, clients *clients.Clients, handler http.
...
+							Rule: v1.Rule{
+								APIGroups:   []string{"management.cattle.io"},
+								APIVersions: []string{"v3"},
+								Resources:   []string{"fleetworkspaces"},
+								Scope:       &clusterScope,
+							},
+						},
+					},
+					FailurePolicy:           &failPolicyFail,
+					SideEffects:             &sideEffectClassNone,
```
dramich, line 200: This does have side effects, so needs to be SideEffectClassNoneOnDryRun and then don't create anything on dryrun

StrongMonkey: Dryrun has properly been handled in here. https://github.com/rancher/webhook/pull/14/commits/b57e47194500ef13a3037afcca76896467678836#diff-0c5685479c4cd51ebba475dde0ae84021c2f1f333619a544c5c0d6eddf75a558R44-R46

StrongMonkey:
https://github.com/rancher/fleet/issues/140

pkg/server/server.go
dramich: Still need to update this field to `SideEffectClassNoneOnDryRun`

StrongMonkey: Changed

pkg/resources/fleetworkspace/fleetworkspace.go
```diff
@@ -0,0 +1,183 @@
...
+			return err
+		}
+		ns, err = m.namespaces.Get(namespace.Name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
+	// create rolebinding to bind current user with fleetworkspace-admin role in current namespace
+	if err := m.createAdminRoleAndBindings(request, fw); err != nil {
```
ibuildthecloud, line 79: What if these objects already exist?

PR #16, opened 2020-12-14
Fix wrong apiVersion and group for fleetworkspace

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/30452

StrongMonkey:
Turns out that we are checking the wrong api version and group for fleetworkspace

dramich (approved):

PR #20, opened 2021-04-19
Bump go, helm and golangci versions

dramich:


Dockerfile.dapper
```diff
@@ -1,14 +1,14 @@
-FROM golang:1.14.7
+FROM golang:1.16
```
StrongMonkey, line 1: Should we pin to 1.16.3?

dramich: Yeah, probably, I was following Darren here https://github.com/rancher/rancher/blob/master/Dockerfile.dapper#L28 😄

StrongMonkey (approved):

PR #21, opened 2021-04-19
[0.1] Bump go, helm and golangci versions

dramich:


Dockerfile.dapper
```diff
@@ -1,14 +1,14 @@
-FROM golang:1.14.7
+FROM golang:1.16
```
StrongMonkey, line 1: Same, 1.16.3?

StrongMonkey (approved):

PR #35, opened 2021-07-01
[DNM|release/v0.1] Add ability to turn on debug logging, add access logs

ryansann:
No issue for this. Proposing this as an improvement for debugging purposes. Setting the either env var will turn on the access logs


ryansann:
Closed this in favor of putting this logic directly in rancher/wrangler

PR #36, opened 2021-07-01
Add mutation and validation for provisioning clusters

dramich:
Add the creatorID annotation onto provisioning.cattle.io clusters based on the user who created the object. Add validation to enforce that the only change to this field is that it can be deleted. It can't be updated after creation of the cluster object.

A large part of this change is moving resources under their respective webhook type so an object can have both without putting it into the same file.


ryansann (approved):
This looks good. We're enforcing that the cluster object gets a creatorId annotation on Create and that annotation is never allowed to be modified, only deleted. Along with some nice cleanups. Approved

kinarashah (approved):

PR #37, opened 2021-07-02
[release/v0.1] Fix for panic and to allow GRB deletion

ryansann:
Issue:
- https://github.com/rancher/rancher/issues/33429

These changes pull in a new wrangler version which has a fix for a panic in the webhook handler logic. Also included is allowance for update & delete of GRBs if the corresponding GR isn't found, i.e. it was deleted. Updates to GRBs are only allowed if the DeletionTimestamp != nil, this only allows updates to finalizers when there is no associated GR

### DIFF
--- a/pkg/resources/globalrolebinding/globarolebinding.go
+++ b/pkg/resources/globalrolebinding/globarolebinding.go
@@ -52,16 +52,15 @@ func (grbv *globalRoleBindingValidator) Admit(response *webhook.Response, reques
 				response.Allowed = true
 				return nil
 			}
-			fallthrough
-		default: // other operations not allowed when GR is not found
-			response.Result = &metav1.Status{
-				Status:  "Failure",
-				Message: fmt.Sprintf("referenced globalRole %s not found, only deletions allowed", newGRB.Name),
-				Reason:  metav1.StatusReasonUnauthorized,
-				Code:    http.StatusUnauthorized,
-			}
-			return nil
 		}
+		// other operations not allowed
+		response.Result = &metav1.Status{
+			Status:  "Failure",
+			Message: fmt.Sprintf("referenced globalRole %s not found, only deletions allowed", newGRB.Name),
+			Reason:  metav1.StatusReasonUnauthorized,
+			Code:    http.StatusUnauthorized,
+		}
+		return nil
 	}
 
 	return grbv.escalationChecker.ConfirmNoEscalation(response, request, globalRole.Rules, "")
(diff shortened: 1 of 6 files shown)


pkg/resources/globalrolebinding/globarolebinding.go
```diff
@@ -36,6 +40,27 @@ func (grbv *globalRoleBindingValidator) Admit(response *webhook.Response, reques
 	// Pull the global role to get the rules
 	globalRole, err := grbv.globalRoles.Get(newGRB.GlobalRoleName)
 	if err != nil {
+		if errors.IsNotFound(err) {
```
dramich, line 43: Can we do an opposite check here and then return if it isn't then the fallthrough is what exists inside this if check

pkg/resources/globalrolebinding/globarolebinding.go
```diff
@@ -36,6 +40,27 @@ func (grbv *globalRoleBindingValidator) Admit(response *webhook.Response, reques
...
+				return nil
+			// only allow updates to the finalizers if the GR associated to the GRB is not found
+			case admissionv1.Update:
+				if newGRB.DeletionTimestamp != nil {
+					response.Allowed = true
+					return nil
+				}
+				response.Result = &metav1.Status{
+					Status:  "Failure",
+					Message: fmt.Sprintf("GlobalRoleBinding %s associated GlobalRole not found. Only deletion and updates to finalizers are permitted.", newGRB.Name),
```
dramich, line 57: Errors should start with lowercase and this message is confusing. Can it be updated to something like `referenced globalRole %v not found, updates not allowed`. I don't think we need to call out the finalizer stuff as that's an implementation detail of rancher and how we object lifecycles and probably won't be useful to a user.

ryansann: Yes, agreed, I was trying to figure out a good way to say this. I'll change it to this

pkg/resources/globalrolebinding/globarolebinding.go
ryansann: Yep, I'll change

pkg/resources/globalrolebinding/globarolebinding.go
```diff
@@ -36,7 +40,28 @@ func (grbv *globalRoleBindingValidator) Admit(response *webhook.Response, reques
...
+		case admissionv1.Delete: // allow delete operations if the GR is not found
+			response.Allowed = true
+			return nil
+		case admissionv1.Update: // only allow updates to the finalizers if the GR is not found
+			if newGRB.DeletionTimestamp != nil {
+				response.Allowed = true
+				return nil
+			}
+			fallthrough
+		default: // other operations not allowed when GR is not found
```
dramich, line 56: These lines aren't needed since default and no match is the same operation

dramich (approved):

PR #38, opened 2021-07-02
Add ACE validation for v1.Clusters

thedadams:
When a v1.Cluster is created/updated, the ACE configuration should be
validated. The webhook will respond with an error if a cert is provided
without a FQDN.

Issue:
https://github.com/rancher/rancher/issues/32388

### DIFF
--- a/pkg/resources/validation/cluster/provisioningcluster.go
+++ b/pkg/resources/validation/cluster/provisioningcluster.go
@@ -6,10 +6,10 @@ import (
 
 	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	"github.com/rancher/webhook/pkg/auth"
+	objectsv1 "github.com/rancher/webhook/pkg/generated/objects/provisioning.cattle.io/v1"
 	"github.com/rancher/wrangler/pkg/webhook"
 	admissionv1 "k8s.io/api/admission/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -110,7 +110,7 @@ func listenAndServe(ctx context.Context, clients *clients.Clients, handler http.
 			Service: &v1.ServiceReference{
 				Namespace: namespace,
 				Name:      serviceName,
-				Path:      &validationPath,
+				Path:      &mutationPath,
```
dramich, line 113: @thedadams Can you open a new pr with just this change so we can get it merged ASAP

thedadams: Done; https://github.com/rancher/webhook/pull/39

nickgerace (approved):

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -22,63 +22,64 @@ type provisioningClusterValidator struct{}
...
+
+	if response.Result, err = validateACEConfig(cluster); err != nil || response.Result != nil {
+		return err
+	}
+
+	response.Allowed = true
+	return nil
+}
+
+func validateCreatorID(request *webhook.Request, oldCluster, cluster *v1.Cluster) (*metav1.Status, error) {
```
dramich, line 42: Neither function returns or has the ability to return an error so it doesn't need to be here

thedadams: Good catch!

dramich (approved):

PR #42, opened 2021-07-27
Add generation for getting objects from requests

thedadams:
In order to validate objects, the new and old objects, if they exist,
should be pull from the webhook request. This process is identical for
all objects except that the return types would be different. This is a
candidate for code generation, and this generation is added.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,5 +1,5 @@
 //go:generate go run pkg/codegen/cleanup/main.go
-//go:generate go run pkg/codegen/main.go
+//go:generate go run pkg/codegen/main.go pkg/codegen/template.go
 package main
 
 import (
--- a/pkg/codegen/template.go
+++ b/pkg/codegen/template.go
@@ -1,0 +1,61 @@
+package main
+
+const objectsFromRequestTemplate = `
+package {{ .package }}
+
+import (
+	{{ range .types }}
+	"{{ .Package }}"{{ end }}
+	"github.com/rancher/wrangler/pkg/webhook"
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+{{ range .types }}
+
+// {{ .Name }}OldAndNewFromRequest gets the old and new {{ .Name }} objects, respectively, from the webhook request.
+// If the request is a Delete operation, then the new object is the zero value for {{ .Name }}.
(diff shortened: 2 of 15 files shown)


pkg/codegen/template.go
```diff
@@ -0,0 +1,56 @@
...
+
+	oldObject, err := request.DecodeOldObject()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return oldObject.({{ .Type }}), object.({{ .Type }}), nil
+}
+
+func {{ .Name }}ObjectFromRequest(request *webhook.Request) ({{ .Type }}, error) {
```
dramich, line 40:
```suggestion
func {{ .Name }}FromRequest(request *webhook.Request) ({{ .Type }}, error) {
```

pkg/codegen/template.go
```diff
@@ -0,0 +1,56 @@
...
+	{{ range .types }}
+	"{{ .Package }}"{{ end }}
+	"github.com/rancher/wrangler/pkg/webhook"
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+{{ range .types }}
+
+func {{ .Name }}OldAndNewFromRequest(request *webhook.Request) ({{ .Type }}, {{ .Type }}, error) {
```
dramich, line 16: Add a comment above each function to describe what it does

main.go
```diff
@@ -1,5 +1,5 @@
 //go:generate go run pkg/codegen/cleanup/main.go
-//go:generate go run pkg/codegen/main.go
+//go:generate go run pkg/codegen/main.go pkg/codegen/template.go
```
dramich, line 2: Is this change needed? that file is part of the main package and doesn't run any code

thedadams: Yes, it is required: `go run` doesn't look at files not passed as arguments when compiling. Without this change, the template isn't found because it is in `pkg/codegen/template.go`.

pkg/codegen/template.go
thedadams: Done.

pkg/codegen/template.go
thedadams: Done.

ryansann (approved):
LGTM, nice cleanup in the handlers

dramich (approved):

PR #44, opened 2021-08-09
Check provisioning cluster for cloudCred access

dramich:
Problem:
Any cloud cred can be used without permissions to view that secret

Solution:
Validate user creating or modifying the provisioning cluster has get
access to the referenced cloudCredential. When modifying the cluster,
permission is only checked if the cloudCred changes from existing. The
users permissions on the cluster say if they can use the existing
cloudCred.
This supports both old and new style cloudCred naming and namespace


ryansann (approved):
- Validating v1.Cluster on Create & Update
- Performing a SAR if the cloud cred is non-empty and was changed 
- Handling both cred naming strategies
- Not allowing if SAR failed

This LGTM

PR #48, opened 2021-08-21
Feature validation

rmweir:


pkg/generated/controllers/management.cattle.io/v3/feature.go
```diff
@@ -0,0 +1,376 @@
...
+		return nil, err
+	}
+	if !exists {
+		return nil, errors.NewNotFound(c.resource, name)
+	}
+	return obj.(*v3.Feature), nil
+}
+
+func (c *featureCache) List(selector labels.Selector) (ret []*v3.Feature, err error) {
+
```
brendarearden, line 238: nit: this is the only one with a space after the func declaration

rmweir: This file is generated, tried to undo and `go generate` just reverted. The other comparable files have the same at the same line.

brendarearden: I did miss that it was in the generated directory - my bad!

brendarearden (approved):
LGTM

pkg/resources/validation/feature/feature.go
```diff
@@ -0,0 +1,63 @@
...
+	return &featureValidator{}
+}
+
+type featureValidator struct{}
+
+func (fv *featureValidator) Admit(response *webhook.Response, request *webhook.Request) error {
+	listTrace := trace.New("featureValidator Admit", trace.Field{Key: "user", Value: request.UserInfo.Username})
+	defer listTrace.LogIfLong(2 * time.Second)
+
+	if request.Operation == admissionv1.Create {
```
dramich, line 25: How do we add new features?

pkg/resources/validation/feature/feature.go
```diff
@@ -0,0 +1,63 @@
...
+		response.Result = &metav1.Status{
+			Status:  "Failure",
+			Message: "Cannot create new Features",
+			Reason:  metav1.StatusReasonMethodNotAllowed,
+			Code:    http.StatusMethodNotAllowed,
+		}
+		response.Allowed = false
+		return nil
+	}
+	if request.Operation == admissionv1.Delete {
```
dramich, line 35: How do we cleanup features in the future?

dramich (commented):
This seems oddly restrictive and I don't understand why we are trying to do this. Blocking all creates and deletes will cause issues in the future for Rancher. We don't provide permissions for doing creates or deletes in any role currently so you are attempting to block an admin from doing something which is pointless.

pkg/generated/controllers/management.cattle.io/v3/feature.go
dramich: @thedadams ^

pkg/resources/validation/feature/feature.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		response.Allowed = false
+		return nil
+	}
+
+	newFeature, oldFeature, err := objectsv3.FeatureOldAndNewFromRequest(request)
+	if err != nil {
+		return err
+	}
+	if oldFeature.Status.LockedValue != nil {
```
dramich, line 50: This check is incomplete, what if the update is to the annotations? Or description? Saying 'No updates every' because the value is locked is a bad policy.

rmweir:
@dramich I think I was too focused on matching the current validation for feature and didn't consider how the webhook affects all requests. Removed create/delete checks and narrowed scope of value/lockedValue validation to prevent just the specific case that matters.

pkg/server/server.go
```diff
@@ -207,6 +207,19 @@ func listenAndServe(ctx context.Context, clients *clients.Clients, handler http.
 								Scope:       &clusterScope,
 							},
 						},
+						{
+							Operations: []v1.OperationType{
+								v1.Create,
+								v1.Update,
+								v1.Delete,
```
dramich, line 214: This should be updated to just update

PR #52, opened 2021-09-11
Validate provisioning cluster names on create

thedadams:
Users should not be able to create a cluster with name 'local' or of the
form 'c-xxxxx' because this conflicts with logic related to v3.Clusters.
The webhook now checks for such names and denies creates.

Issue:
https://github.com/rancher/rancher/issues/33694

### DIFF
--- a/pkg/resources/validation/cluster/provisioningcluster.go
+++ b/pkg/resources/validation/cluster/provisioningcluster.go
@@ -2,15 +2,19 @@ package cluster
 
 import (
 	"net/http"
+	"regexp"
 	"time"
 
 	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	"github.com/rancher/webhook/pkg/auth"
+	"github.com/rancher/webhook/pkg/clients"
+	v3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
 	objectsv1 "github.com/rancher/webhook/pkg/generated/objects/provisioning.cattle.io/v1"
 	"github.com/rancher/wrangler/pkg/kv"
 	"github.com/rancher/wrangler/pkg/webhook"
 	admissionv1 "k8s.io/api/admission/v1"
 	authv1 "k8s.io/api/authorization/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	authorizationv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/utils/trace"
@@ -18,14 +22,18 @@ import (
 
 const globalNamespace = "cattle-global-data"
 
-func NewProvisioningClusterValidator(sar authorizationv1.SubjectAccessReviewInterface) webhook.Handler {
+var mgmtNameRegex = regexp.MustCompile("^c-[a-z0-9]{5}$")
+
+func NewProvisioningClusterValidator(client *clients.Clients) webhook.Handler {
 	return &provisioningClusterValidator{
-		sar: sar,
+		sar:               client.K8s.AuthorizationV1().SubjectAccessReviews(),
+		mgmtClusterClient: client.Management.Cluster(),
 	}
 }
 
 type provisioningClusterValidator struct {
-	sar authorizationv1.SubjectAccessReviewInterface
+	sar               authorizationv1.SubjectAccessReviewInterface
+	mgmtClusterClient v3.ClusterClient
 }
 
 func (p *provisioningClusterValidator) Admit(response *webhook.Response, request *webhook.Request) error {
@@ -33,6 +41,10 @@ func (p *provisioningClusterValidator) Admit(response *webhook.Response, request
 	defer listTrace.LogIfLong(2 * time.Second)
 	oldCluster, cluster, err := objectsv1.ClusterOldAndNewFromRequest(request)
 	if err != nil {
+		return err
+	}
+
+	if err := p.validateClusterName(request, response, cluster); err != nil || response.Result != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -1,21 +1,32 @@
 package cluster
 
 import (
+	"net/http"
+	"regexp"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
dramich, line 8: Format imports

dramich (changes requested):
This needs to be moved to a validation webhook, not mutation.

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -135,6 +147,36 @@ func validateCreatorID(request *webhook.Request, oldCluster, cluster *v1.Cluster
 	return nil
 }
 
+func (p *provisioningClusterValidator) validateClusterName(request *webhook.Request, response *webhook.Response, cluster *v1.Cluster) error {
+	if request.Operation != admissionv1.Create {
+		return nil
+	}
+
+	_, err := p.mgmtClusterClient.Get(cluster.Name, metav1.GetOptions{})
```
snasovich, line 155: Do we really need to call this `Get` when we're not using its response?
Is this validation really concerned with whether the cluster with such a name exists?

thedadams: We want to know if a v3.Cluster with such a name exists. If it does, then creating the provisioning cluster with this name is valid (the controller is creating it as expected). If no v3.Cluster exists, then the user is trying to create a provisioning cluster with an invalid name and we wish to block it.

snasovich: Thank you for the quick explanation response, I though I was missing some important detail here.

pkg/server/validation.go
```diff
@@ -31,7 +31,7 @@ func Validation(clients *clients.Clients) (http.Handler, error) {
 		prtbs := projectroletemplatebinding.NewValidator(clients.Management.RoleTemplate().Cache(), clients.EscalationChecker)
 		crtbs := clusterroletemplatebinding.NewValidator(clients.Management.RoleTemplate().Cache(), clients.EscalationChecker)
 		roleTemplates := roletemplate.NewValidator(clients.EscalationChecker)
-		provisioningCluster := cluster.NewProvisioningClusterValidator(clients.K8s.AuthorizationV1().SubjectAccessReviews())
+		provisioningCluster := cluster.NewProvisioningClusterValidator(clients)
```
dramich, line 34: Do we only create the provisioning cluster if MCM is enabled?

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -135,6 +147,36 @@ func validateCreatorID(request *webhook.Request, oldCluster, cluster *v1.Cluster
 	return nil
 }
 
+func (p *provisioningClusterValidator) validateClusterName(request *webhook.Request, response *webhook.Response, cluster *v1.Cluster) error {
```
snasovich, line 150: So `cluster` is coming from `request` so ideally we want to either pass just `request` and have func extract what it needs or pass individual pieces it needs from `request`. In this case, I would probably go with latter as the only piece we need from request is operation.
The rationale here is that current signature allows passing `request` that's for cluster that is different from passed `cluster`.
I see other funcs here do the same, so I'm not hell-bent on having this fixed.

snasovich (dismissed):
LGTM from general code style/logic perspective; one minor comment

pkg/server/validation.go
thedadams: According to what I am seeing in the rancher/rancher code: no, the provisioning clusters are created for v3.Clusters regardless of whether MCM is enabled.

dramich: So this would not work if MCM is disabled

thedadams: Could you explain why so that I might find another solution?

dramich: Yeah, this is inside a `if clients.MultiClusterManagement {` check, so if mcm is disabled we don't register these handlers.

thedadams: Is the solution as simple as moving the relevant code outside the if-block?

rmweir: Yes.

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -135,6 +147,36 @@ func validateCreatorID(request *webhook.Request, oldCluster, cluster *v1.Cluster
...
+	if request.Operation != admissionv1.Create {
+		return nil
+	}
+
+	_, err := p.mgmtClusterClient.Get(cluster.Name, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
```
rmweir, line 158: Wouldn't it be an issue if the cluster name is already in use? Or are you just going to let kubernetes return that error?

rmweir (changes requested):

pkg/resources/validation/cluster/provisioningcluster.go
thedadams: Rancher will create a provisioning cluster for every management cluster. That is, if a user creates an RKE1 cluster, for example, then the Rancher controller will create a provisioning cluster of the same name. We want to allow this.

pkg/resources/validation/cluster/provisioningcluster.go
thedadams: I understand where you are coming from, but I would like to keep the code consistent for now.

pkg/server/validation.go
thedadams: I tested the webhook with moving this outside of the if-block and it works as expected when MCM is disabled.

pkg/resources/validation/cluster/provisioningcluster.go
rmweir: Oh I see, bc this validation runs on client requests to we have to allow it or the controller would not be able to do that.

rmweir: Then what is this get checking for exactly?

thedadams: If a management cluster exists with the same name as the provisioning cluster that is being created, then that is allowed because Rancher is doing that.

If no such management cluster exists, then the don't allow the provisioning cluster to be created because that means the user is trying to create a provisioning cluster with the "name format" of a management cluster. We don't want that to happen because there is Rancher logic that expects provisioning clusters with that name format to have come from an existing management cluster.

dramich: @thedadams Based on these comments can you include this as code comments, I suspect this will come up again in the future.

thedadams: I augmented my comment below. Please let me know if you think more comments are necessary.

dramich: Can you comment above the get call as well?

thedadams: Done.

snasovich (approved):

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -135,6 +147,36 @@ func validateCreatorID(request *webhook.Request, oldCluster, cluster *v1.Cluster
...
+		// Also, a provisioning cluster with name of the form "c-xxxxx" is expected to be created if a management cluster
+		// of the same name already exists because Rancher will create such a provisioning cluster.
+		// Any other situation should not be allowed.
+		response.Result = &metav1.Status{
+			Status:  "Failure",
+			Message: "cluster name cannot be \"local\" nor of the form \"c-xxxxx\"",
+			Reason:  metav1.StatusReasonInvalid,
+			Code:    http.StatusUnprocessableEntity,
+		}
+	}
```
rmweir, line 175: Do you think you could put this in its own function and write a few unit test case for it?

rmweir (changes requested):
Current code LGTM, I think some unit tests would be a good addition.

pkg/resources/validation/cluster/provisioningcluster.go
snasovich: @rmweir , generally I would be all for doing this, but on what is essentially the last day before "code freeze" I think it's not the best time to start enforcing more unit tests. :) @thedadams , unless you have some extra bandwidth to get it in.

thedadams: I can put some in tomorrow.

thedadams: Unit test has been added.

dramich (dismissed):

pkg/resources/validation/cluster/provisioningcluster.go
dramich: Thanks for adding the test, it looks good.

snasovich (approved):

rmweir (dismissed):
LGTM

dramich (approved):

PR #53, opened 2021-09-15
Omit rules from webhook configs if mcm is disabled

ryansann:
Issue: 
- https://github.com/rancher/rancher/issues/34691

When MCM is disabled we are not mounting RBAC resource validation routes in the webhook. This needs to also be reflected in the webhook configurations. As part of this I moved the rules out of the server.go file


dramich (approved):

rmweir (approved):
LGTM

PR #55, opened 2021-09-23
[2.5] add the annotation for setting the range of supported rancher version 

jiaqiluo:
`>= 2.5.10-0` since the post-delete-hook needs rancher 2.5.10 and above

### DIFF
--- a/charts/rancher-webhook/Chart.yaml
+++ b/charts/rancher-webhook/Chart.yaml
@@ -9,3 +9,4 @@ annotations:
   catalog.cattle.io/namespace: cattle-system
   catalog.cattle.io/release-name: rancher-webhook
   catalog.cattle.io/os: linux
+  catalog.cattle.io/rancher-version: '>= 2.5.10-0 <2.5.99-0'


StrongMonkey (approved):

charts/rancher-webhook/Chart.yaml
```diff
@@ -9,3 +9,4 @@ annotations:
   catalog.cattle.io/namespace: cattle-system
   catalog.cattle.io/release-name: rancher-webhook
   catalog.cattle.io/os: linux
+  catalog.cattle.io/rancher-version: '>= 2.5.10-0 <2.5.99-0'
```
thedadams, line 12: Based on our conversation on the other PR, I believe we want to remove the `-0` on `2.5.99`. That way, if we somehow get to `2.5.99`, then this will still get installed.

thedadams (changes requested):

thedadams (approved):

jiaqiluo:
@caleb Need your review on this PR because this PR will make the new chart version available on only rancher 2.5.10 and above, so a rancher v2.5.9 setup will never get new chart versions starring from this version. 

the reason for this lower boundary  
```
>= 2.5.10-0 since the newly-added post-delete-hook needs rancher 2.5.10 and above
```

jiaqiluo:
Closing this PR. 
there is no need to add the rancher-version annotation since the chart `rancher-webhook` is NOT user-facing.

PR #56, opened 2021-09-23
[2.6] add the annotation for setting the range of supported rancher version.

jiaqiluo:


charts/rancher-webhook/Chart.yaml
```diff
@@ -9,6 +9,7 @@ annotations:
   catalog.cattle.io/namespace: cattle-system
   catalog.cattle.io/release-name: rancher-webhook
   catalog.cattle.io/os: linux
+  catalog.cattle.io/rancher-version: '>2.6.0-0'
```
thedadams, line 12: Should this be `>2.6.0-alpha`?

thedadams (changes requested):

charts/rancher-webhook/Chart.yaml
jiaqiluo: '>2.6.0-0' is lower than `2.6.0-alpha`, and we decide to go with `2.6.0-0` for all charts. 
See the comparison at https://semvercompare.azurewebsites.net/?version=2.6.0-1&version=2.6.0-alpha&version=2.6.0-0&version=2.6.0-beta

thedadams (approved):

StrongMonkey (approved):

jiaqiluo:
Closing this PR. 
there is no need to add the rancher-version annotation since the chart `rancher-webhook` is NOT user-facing.

PR #59, opened 2021-11-03
Disallow disabling RKE2 feature if RKE2 clusters exist

thedadams:
Users should not be allowed to disable RKE2 feature flag if RKE2
provisioned clusters exist. When Rancher provisions an RKE2 cluster, an
annotation will be put on the RKE2 feature. If this annotation exists,
then changing the value of the RKE2 feature flag will not be allowed.

Issue:
https://github.com/rancher/rancher/issues/35179

### DIFF
--- a/pkg/resources/validation/feature/feature.go
+++ b/pkg/resources/validation/feature/feature.go
@@ -5,10 +5,16 @@ import (
 	"net/http"
 	"time"
 
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	objectsv3 "github.com/rancher/webhook/pkg/generated/objects/management.cattle.io/v3"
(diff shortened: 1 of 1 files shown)


cmurphy (commented):
I see something called `LockedValue` on the status of features, is that something that would be appropriate to use here in place of a new annotation? https://github.com/rancher/rancher/blob/release/v2.6/pkg/api/norman/customization/feature/feature.go#L42-L44

thedadams:
@cmurphy Yes, I think it would. If this is the case, then this PR can be closed. I will test that approach and close this PR and update the Rancher PR if it works.

Thanks for your suggestion!

thedadams:
Closing because Colleen's suggestion worked and this is no longer required.

PR #68, opened 2022-01-07
Update the versions of dynamiclistener and wrangler (by extension)

genexpr:
Related to https://github.com/rancher/rancher/issues/36079.

We should make a new tag for webhook - v0.2.4-rc3. This updates the latest webhook with the latest changes of dynamiclistener. Rancher v2.6.3's webhook is not using the latest dynamiclistener, it misses a commit that fixed the behavior of automatic cert update when a cert is close to expiration (https://github.com/rancher/dynamiclistener/commit/284cc004e84101e4a5b9fa50c00e101095db83f6).

Let's update the dynamiclistener for the newest webhook to be used in Rancher 2.6.4.

Wrangler is updated in the dynamiclistener, and so webhook must update it too. That's why there is a change in `main.go` to accommodate for the slight top-level Wrangler API change.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -812,8 +812,8 @@ github.com/rancher/aks-operator v1.0.1-rc11 h1:OGxwaQ+MmqTNl/45wfS0PuI78mvo/FXfB
 github.com/rancher/aks-operator v1.0.1-rc11/go.mod h1:V43rw+oKpjQ2tlIITgejlatpNBGi61DdB+RldJyrh/8=
 github.com/rancher/client-go v1.22.3-rancher.1 h1:aNVLaIY5YGah1i9wRVXLOGRbLyekohjQAKHXeQm6Cxo=
 github.com/rancher/client-go v1.22.3-rancher.1/go.mod h1:ElDjYf8gvZsKDYexmsmnMQ0DYO8W9RwBjfQ1PI53yow=
-github.com/rancher/dynamiclistener v0.3.1-0.20211119182849-962b63526960 h1:ahE9XbsyFzWteQB5etFl9OacXzVGtm0XJe1tHoOtC+4=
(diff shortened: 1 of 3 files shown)


cbron (dismissed):
lgtm, but wait to merge for @cmurphy or Ricky as they have more knowledge on wrangler/dl.

go.sum
```diff
@@ -812,8 +812,8 @@ github.com/rancher/aks-operator v1.0.1-rc11 h1:OGxwaQ+MmqTNl/45wfS0PuI78mvo/FXfB
 github.com/rancher/aks-operator v1.0.1-rc11/go.mod h1:V43rw+oKpjQ2tlIITgejlatpNBGi61DdB+RldJyrh/8=
 github.com/rancher/client-go v1.22.3-rancher.1 h1:aNVLaIY5YGah1i9wRVXLOGRbLyekohjQAKHXeQm6Cxo=
 github.com/rancher/client-go v1.22.3-rancher.1/go.mod h1:ElDjYf8gvZsKDYexmsmnMQ0DYO8W9RwBjfQ1PI53yow=
-github.com/rancher/dynamiclistener v0.3.1-0.20211119182849-962b63526960 h1:ahE9XbsyFzWteQB5etFl9OacXzVGtm0XJe1tHoOtC+4=
```
cbron, line 815: Should squash commits actually

genexpr: I will just let GitHub do it on merge. There is a Squash and Merge option on that button.

thedadams (approved):

go.sum
cmurphy: @genexpr it would be better to do it manually, that way reviewers can review the commit itself before it gets merged

cmurphy (changes requested):

cmurphy (approved):

cbron (approved):

PR #69, opened 2022-01-13
Adding the ability to set tolerations for the webhook deployment

MbolotSuse:


charts/rancher-webhook/templates/deployment.yaml
```diff
@@ -18,6 +18,10 @@ spec:
       {{- if .Values.global.hostNetwork }}
       hostNetwork: true
       {{- end }}
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
```
genexpr, line 23: Is the leading dash before `toYaml` needed?

MbolotSuse: It's not strictly necessary (chart still renders) but there's some extra whitespace that I'd rather not have. It's the difference between:

(w/o dash)
```
      tolerations:

        - effect: NoSchedule
          key: test1
          operator: Equals
          value: test1
```
and 

(with dash)
```
      tolerations:
        - effect: NoSchedule
          key: test1
          operator: Equals
          value: test1
```

MbolotSuse:
Part of rancher/rancher#34159 .

rmweir (approved):
LGTM

PR #73, opened 2022-01-18
[v2.6.3 patch1] Include changes of webhook certificate auto-renewal

maxsokolovsky:
Related to https://github.com/rancher/rancher/issues/36177.

Cherry-pick commits from tag `v0.2.3` until and including the dynamic listener update commit – to enable automatic cert renewal.
Once this PR is merged, I will tag the resulting latest commit as `v0.2.4` (to be used only in the `v2.6.3-patch1` release).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,11 +33,11 @@ replace (
 
 require (
 	github.com/gorilla/mux v1.8.0
-	github.com/rancher/dynamiclistener v0.3.1-0.20211119182849-962b63526960
+	github.com/rancher/dynamiclistener v0.3.2-0.20211221223804-148d38076d3e
 	github.com/rancher/lasso v0.0.0-20210709145333-6c6cd7fd6607
 	github.com/rancher/lasso/controller-runtime v0.0.0-20210608205930-775fcaf2f523
 	github.com/rancher/rancher/pkg/apis v0.0.0-20210628154046-7a2fc74f9598
-	github.com/rancher/wrangler v0.8.7
+	github.com/rancher/wrangler v0.8.9
 	github.com/sirupsen/logrus v1.8.1
 	golang.org/x/tools v0.1.2
 	k8s.io/api v0.22.3
--- a/go.sum
+++ b/go.sum
@@ -774,8 +774,8 @@ github.com/rancher/aks-operator v1.0.1-rc11 h1:OGxwaQ+MmqTNl/45wfS0PuI78mvo/FXfB
 github.com/rancher/aks-operator v1.0.1-rc11/go.mod h1:V43rw+oKpjQ2tlIITgejlatpNBGi61DdB+RldJyrh/8=
 github.com/rancher/client-go v1.22.3-rancher.1 h1:aNVLaIY5YGah1i9wRVXLOGRbLyekohjQAKHXeQm6Cxo=
 github.com/rancher/client-go v1.22.3-rancher.1/go.mod h1:ElDjYf8gvZsKDYexmsmnMQ0DYO8W9RwBjfQ1PI53yow=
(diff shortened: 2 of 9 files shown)


cmurphy (commented):
Are you sure the CAPI changes in b912e8c6034529c12716e63e4edc2d692cbc6ef2 are meant to be included in 2.6.3-patch1? This would have an impact on QA.

maxsokolovsky:
> Are you sure the CAPI changes in [b912e8c](https://github.com/rancher/webhook/commit/b912e8c6034529c12716e63e4edc2d692cbc6ef2) are meant to be included in 2.6.3-patch1? This would have an impact on QA.

@cmurphy, good point. I included it because the latest relevant commit was created after that CAPI bump one, so I thought, it may be needed for my latest commit. My plan was to include everything from the tag `v0.2.3` until my change, inclusively.

I am not 100% certain that my commit would have completely functional code without that earlier CAPI bump. By looking at it, I don't think that bump affects it. But may have to test first locally. Should we remove that commit?

thedadams:
@maxsokolovsky The CAPI stuff should not be in 2.6.3-patch1. Your dependency bump shouldn't require those changes.

maxsokolovsky:
Thanks, updated the PR, ready for review.

cmurphy (approved):
Thanks!

go.mod
```diff
@@ -33,11 +33,11 @@ replace (
```
cbron, line 33: > Once this PR is merged, I will tag the resulting latest commit as v0.2.4 (to be used only in the v2.6.3-patch1 release).

Are we going to release this tagged as 0.2.4, or as 2.6.3-patch1

cbron (commented):
Looks ok to me. Hard to tell with wrangler but both of those 0.8.8 and 0.8.9 releases look fairly unrelated. One question on tagging.

go.mod
maxsokolovsky: I had checked before opening the PR with @aiyengar2. We'd prefer not to disturb the `v0.2.4` series of chart RCs if possible.
Assuming we can have a chart with `2.6.3-patch1` as the version, we'd choose to go that route?

cbron: I think that makes sense

cbron (approved):
lgtm, if @thedadams is around maybe he can also approve the patch tag approach (see convo above).

go.mod
thedadams: @maxsokolovsky @cbron We should not tag this release of rancher-webhook as `2.6.3-patch1`:

Anyone that updates to Rancher v2.6.3-patch1 would then get rancher-webhook `v2.6.3-patch1`. If they then tried to update to Rancher v2.6.4 when it is released, then the logic in Rancher would keep rancher-webhook version `v2.6.3-patch1` because this version is greater than the `MIN_WEBHOOK_VERSION` in Rancher v2.6.4 (which will be `v0.2.4`).

The problem with this is that rancher-webhook `v0.2.4` will have important updates in it related to CAPI that `v2.6.3-patch1` will not have. This is likely to break Rancher.

thedadams (approved):
As long as we don't tag this version of rancher-webhook as `v2.6.3-patch1`, then I am good with these changes.

cbron:
Thats exactly what we were going to do 😄

go.mod
cbron: Sorry we got mixed up on versions, we are going to tag this as 0.2.3-patch1. So you have: 
2.6.3: webhook 0.2.3
2.6.3-patch1: webhook 0.2.3-patch1
2.6.4: webhook 0.2.4
Going that route we should be fine, as 0.2.4 is greater than 0.2.3-patch1.

PR #78, opened 2022-03-29
Update CRTB/PRTB validation handlers to disallow RoleTemplateName updates

ryansann:
These changes bring the CRTB and PRTB CRs more in line with the underlying k8s RBAC types, which disallow updates to RoleRef. This also addresses a bug with Apply in Rancher, where a CRTB can't be synced because a RoleBinding is applied with a different RoleRef after the CRTB had it's RoleTemplateName updated

### DIFF
--- a/pkg/resources/validation/clusterroletemplatebinding/clusterrtb.go
+++ b/pkg/resources/validation/clusterroletemplatebinding/clusterrtb.go
@@ -1,13 +1,17 @@
 package clusterroletemplatebinding
 
 import (
+	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/rancher/webhook/pkg/auth"
 	v3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
 	objectsv3 "github.com/rancher/webhook/pkg/generated/objects/management.cattle.io/v3"
 	"github.com/rancher/wrangler/pkg/webhook"
+	admissionv1 "k8s.io/api/admission/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/trace"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/resources/validation/clusterroletemplatebinding/clusterrtb.go
```diff
@@ -31,12 +35,28 @@ func (c *clusterRoleTemplateBindingValidator) Admit(response *webhook.Response,
 	if err != nil {
 		return err
 	}
-
-	if crtb.ClusterName != "local" {
+	if crtb.ClusterName != "local" { // the prtb custom resource only exists in the local cluster
```
MbolotSuse, line 38: Two Things here:

1. I don't think that the ClusterName in this case refers to the name where the object is in. I think this is the name of the cluster that the binding is in effect for (e.x. user y has access to role rt in cluster x -> in this case, x is the ClusterName). Do we really want to have a short circut in here? I kind of think that this should be removed.
2. Nit: this file is for crtbs, but the comment references prtbs. I recommend revising this.

pkg/resources/validation/projectroletemplatebinding/projectrtb.go
```diff
@@ -34,12 +38,28 @@ func (p *projectRoleTemplateBindingValidator) Admit(response *webhook.Response,
 	}
 
 	clusterID, projectNS := clusterFromProject(prtb.ProjectName)
-
-	if clusterID != "local" {
+	if clusterID != "local" { // the crtb custom resource only exists in the local cluster
```
MbolotSuse, line 41: Similar comment here as on the above (comment refers to crtbs but this file is for prtbs, and the cluster ID refers to the cluster that the permissions are effective for). Leaving this comment so we have a note here, but let's reserve discussion for the above comment.

MbolotSuse (changes requested):

MbolotSuse (approved):
LGTM

rmweir (approved):
LGTM

PR #80, opened 2022-04-04
Move base image to SLE BCI and run as non-privileged user

macedogm:
Move base image to SLE BCI and run as non-privileged user.

Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,10 +1,9 @@
-FROM golang:1.17
+FROM registry.suse.com/bci/golang:1.17-11.51
 
 ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
 
-RUN apt update && \
-    apt install -y bash git gcc docker.io vim less file curl wget ca-certificates
+RUN zypper -n install git docker vim less file curl wget
 RUN if [ "${ARCH}" = "amd64" ]; then \
     curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.44.0; \
     fi
@@ -15,6 +14,7 @@ ENV DAPPER_SOURCE /go/src/github.com/rancher/webhook/
 ENV DAPPER_OUTPUT ./bin ./dist
 ENV DAPPER_DOCKER_SOCKET true
 ENV DAPPER_RUN_ARGS "-v webhook-pkg:/go/pkg -v webhook-cache:/root/.cache"
+ENV GOPATH /go
 WORKDIR ${DAPPER_SOURCE}
 
 ENTRYPOINT ["./scripts/entry"]
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,3 +1,14 @@
-FROM alpine
+FROM registry.suse.com/bci/bci-micro:15.3.16.3
+
+ARG user=webhook
+
+RUN echo "$user:x:1000:1000::/home/$user:/bin/bash" >> /etc/passwd && \
+    echo "$user:x:1000:" >> /etc/group && \
+    mkdir /home/$user && \
+    chown -R $user:$user /home/$user
+
 COPY bin/webhook /usr/bin/
+
+USER $user
+
 CMD ["webhook"]
--- a/.drone.yml
+++ b/.drone.yml
@@ -124,68 +124,6 @@ volumes:
 
 ---
 kind: pipeline
-name: arm
-
-platform:
-  os: linux
-  arch: arm
-
-steps:
-- name: build
-  image: rancher/dapper:v0.4.1
-  commands:
-  - dapper ci
-  volumes:
-  - name: docker
-    path: /var/run/docker.sock
-
-- name: github_binary_release
-  image: plugins/github-release
-  settings:
-    api_key:
-      from_secret: github_token
-    prerelease: true
-    checksum:
-    - sha256
-    checksum_file: CHECKSUMsum-arm.txt
-    checksum_flatten: true
-    files:
-    - "dist/artifacts/*"
-  when:
-    instance:
-    - drone-publish.rancher.io
-    ref:
-    - refs/head/master
-    - refs/tags/*
-    event:
-    - tag
-
-- name: docker-publish
-  image: plugins/docker
-  settings:
-    dockerfile: package/Dockerfile
-    password:
-      from_secret: docker_password
-    repo: "rancher/rancher-webhook"
-    tag: "${DRONE_TAG}-arm"
-    username:
-      from_secret: docker_username
-  when:
-    instance:
-    - drone-publish.rancher.io
-    ref:
-    - refs/head/master
-    - refs/tags/*
-    event:
-    - tag
-
-volumes:
-- name: docker
-  host:
-    path: /var/run/docker.sock
-
----
-kind: pipeline
 name: s390x
 
 platform:
@@ -272,7 +210,6 @@ steps:
     platforms:
       - linux/amd64
       - linux/arm64
-      - linux/arm
       - linux/s390x
     target: "rancher/rancher-webhook:${DRONE_TAG}"
     template: "rancher/rancher-webhook:${DRONE_TAG}-ARCH"
@@ -288,5 +225,4 @@ steps:
 depends_on:
 - amd64
 - arm64
-- arm
 - s390x


package/Dockerfile
```diff
@@ -1,3 +1,6 @@
-FROM alpine
+FROM registry.suse.com/suse/sle15:15.3
+RUN zypper -n update && \
```
dcermak, line 2: This shouldn't be necessary, we update the bci base image frequently.

macedogm: I added this just to be on the safe side. Twice I pulled an image and `zypper update` downloaded a few packages. Maybe it was related to that issue where the packages were not being properly refreshed? 🤷🏻‍♂️ 

Do you believe it's safe to remove it?

ryansann (approved):
LGTM

package/Dockerfile
```diff
@@ -1,3 +1,7 @@
...
+    zypper -n clean -a && \
+    rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* /usr/share/doc/manual/* /var/log/*
```
dcermak, line 4: Since you really don't need any additional packages, why not use `bci-micro` instead? It's much smaller
```suggestion
FROM registry.suse.com/bci/bci-micro:15.3
```
Note, that `bci-micro` doesn't ship zypper, so the `RUN` part has to be dropped.

package/Dockerfile
```diff
@@ -1,3 +1,7 @@
-FROM alpine
+FROM registry.suse.com/bci/bci-base:15.3
+RUN zypper -n update && \
+    zypper -n clean -a && \
+    rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* /usr/share/doc/manual/* /var/log/*
 COPY bin/webhook /usr/bin/
+USER 1000
```
MbolotSuse, line 6: BCI base does not appear to have a user with an ID of 1000 by default.[Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user), indicate that you are going to want to create the user before attempting to use it. I'm unsure of the exact impact of trying to use a user like this which doesn't have a corresponding entry in /etc/passwd, but I would prefer that we add the user first unless you have a specific reason not to.

Dockerfile.dapper
```diff
@@ -15,6 +14,7 @@ ENV DAPPER_SOURCE /go/src/github.com/rancher/webhook/
 ENV DAPPER_OUTPUT ./bin ./dist
 ENV DAPPER_DOCKER_SOCKET true
 ENV DAPPER_RUN_ARGS "-v webhook-pkg:/go/pkg -v webhook-cache:/root/.cache"
+ENV GOPATH /go
```
MbolotSuse, line 17: Would you mind explaining this line a bit? From the docs, the GOPATH should default to `$HOME/go` for the user that this is running under - why do we want to specify a custom location?

package/Dockerfile
MbolotSuse: @macedogm Any comments on the above? I think that the final image still uses bci-base (rather than micro/minimal) - would you mind elaborating on why?

package/Dockerfile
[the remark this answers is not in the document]
macedogm: Good point, let me update it.

Dockerfile.dapper
dcermak: You can omit this one, the next version of the go image will have this setting included.

macedogm: This was a tricky one to find the issue and fix it. We set `ENV DAPPER_SOURCE /go/src/github.com/rancher/webhook/` and the previous image `golang:1.17` set by default `GOPATH=/go`.

```bash
% docker run --rm -it golang:1.17 env
PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
GOLANG_VERSION=1.17.9
GOPATH=/go
TERM=xterm
HOME=/root
```

However `bci/golang` doesn't set `GOPATH`, so the packages will end up in `$HOME/go` while the code being compiled is in `ENV DAPPER_SOURCE /go/src/github.com/rancher/webhook/`. This was breaking the build.

```bash
% docker run --rm -it registry.suse.com/bci/golang:1.17 env
GOLANG_VERSION=1.17
PATH=/go/bin:/usr/local/go/bin:/root/go/bin/:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
TERM=xterm
HOME=/root
```

macedogm: @dcermak if I omit this now it will not build.

package/Dockerfile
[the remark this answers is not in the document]
macedogm: Fixed by creating an user and group to run the container.

Dockerfile.dapper
dcermak: @macedogm yeah, apparently the new version is not yet up

package/Dockerfile
MbolotSuse: I understand doing that in the case of the core rancher containers, but does this really provide value in the case of side functionality like the webhook? I would also expect that running as a non-root, non-wheeled user in the container would prevent you from using zypper in the first place - if you disagree let me know.

macedogm: You are right in the rationale. When running as non-root user we could move directly to [bci/bci-minimal](https://registry.suse.com/static/bci/bci-minimal/index.html) or [bci/bci-micro](https://registry.suse.com/static/bci/bci-micro/index.html), but for the moment they are still in `Tech Preview`. I would say that we could stay with `bci/bci-base` for now and then switch.

package/Dockerfile
```diff
@@ -1,3 +1,7 @@
-FROM alpine
+FROM registry.suse.com/bci/bci-base:15.3
+RUN zypper -n update && \
+    zypper -n clean -a && \
+    rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/* /usr/share/doc/manual/* /var/log/*
```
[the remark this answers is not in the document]
MbolotSuse: My only concern there would be that zypper's not the only program that uses those directories - while it may be tolerant of these changes other programs may not be. That being said, we are probably ok here for the following reasons:

1) We aren't installing anything new.

2) The only other dir in `/var/log` was `krb5` (kerberos - I don't see any reason why the webhook would need this so if this does cause an issue with kerberos functionality that's probably ok).

@macedogm would you mind adding a comment here (or somewhere adjacent) so that this is documented? I'm thinking something like "Removing potential sparse files created by zypper update to save space"

package/Dockerfile
macedogm: Another point is that `bci/bci-minimal` and `bci/bci-micro` don't have `useradd` command, so either we create the user by hand, ask BCI team to add it or stick with the `USER` command in Dockerfile. Either way we have some trade-offs to choose.

dcermak: If the busybox userland is sufficient for you, then you could also use `bci/busybox`, which ships `useradd` (the busybox implementation though so there might be :dragon:).

dcermak: > I did not realize that minimal was in tech preview - good point (though micro seems to be out of tech preview for the non-latest tags). 

All of the images, except `bci-base` should be tech-preview. Note that this really just means that the image is not officially supported by SUSE's commercial support and that's about it.

> As far as `useradd` goes - that's a really good point that we should mention regardless of what we end up doing here. @dcermak do you happen to know why these aren't included in these images? The "add a non-root user and use that" pattern is fairly common (if I am not mistaken).

We haven't included it, because no one requested it and we only put the bare minimum in. You could work around this with the minimal image though by downloading the useradd rpm in a previous stage, copying it into the minimal image, installing it via rpm, adding the user and then removing it in the same step. Or you can make a request for useradd to be included in these images.

package/Dockerfile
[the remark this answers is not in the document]
macedogm: Base image moved to bci-micro. No need to run zypper anymore and we only clean the `zypp` log directory now.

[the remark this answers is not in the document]
macedogm: The directory is in the image:

```
$ docker run --rm registry.suse.com/bci/bci-micro:15.3 ls -lhaR /var/log
/var/log:
total 12K
drwxr-xr-x  3 root root 4.0K Mar 15 11:35 .
drwxr-xr-x 10 root root 4.0K Mar 22 18:07 ..
drwxr-xr-x  2 root root 4.0K Mar 22 18:07 zypp

/var/log/zypp:
total 20K
drwxr-xr-x 2 root root 4.0K Mar 22 18:07 .
drwxr-xr-x 3 root root 4.0K Mar 15 11:35 ..
-rw-r--r-- 1 root root 9.3K Mar 22 18:07 history
```

I'll file a bug report.

MbolotSuse (changes requested):
Similar comments as in https://github.com/rancher/webhook-receiver/pull/11, reserving discussion for that PR.

macedogm:
> Similar comments as in [rancher/webhook-receiver#11](https://github.com/rancher/webhook-receiver/pull/11), reserving discussion for that PR.

Answers are in the there [rancher/webhook-receiver#11](https://github.com/rancher/webhook-receiver/pull/11) PR.

MbolotSuse (approved):
LGTM - feedback was resolved in the receiver PR, no further issues, container seems to work in basic functionality test.

PR #77, opened 2022-03-17
Adding default linux tolerations and node selectors

MbolotSuse:
Adding default toleration/node selector to ensure webhook pods
are scheduled on linux nodes. Related to rancher/rancher#36401

PR #81, opened 2022-04-20
Update chart for new anno

ryansann:


MbolotSuse (changes requested):
If you want to permit this in mixed clusters, you will need to add node selectors to ensure that the webhook is scheduled on a linux node (from what I know about this, you can't run webhook, which is a linux container on a windows node, which can only run windows containers). You can see what this looks like here: https://github.com/rancher/webhook/pull/77/files .

ryansann:
> If you want to permit this in mixed clusters, you will need to add node selectors to ensure that the webhook is scheduled on a linux node (from what I know about this, you can't run webhook, which is a linux container on a windows node, which can only run windows containers). You can see what this looks like here: https://github.com/rancher/webhook/pull/77/files .

This doesn't control what nodes it gets deployed to. This anno allows it to be deployed into hybrid setups. The `os` anno controls what nodes it gets deployed on. This anno is patched in anyway as part of `make charts`, putting here in the upstream chart just allows for us to not have to make manual changes to generated changes in the charts repo

PR #85, opened 2022-06-03
Upgrade kubectl version to v1.23.3

rohitsakala:
This version supports s390x arch.


kinarashah (approved):
LGTM, can update to 1.24.x after 1.24 support is introduced.

PR #86, opened 2022-06-13
Enable Dependabot for image updates

macedogm:


KevinJoiner (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
...
+ updates:
+   - package-ecosystem: "docker"
+     directory: "/"
+     schedule:
+       interval: "daily"
+     open-pull-requests-limit: 10
+     pull-request-branch-name:
+       separator: "-"
+     reviewers:
+       - "teams/rancher-security"
```
KevinJoiner, line 11: Do we want to use this team or something like `teams/rancher-team-1-neo-dev`

macedogm: I think that for the moment `teams/rancher-security` is fine, so we don't overwhelm the entire team. If this trial process works fine and we expand it, then adding `teams/rancher-team-1-neo-dev` team is the logical step.

PR #100, opened 2022-07-26
Adds check for circular references to RoleTemplates

MbolotSuse:
When 2 role templates inherit from each other, this can cause
performance issues or an outright crash. This change aims to
ensure that role templates can't form circular references like
this in the future.

### DIFF
--- a/pkg/resources/validation/roletemplate/roletemplate.go
+++ b/pkg/resources/validation/roletemplate/roletemplate.go
@@ -1,10 +1,13 @@
 package roletemplate
 
 import (
+	"fmt"
 	"net/http"
 	"time"
 
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/webhook/pkg/auth"
+	controllerv3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
 	objectsv3 "github.com/rancher/webhook/pkg/generated/objects/management.cattle.io/v3"
 	"github.com/rancher/wrangler/pkg/webhook"
 	"github.com/sirupsen/logrus"
@@ -19,14 +22,16 @@ var roleTemplateGVR = schema.GroupVersionResource{
 	Resource: "roletemplates",
 }
 
-func NewValidator(escalationChecker *auth.EscalationChecker) webhook.Handler {
+func NewValidator(templates controllerv3.RoleTemplateCache, escalationChecker *auth.EscalationChecker) webhook.Handler {
 	return &roleTemplateValidator{
 		escalationChecker: escalationChecker,
+		templates:         templates,
 	}
 }
 
 type roleTemplateValidator struct {
 	escalationChecker *auth.EscalationChecker
+	templates         controllerv3.RoleTemplateCache
 }
 
 func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webhook.Request) error {
@@ -42,6 +47,22 @@ func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webho
 	// this admits update operations that happen to remove finalizers
 	if roleTemplate.DeletionTimestamp != nil {
 		response.Allowed = true
+		return nil
+	}
+	//check for circular references produced by this role
+	circularTemplate, err := r.checkCircularRef(roleTemplate)
+	if err != nil {
+		logrus.Errorf("Error when trying to check for a circular ref: %s", err)
+		return err
+	}
+	if circularTemplate != nil {
+		response.Result = &metav1.Status{
+			Status:  "Failure",
+			Message: fmt.Sprintf("Circular Reference: RoleTemplate %s already inherits RoleTemplate %s", circularTemplate.Name, roleTemplate.Name),
+			Reason:  metav1.StatusReasonBadRequest,
+			Code:    http.StatusBadRequest,
+		}
+		response.Allowed = false
 		return nil
 	}
 
@@ -76,4 +97,35 @@ func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webho
 	}
 
 	return r.escalationChecker.ConfirmNoEscalation(response, request, rules, "")
+}
+
+// checkCircularRef looks for a circular ref between this role template and any role template that it inherits
+// for example - template 1 inherits template 2 which inherits template 1. These setups can cause high cpu usage/crashes
+// If a circular ref was found, returns the first template which inherits this role template. Returns nil otherwise.
+// Can return an error if any role template was not found.
+func (r *roleTemplateValidator) checkCircularRef(template *v3.RoleTemplate) (*v3.RoleTemplate, error) {
+	seen := make(map[string]struct{})
+	queue := []*v3.RoleTemplate{template}
+	for len(queue) > 0 {
+		current := queue[0]
+		queue = queue[1:]
+		for _, inherited := range current.RoleTemplateNames {
+			// if we found a circular reference, exit here and go no further
+			if inherited == template.Name {
+				// note: we only look for circular references to this role. We don't check for circular dependencies which
+				// don't have this role as one of the targets. Those should have been taken care of when they were originally made
+				return current, nil
+			}
+			// if we haven't seen this yet, we add to the queue to process
+			if _, ok := seen[inherited]; !ok {
+				newTemplate, err := r.templates.Get(inherited)
+				if err != nil {
+					return nil, fmt.Errorf("unable to get roletemplate %s with error %w", inherited, err)
+				}
+				seen[inherited] = struct{}{}
+				queue = append(queue, newTemplate)
+			}
+		}
+	}
+	return nil, nil
 }
--- a/pkg/resources/validation/roletemplate/roletemplate_test.go
+++ b/pkg/resources/validation/roletemplate/roletemplate_test.go
@@ -1,0 +1,143 @@
+package roletemplate
+
+import (
+	"strconv"
+	"testing"
+
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	controllerv3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
+	"github.com/rancher/webhook/pkg/mocks"
+	"github.com/stretchr/testify/assert"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var defaultRules = []rbacv1.PolicyRule{
+	{
+		APIGroups: []string{""},
+		Resources: []string{"pods"},
+		Verbs:     []string{"get", "list", "watch"},
+	},
+}
+
+var circleRoleTemplateName = "circleRef"
+
+func TestCheckCircularRef(t *testing.T) {
+	tests := []struct {
+		name           string
+		depth          int
+		circleDepth    int
+		errorDepth     int
+		hasCircularRef bool
+		errDesired     bool
+	}{
+		{
+			name:           "basic test case - no inheritance, no circular ref or error",
+			depth:          0,
+			circleDepth:    -1,
+			errorDepth:     -1,
+			hasCircularRef: false,
+			errDesired:     false,
+		},
+		{
+			name:           "basic inheritance case - depth 1 of input is circular",
+			depth:          1,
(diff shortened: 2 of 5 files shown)


MbolotSuse:
## Background

Related to rancher/rancher#38419. Basically when any two roleTemplates reference one another, this forms a circular reference. When rancher attempts to discover how many roles the user needs to be bound to, it uses a high amount of CPU or produces a stack overflow, crashing rancher.

## Changes

- Adds new `checkCircularRef` function which:
  - Takes a role template as input
  - Looks through the inherited role templates, using a depth-first-search (queue to handle which item is processed next, set to mark nodes that we are already processing)
  - If we find the input function in any inherited role templates, break and return that role templates
  - Return an error if any specified role template is not in the cache
- Adds new logic to return a 400 error if a circular dependency was found, which specifies which role template is the source of the circular dependency.

## FAQ

Why does this only check for circular references to the original role? Isn't it possible that there could be a circular reference in another part of the dependency graph which doesn't involve the original role?

In order to maintain simplicity, this only checks for circular dependencies with regard to the changed role template. The idea is that if a customer wanted to create a circular place elsewhere, they would need to individually change that role. In that case, this other role would be the primary changed role template, and they would need to change this other role, and this code would prevent that issue at that point.

This will require customers to adjust any active circular templates on their own, but given that these roleTemplates are customer managed I think that's ok. In addition, this will prevent confusing errors. For example say I have 3 roles of the structure, 1 -> 2 -> 3 -> 2. In addition to getting errors if I try to change 2/3 without resolving the circular dependency, I will also get an error when attempting to update 1, despite the fact that 1 is not the source of the issue.

pkg/resources/validation/roletemplate/roletemplate.go
```diff
@@ -77,3 +98,35 @@ func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webho
 
 	return r.escalationChecker.ConfirmNoEscalation(response, request, rules, "")
 }
+
+// checkCircularRef looks for a circular ref between this role template and any role template that it inherits
+// for example - template 1 inherits template 2 which inherits template 1. These setups can cause high cpu usage/crashes
+// If a circular ref was found, returns the first template which inherits this role template. Returns nil otherwise.
+// Can return an error if any role template was not found
+func (r *roleTemplateValidator) checkCircularRef(template *v3.RoleTemplate) (*v3.RoleTemplate, error) {
+	seen := make(map[string]struct{}, 0)
```
maxsokolovsky, line 107: Nit: you don't need to provide the second argument to `make` if you create and initialize a map (unless you want a specific starting size).

pkg/resources/validation/roletemplate/roletemplate.go
```diff
@@ -77,3 +98,35 @@ func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webho
...
+	queue := make([]*v3.RoleTemplate, 0)
```
maxsokolovsky, line 108: Nit: this could be simplified to just `queue := []*v3.RoleTemplate{template}`.

pkg/resources/validation/roletemplate/roletemplate.go
MbolotSuse: Done - thanks for the suggestion.

pkg/resources/validation/roletemplate/roletemplate.go
MbolotSuse: Done - thanks for the suggestion.

maxsokolovsky (changes requested):
I'd say this warrants a unit test of the behavior of the `checkCircularRef` method.
The only problem is mocking the `r.templates.Get` call.
Or could this method be refactored into a function that receives a RoleTemplateCache as its parameter? And then making or using an existing test version of the cache is easier?

MbolotSuse:
@maxsokolovsky Given the recent discussion I can add a test. That being said, I'll probably test Admit directly since that's the exported method. I don't think that a refactored method would be any easier to test and would be less readable, so I'm not going to change that value. I will probably need to create a mock version of the Cache - I don't think there exists one today.

MbolotSuse:
Added some unit tests here for the side function (exported function had a good bit of dependencies - would have been a lot of code to test that out).

One item of note - I've created a `mocks` folder at the root of the project which one can use to store mock versions of objects. If there's an objection to this structure let me know - happy to discuss alternatives.

pkg/resources/validation/roletemplate/roletemplate_test.go
```diff
@@ -0,0 +1,142 @@
...
+			circleDepth:    -1,
+			errorDepth:     0,
+			hasCircularRef: false,
+			errDesired:     true,
+		},
+	}
+
+	for _, testCase := range tests {
+		t.Run(testCase.name, func(t *testing.T) {
+			mockCache := mocks.NewMockRoleTemplateCache()
```
maxsokolovsky, line 77: Looks like you could run these subtests in parallel, since each test gets its own template cache, and there would be no concurrent writes to anything.
`t.Parallel()` inside the subtest function would do that, but you'd also need to capture the `testCase` variable inside the for loop but outside the subtest function.

```
for _, testCase := range tests {
    testCase := testCase
    t.Run(testCase.name, func(t *testing.T) {
        t.Parallel()
        mockCache := mocks.NewMockRoleTemplateCache()
```

pkg/resources/validation/roletemplate/roletemplate_test.go
```diff
@@ -0,0 +1,142 @@
...
+	{
+		APIGroups: []string{""},
+		Resources: []string{"pods"},
+		Verbs:     []string{"get", "list", "watch"},
+	},
+}
+
+var circleRoleTemplateName = "circleRef"
+
+func Test_checkCircularRef(t *testing.T) {
```
maxsokolovsky, line 24: The name, by convention, would just be `TestCheckCircularRef` in this case.

pkg/resources/validation/roletemplate/roletemplate_test.go
```diff
@@ -0,0 +1,142 @@
...
+		t.Run(testCase.name, func(t *testing.T) {
+			mockCache := mocks.NewMockRoleTemplateCache()
+			rtName := "input-role"
+			if testCase.circleDepth == 0 && testCase.hasCircularRef {
+				rtName = circleRoleTemplateName
+			}
+			inputRole := createNestedRoleTemplate(rtName, mockCache, testCase.depth, testCase.circleDepth, testCase.errorDepth)
+			validator := createValidator(mockCache)
+			result, err := validator.checkCircularRef(inputRole)
+			t.Logf("Running %v got %v and %v", testCase.name, result, err)
```
maxsokolovsky, line 85: Do you need this log statement? The test runner would show the name of the subtest anyways, and your error checks already look at the result...

pkg/resources/validation/roletemplate/roletemplate.go
```diff
@@ -44,6 +49,22 @@ func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webho
...
+	//check for circular references produced by this role
+	circularTemplate, err := r.checkCircularRef(roleTemplate)
+	if err != nil {
+		logrus.Errorf("Error when trying to check for a circular ref: %s", err)
+		return err
+	}
+	if circularTemplate != nil {
+		response.Result = &metav1.Status{
+			Status:  "Failure",
+			Message: fmt.Sprintf("Circular Reference: RoleTemplate %s alrady inherits RoleTemplate %s", circularTemplate.Name, roleTemplate.Name),
```
maxsokolovsky, line 61: Typo: already.

pkg/resources/validation/roletemplate/roletemplate_test.go
MbolotSuse: That's not how at least one of the other webhook's tests are named. See: https://github.com/rancher/webhook/blob/53b1ff449ed461449beeca22b57e7aa212052c8d/pkg/resources/validation/cluster/provisioningcluster_test.go#L7 .

I can switch it to the standard go testing convention, but I think this discrepancy should be noted.

pkg/resources/validation/roletemplate/roletemplate.go
MbolotSuse: Good catch thanks.

pkg/resources/validation/roletemplate/roletemplate_test.go
maxsokolovsky: Yeah, it's all immaterial really, that other example is OK despite not following the convention, but I don't believe that convention is strict...

pkg/resources/validation/roletemplate/roletemplate_test.go
MbolotSuse: I can put this in there, but the performance impact here was negligible (hovers around `~0.008s` either way).

maxsokolovsky: This is not so much about this test on its own but rather all the tests. `t.Parallel` allows a given test to run concurrently with other tests.

pkg/resources/validation/roletemplate/roletemplate_test.go
MbolotSuse: Change made.

pkg/resources/validation/roletemplate/roletemplate.go
MbolotSuse: Change made.

pkg/resources/validation/roletemplate/roletemplate_test.go
MbolotSuse: Change made.

pkg/resources/validation/roletemplate/roletemplate_test.go
```diff
@@ -0,0 +1,142 @@
...
+			depth:          1,
+			circleDepth:    -1,
+			errorDepth:     0,
+			hasCircularRef: false,
+			errDesired:     true,
+		},
+	}
+
+	for _, testCase := range tests {
+		t.Run(testCase.name, func(t *testing.T) {
```
maxsokolovsky, line 77: I believe you'd need to capture the range variable just above the call to `Run`, like `testCase := testCase`.

MbolotSuse: I'm not sure what you mean here. I don't really see a need to capture any additional variables - can you explain exactly what you want to see and why?

MbolotSuse: Done - explained offline.

maxsokolovsky (approved):

pkg/resources/validation/roletemplate/roletemplate.go
```diff
@@ -77,3 +98,34 @@ func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webho
...
+			if inherited == template.Name {
+				// note: we only look for circular references to this role. We don't check for circular dependencies which
+				// don't have this role as one of the targets. Those should have been taken care of when they were originally made
+				return current, nil
+			}
+			// if we haven't seen this yet, we add to the queue to process
+			if _, ok := seen[inherited]; !ok {
+				newTemplate, err := r.templates.Get(inherited)
+				if err != nil {
+					return nil, err
```
KevinJoiner, line 123: Wrap error

pkg/resources/validation/roletemplate/roletemplate.go
```diff
@@ -77,3 +98,34 @@ func (r *roleTemplateValidator) Admit(response *webhook.Response, request *webho
 
 	return r.escalationChecker.ConfirmNoEscalation(response, request, rules, "")
 }
+
+// checkCircularRef looks for a circular ref between this role template and any role template that it inherits
+// for example - template 1 inherits template 2 which inherits template 1. These setups can cause high cpu usage/crashes
+// If a circular ref was found, returns the first template which inherits this role template. Returns nil otherwise.
+// Can return an error if any role template was not found
```
KevinJoiner, line 105:
```suggestion
// Can return an error if any role template was not found.
```

KevinJoiner (changes requested):

pkg/resources/validation/roletemplate/roletemplate.go
MbolotSuse: Done, added some extra detail here as well.

KevinJoiner (approved):

rmweir (approved):
LGTM, left comment regarding mock struct

PR #121, opened 2022-09-15
Adds priorityClassName to the webhook deployment.

KevinJoiner:
## Issue: https://github.com/rancher/rancher/issues/37927
Adds the ability for users to set the priorityClassName of rancher-webhook via helm values.

### DIFF
--- a/charts/rancher-webhook/templates/deployment.yaml
+++ b/charts/rancher-webhook/templates/deployment.yaml
@@ -50,3 +50,6 @@ spec:
         - name: tls
           mountPath: /tmp/k8s-webhook-server/serving-certs
       serviceAccountName: rancher-webhook
+      {{- if .Values.priorityClassName }}
+      priorityClassName: "{{.Values.priorityClassName}}"
+      {{- end }}
--- a/charts/rancher-webhook/values.yaml
+++ b/charts/rancher-webhook/values.yaml
@@ -23,3 +23,6 @@ preDelete:
 # tolerations for the webhook deployment. See https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ for more info
 tolerations: []
 nodeSelector: {}
+
+## PriorityClassName assigned to deployment.
+priorityClassName: ""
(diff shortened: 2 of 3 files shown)


cmurphy (approved):

charts/rancher-webhook/templates/deployment.yaml
```diff
@@ -50,3 +50,6 @@ spec:
         - name: tls
           mountPath: /tmp/k8s-webhook-server/serving-certs
       serviceAccountName: rancher-webhook
+      {{- if .Values.priorityClassName }}
+      priorityClassName: "{{.Values.priorityClassName}}"
```
MbolotSuse, line 54: Per the [k8s docs](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority) I don't think that we want quotation marks here. I could be wrong though, don't know much about this field.

charts/rancher-webhook/values.yaml
```diff
@@ -22,4 +22,5 @@ preDelete:
 
 # tolerations for the webhook deployment. See https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ for more info
 tolerations: []
-nodeSelector: {}
\ No newline at end of file
+nodeSelector: {}
+priorityClassName: ""
```
MbolotSuse, line 26: Nit: If you end up needing to edit this, would you mind adding a comment indicating what this is for? I'm not too attached to it since it maps exactly to a k8s value, but I think ti would be nice to mirror what we do for tolerations if we can.

MbolotSuse (changes requested):
One other comment:

There's another workload type that this chart deploys - a [pre-delete job](https://github.com/rancher/webhook/blob/release/v0.3/charts/rancher-webhook/templates/pre-delete-hook-job.yaml). We might want to consider adding this priority class there too.

I don't think it's strictly necessary since this is more of a clean-up job that would only be executed during uninstall, and I think it's reasonable that something like that would not have the same priority class as the core deployment. Though I can see the argument for it as part of a holistic approach.

cmurphy (approved):

PR #122, opened 2022-09-19
Role resolvers

KevinJoiner:
This PR adds the ability to get a user's rules in a downstream cluster/project via CRTBs/PRTBs. This removes the need for the escalate verb check.


pkg/resolvers/resolvers.go
```diff
@@ -0,0 +1,58 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+// ErrUnimplemented is an error returned when a function is not implemented.
+var ErrUnimplemented = fmt.Errorf("not implemented")
+
+// ruleAccumulator based off kubernetes struct
+// https://github.com/kubernetes/kubernetes/blob/d5fdf3135e7c99e5f81e67986ae930f6a2ffb047/pkg/registry/rbac/validation/rule.go#L124#L137
+type ruleAccumulator struct {
+	rules  []rbacv1.PolicyRule
```
rmweir, line 16: I think in this case it may be a good idea to use a slice of pointers over values. There will be tons of rules, many of which redundant I imagine, in a scaled setup. Adding another object to memory every time we append won't be necessary and we aren't writing to the rules or doing anything dangerous like that.

rmweir (changes requested):
Good job finding a solution that adds to the resolver.

Some thoughts:

Are we comfortable using  xrtb's to determine someone's access? Afaik this should be fine in this context since it is for validating the creation of xrtb's however it is worth noting that this will totally ignore any bindings that are added via kubectl or one that an app adds. cc: @cbron @MbolotSuse

@cbron can we have scaling QA see if there are any spikes when create a new prtb/crtb. If there are so many projects (from all clusters), will there be a spike when creating these slices of rules? I have witnessed 15k+ prtb's in a setup. There is probably no reason to optimize it at this time but I feel like there will be so many redundant rules returned by the resolvers that I want to make sure it isn't necessary to refine the rules we are passing around. I do worry about this less if we switch to using pointers for the PolicyRules.

KevinJoiner:
@rmweir I think this is a good question, and I will just add a little more context.
> Are we comfortable using xrtb's to determine someone's access?

The previous behavior was to only check a users permissions in the local cluster to determine access for both local and downstream clusters. This new approach keeps that behavior, but also checks the users permissions in local and downstream clusters via xrtbs as well.

> however it is worth noting that this will totally ignore any bindings that are added via kubectl or one that an app adds.

I don't believe this is true since all kube-api xrtb operation will pass through the webhook regardless of who originated them. Furthermore, the xrtbs that are used to determine permissions are taken from kube-api via a wrangler cache.

pkg/resolvers/resolvers.go
KevinJoiner: We would not save any space using pointers here since this list is ultimately returned by the `RulesFor` method which must adhere to the [validation.AuthroizationRuleResolver](https://github.com/kubernetes/kubernetes/blob/500f5aba80d71253cc01ac6a8622b8377f4a7ef9/pkg/registry/rbac/validation/rule.go#L36) interface.  This means we would have to inevitably allocate memory for each item in the list eventually.

An alternative to combat memory usage we have the option of using a map internal for the accumulator. This map would use `rbacv1.PolicyRule.String()` as the key so that we do not store duplicate rules and errors. This approach may save us some memory, but when benchmarked it caused significant speed decrease
```
goos: darwin
goarch: arm64
pkg: github.com/rancher/webhook/pkg/resolvers
BenchmarkGotRules-10                        3807            274032 ns/op
BenchmarkGotRulesWithMapCheck-10             852           1195656 ns/op
BenchmarkGotRulesWithMapOverride-10          993           1255381 ns/op
```
 **note this BenchMark resolved rules for users with 10 roleTemplates that each contained 100 rules.**

Since this list is evaluated then discarded the total memory usage should not be a large concern. What we really care about is the added time needed to allocate extra memory.

pkg/resolvers/resolvers.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// visitRules calls visitor on each rule in the list with the given Stringer and error.
+func visitRules(source fmt.Stringer, rules []rbacv1.PolicyRule, err error, visitor func(source fmt.Stringer, rule *rbacv1.PolicyRule, err error) bool) bool {
+	if err != nil && rules == nil {
+		return visitor(source, nil, err)
+	}
+	for i := range rules {
+		// we do not care about the return here
+		if !visitor(source, &rules[i], err) {
```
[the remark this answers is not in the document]
KevinJoiner: > Isn't that going to lead to a huge list of redundant errors that are not scoped per rule?

This is true, But we can only guarantee that errors are not scoped per rule with our implementation of `visitor func(source fmt.Stringer, rule *rbacv1.PolicyRule, err error)` which is why I was hesitant to add short circuit logic.
That being said, the interface of [validation.AuthroizationRuleResolver](https://github.com/kubernetes/kubernetes/blob/500f5aba80d71253cc01ac6a8622b8377f4a7ef9/pkg/registry/rbac/validation/rule.go#L36) does not provide any documentation stating errors must corrilate to a rule so I am okay with making the change.

pkg/resolvers/resolvers.go
```diff
@@ -0,0 +1,58 @@
...
+	return fmt.Errorf("[%s]", errorStr[leadingChars:])
+}
+
+// visitRules calls visitor on each rule in the list with the given Stringer and error.
+func visitRules(source fmt.Stringer, rules []rbacv1.PolicyRule, err error, visitor func(source fmt.Stringer, rule *rbacv1.PolicyRule, err error) bool) bool {
+	if err != nil && rules == nil {
+		return visitor(source, nil, err)
+	}
+	for i := range rules {
+		// we do not care about the return here
```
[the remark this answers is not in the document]
KevinJoiner: This was a left over comment from my first iteration where I was ignoring the return value of `visit`.

pkg/resolvers/aggregateResolver.go
```diff
@@ -0,0 +1,48 @@
...
+
+// NewAggregateRuleResolver creates a new AggregateRuleResolver that will combine the outputs of all resolvers provided.
+func NewAggregateRuleResolver(resolvers ...validation.AuthorizationRuleResolver) *AggregateRuleResolver {
+	return &AggregateRuleResolver{
+		resolvers: resolvers,
+	}
+}
+
+// GetRoleReferenceRules calls GetRoleReferenceRules on each resolver and returns all returned rules and errors.
+func (a *AggregateRuleResolver) GetRoleReferenceRules(roleRef rbacv1.RoleRef, namespace string) ([]rbacv1.PolicyRule, error) {
```
MbolotSuse, line 24: I'm a bit uncomfortable with our AggregateResolver atttempting to use this interface on the internal resolvers. We don't implement this method for our other resolvers, and this could easily lead to a panic if someone wasn't aware of that, or if the way that k8s used this method differed from the current impelmentation.

If we are going to fully implement this interface I think we should implement the interface for the crtb/prtb one as well.

rmweir (approved):
My question was actually around role bindings that were created by something other then the xrtb controllers. @KevinJoiner pointed out to me that the default resolver addresses this. LGTM.

PR #129, opened 2022-10-03
Adds ProjectName validation to PRTB.

KevinJoiner:
Adds ProjectName check to PRTB validation.


crobby (approved):
This looks reasonable to me and the added tests cover the new cases.

MbolotSuse (approved):
LGTM. Thank you for expanding the tests to include this new scenario.

Do be careful on milestones. With the upcoming 2.6.9 and 2.7.0 we need to make sure this goes (or doesn't go) in the proper release.

PR #136, opened 2022-10-13
[v2.7] Ensure provisioning cluster names conform to RFC-1123

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/39248
 
# Problem

Including `.` in the name of imported clusters or RKE2/K3s clusters results in the failure to deploy required fleet components  on the downstream cluster. Fleet expects cluster names to only include lower case alphanumeric characters and `-`, all other characters are disallowed. Some UI validation has been put in place to ensure that this naming format is not allowed, however it inappropriately permits the use of `.`, resulting in this issue. This issue does not reproduce for RKE1 clusters as the provisioning cluster is given an auto-generated name of the format `c-xxxxx`.

# Solution
Implement stricter validation of provisioning cluster names within the web-hook. While UI validation is helpful for users of the Rancher UI, implementing this check in the web-hook ensures that API calls or other methods of provisioning Rancher clusters are covered as well. The regex used in this PR matches the regex [used by fleet when verifying meta-data names](https://github.com/rancher/fleet/pull/918/files#diff-927d423e0c49728f25e29bb70ac2e0d993b9be8cebd4cfbaea474e18e34f3982R171)
 
# Testing

I've done the following tests, ensuring that the error message `admission webhook "rancherauth.cattle.io" denied the request: cluster name must be 63 characters or fewer, cannot be "local" nor of the form "c-xxxxx", and can only contain only lowercase alphanumeric characters or ' - '` is shown when invalid names are supplied.

I've also updated the unit tests to match this new requirement, and added additional test cases.

1. Attempting to create an RKE2 cluster which contains `.`
  a. Go to the cluster provisioning page, select RKE2 on any provider, and input the name `test.cluster`
  b. The error message was displayed correctly, and the cluster did not get created

2. Attempting to create an K3s cluster which contains `.`
  a. Go to the cluster provisioning page, select K3s on any provider, and input the name `test.cluster`
  b. The error message was displayed correctly, and the cluster did not get created

3. Attempting to create an imported cluster with a name containing `.`
 a. go to the cluster management screen and click `import existing`
 b. select generic kubernetes cluster
 c. enter the name `test.cluster`
 d. the error message was displayed correctly, and the imported cluster did not get created

4. Modifying an existing cluster which already contained `.` in its name 
  a. Create an RKE2/K3s cluster using a prior version of the web-hook, and wait for it to become available. 
  b. Upgrade the web-hook version to include this new naming requirement 
  c. Modify the provisioning cluster object in various ways, including changing registries, modifying aspects of the upgrade strategy, and modifying aspects of the machine pools. 
  d. Cluster modifications happened properly and the new web-hook logic did not interfere with the ability to manage existing clusters with invalid names.


pkg/resources/validation/cluster/provisioningcluster_test.go
```diff
@@ -75,25 +75,39 @@ func Test_isValidName(t *testing.T) {
...
 		{
 			name:             "name length is 253 characters",
-			clusterName:      "DxHt2WGXBASk8LPj4nFqMYCyKCsmZV6bWTL7xeo3NUXnW6tK07voFJJJmEpY6AvDhd03oR2hnW8UQJTDh2LVBPrh4V0rJoChGeaLMPTz4sQt3pt5sTcce4EiRk37YtjFQuHOdXKnQQzpIdll6TXReIQ9pPaASWuwpQ8opadHQITFlN2txSgOwc80WWGkgikcZh6f8fuihvviZF65tn2WBeySuDyEOFgLTaduG1CJwohm7n9Yovrd0fiYxM0BK",
+			clusterName:      "dxht2wgxbask8lpj4nfqmycykcsmzv6bwtl7xeo3nuxnw6tk07vofjjjmepy6avdhd03or2hnw8uqjtdh2lvbprh4v0rjochgealmptz4sqt3pt5stcce4eirk37ytjfquhodxknqqzpidll6txreiq9ppaaswuwpq8opadhqitfln2txsgowc80wwgkgikczh6f8fuihvvizf65tn2wbeysudyeofgltadug1cjwohm7n9yovrd0fiyxm0bk",
 			clusterNamespace: "fleet-default",
 			clusterExists:    false,
 			want:             false,
 		},
+		{
+			name:             "name containing . does not conform to RFC-1123",
```
jakefhyde, line 98: nice

jakefhyde (approved):

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -131,7 +132,7 @@ func (p *provisioningClusterValidator) validateClusterName(request *webhook.Requ
 	if !isValidName(cluster.Name, cluster.Namespace, err == nil) {
 		response.Result = &metav1.Status{
 			Status:  "Failure",
-			Message: "cluster name must be 63 characters or fewer, cannot be \"local\" nor of the form \"c-xxxxx\"",
+			Message: "cluster name must be 63 characters or fewer, cannot be \"local\" nor of the form \"c-xxxxx\", and can only contain only lowercase alphanumeric characters or ' - '",
```
rmweir, line 135:
```suggestion
			Message: "cluster name must be 63 characters or fewer, cannot be \"local\" nor of the form \"c-xxxxx\", and can only contain lowercase alphanumeric characters or ' - '",
```

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -23,6 +23,7 @@ import (
 const globalNamespace = "cattle-global-data"
 
 var mgmtNameRegex = regexp.MustCompile("^c-[a-z0-9]{5}$")
+var fleetNameRegex = regexp.MustCompile("^[-a-z0-9]*$")
```
rmweir, line 26: Should this just match the mgmt regex since each cluster usually has one of each, a provisioning and management cluster object with the same name?

rmweir (changes requested):

pkg/resources/validation/cluster/provisioningcluster.go
HarrisonWAffel: For RKE2/K3s the management and provisioning cluster objects don't share a name. The provisioning cluster is named based off of what the user inputs into the UI, and the management cluster gets a randomly generated name of the form `c-m-xxxxxxxx`.

jakefhyde (dismissed):

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -23,6 +23,7 @@ import (
 const globalNamespace = "cattle-global-data"
 
 var mgmtNameRegex = regexp.MustCompile("^c-[a-z0-9]{5}$")
```
rmweir, line 25: Interesting that the expected format of a creating a management cluster in the api is different than the one that is created as a by product of creating a provisioning cluster. Is that by design?

pkg/resources/validation/cluster/provisioningcluster.go
rmweir, line 26: So is "-----------------" a fine cluster name? And could + be used instead of "*" since we probably don't want an empty name?

rmweir (changes requested):

pkg/resources/validation/cluster/provisioningcluster.go
HarrisonWAffel: Yes it is, my understanding is that the different naming conventions allow us to determine if a cluster is legacy or not.

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -131,7 +132,7 @@ func (p *provisioningClusterValidator) validateClusterName(request *webhook.Requ
 	if !isValidName(cluster.Name, cluster.Namespace, err == nil) {
 		response.Result = &metav1.Status{
 			Status:  "Failure",
-			Message: "cluster name must be 63 characters or fewer, cannot be \"local\" nor of the form \"c-xxxxx\"",
+			Message: "cluster name must be 63 characters or fewer, cannot be \"local\" nor of the form \"c-xxxxx\", and can only contain lowercase alphanumeric characters or ' - '",
```
jakefhyde, line 135: Should be add something along the lines of "cannot start with `-` to the message?

pkg/resources/validation/cluster/provisioningcluster_test.go
```diff
@@ -75,25 +75,39 @@ func Test_isValidName(t *testing.T) {
 		},
 		{
 			name:             "name length is exactly 63 characters",
-			clusterName:      "Cq8Oh6UVnTyPOITcfwrXFjJRuZ4KV2q6ItimQkcGeX1ZQGm7oa3jbLd9N0diIka",
```
jakefhyde, line 78: Can we add a test that ensures something like what ricky suggested isn't a valid name (like `-----------------`)

jakefhyde (changes requested):

pkg/resources/validation/cluster/provisioningcluster_test.go
HarrisonWAffel: Okay, I've added two more test cases for this scenario

pkg/resources/validation/cluster/provisioningcluster.go
HarrisonWAffel: Sure, hopefully the wording i've added doesn't sound odd - it's a long message

jakefhyde (dismissed):
LGTM

rmweir (approved):
LGTM

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -23,6 +23,7 @@ import (
 const globalNamespace = "cattle-global-data"
 
 var mgmtNameRegex = regexp.MustCompile("^c-[a-z0-9]{5}$")
+var fleetNameRegex = regexp.MustCompile("^[^\\-][-a-z0-9]+$")
```
rmweir, line 26: nit: I'm not familiar, but is the \\ necessary here?

rmweir (changes requested):

pkg/resources/validation/cluster/provisioningcluster.go
HarrisonWAffel: I retested and I think you're right, it's not needed. The behavior of the regex doesn't change when its removed, but I'll take it out just for clarity

rmweir (approved):
LGTM

jakefhyde (approved):

snasovich:
@HarrisonWAffel , are we ready to merge this one since it has 2 approvals now?

PR #147, opened 2022-11-09
Updates to Go 1.19

KevinJoiner:
Also updates rancher/pkg/apis and K8s to 1.24.5
Includes generation of mocks to CI steps.


pkg/fakes/cluster.go
```diff
@@ -1,524 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
-// Source: github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3 (interfaces: ClusterController,ClusterClient,ClusterCache)
-
-// Package fakes is a generated GoMock package.
-package fakes
```
MbolotSuse, line 5: You mind elaborating on why this file was deleted?

KevinJoiner: It was not being used, so I didn't add it to the automatic generation.

go.mod
```diff
@@ -1,54 +1,54 @@
...
 
 require (
 	github.com/golang/mock v1.6.0
 	github.com/gorilla/mux v1.8.0
 	github.com/rancher/dynamiclistener v0.3.2-0.20211221223804-148d38076d3e
-	github.com/rancher/lasso v0.0.0-20220214201210-e50bfe3e2871
+	github.com/rancher/lasso v0.0.0-20220627205005-00d9c8e9dda6
 	github.com/rancher/lasso/controller-runtime v0.0.0-20220110204826-724e3e57c47d
-	github.com/rancher/rancher/pkg/apis v0.0.0-20210628154046-7a2fc74f9598
+	github.com/rancher/rancher/pkg/apis v0.0.0-20221103011339-cbbef69d4c2e
```
MbolotSuse, line 40: Thank you for updating this very out of date dependency. I think we should add this to the release steps for the adapter - i.e. update this value when RCs start being cut so that we are in line with rancher as much as possible (ideally we would have exactly what ships with the version that the webhook is for but that probably won't happen).

MbolotSuse (approved):
LGTM - I think the dependency changes should be caught by the build/test phases, so we should be good to go there.

cbron (approved):

PR #150, opened 2022-11-15
Webhook Routing

KevinJoiner:
# Commit 1 Decoding Update

### Old Decoding
- Required an unnecessary DeepCopy() in wrangler.Request.DecodeObject
- Required two unnecessary object cast in the generate objectv1 code
- Required wrangler.Request which needed to be modified with an object template.

### Updated Decoding
- Decodes the objects from the admission Request. Which removes the unneeded dependency on `wrangler.Request`.
- Unmarshals objects directly into the desired object instead of runtime.Object

### Notes
These changes don't have any explicit test files since only the generated code is changed. Though we can verify they work correctly by running unit tests for `pkg/resources/clusterroletemplatebing`

# Commit 2 Patch update

### Old patch
- Required an unnecessary json.Marshal of the original object.
- Could not guarantee that the object used in the patch was the same as the original object.
- Reintroduced the potential for this upstream bug. https://github.com/kubernetes-sigs/kubebuilder/issues/510

### Updated patch
The patch code now takes in a `[]byte` that is sent with the original request. This guarantees that the correct patch will be generated and removes the need to unmarshal and then marshal the original object in the request object.

### Notes
For more context, patching stores modification to an object in MutatingAdmissionWebhooks.
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response

# Commit 3 Routing Update

## Webhook Update:
The old Architecture for defining a new webhook handler had some drawbacks that get amplified as we start moving all validation and mutation to the webhook.

### Old drawbacks
Routing could not handle multiple matches. (management.cattle.io group and clusters.management.cattle.io)
Routing could not handle object Equivalency apps == v1beta-extensions.
Routing could not distinguish a request from the URL since all requests sent to either /validation or /mutation
Webhhooks were all forced to use the same configuration. (e.g. sideEffects,  failPolicy, matchPolicy, mutationPolicy)
Configuration  for each handler was scattered throughout the repo and required working knowledge of the webhook to add a new handler
    1. Define the Handler object.
    2. Add the CRD to the rules.go
    3. Create a new instance in validation.go
    4. Create a new route for your webhook in validation.go
    5. Add your new rule created in rules.go to a new or existing configuration in server.go

### Updated Routing features
Route matching is handled by Kubernetes each handler has a unique endpoint.
Webhhooks  Can use a default webhook configuration and modify it to fit their specific needs.
Configuration is centralized to the creation of one object
    1. Define the Handler object.
    2. Add a new instance in handlers.go


KevinJoiner:
I plan on adding an integration test to the webhook in a separate PR to confirm the behavior stays the same before and after these changes.

pkg/server/server.go
```diff
@@ -158,75 +185,36 @@ func listenAndServe(ctx context.Context, clients *clients.Clients, handler http.
 			},
 			CABundle: secret.Data[corev1.TLSCertKey],
 		}
+		if devURL, ok := os.LookupEnv("CATTLE_WEBHOOK_URL"); ok {
```
crobby, line 188: Love this...thank you

KevinJoiner:
Rebased with integration tests from #151 that hits every endpoint and test for valid and invalid operations from the supported rules.

crobby (dismissed):
I think this looks good.  I've been running on top of this (and building on top) for a week or so and it's working nicely.  Really appreciate the doc updates.

rmweir (changes requested):

pkg/resources/mutation/mutation.go
```diff
@@ -1,20 +1,26 @@
...
+func SetCreatorIDAnnotation(request *admission.Request, response *v1.AdmissionResponse, obj runtime.RawExtension, newObj metav1.Object) error {
 	annotations := newObj.GetAnnotations()
 	if annotations == nil {
 		annotations = map[string]string{}
 	}
 	annotations[auth.CreatorIDAnn] = request.UserInfo.Username
 	newObj.SetAnnotations(annotations)
-
-	return patch.CreatePatch(obj, newObj, response)
+	if err := patch.CreatePatch(obj.Raw, newObj, response); err != nil {
```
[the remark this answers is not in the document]
KevinJoiner: It really shouldn't be. I updated this code in a separate norman -> steve migration PR instead of this one because it focuses on the mutators that use this function. That said, the new PR won't be up this release, so I can move the changes here.

crobby (dismissed):

rmweir (approved):
LGTM

rmweir (dismissed):

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -29,6 +30,9 @@ func (m *mutator) Admit(response *webhook.Response, request *webhook.Request) er
 	if err != nil {
 		return err
 	}
-
-	return mutation.SetCreatorIDAnnotation(request, response, cluster, cluster.DeepCopy())
+	if err := mutation.SetCreatorIDAnnotation(request, response, request.Object, cluster.DeepCopy()); err != nil {
+		return fmt.Errorf("failed to set creatorIDAnnotation: %w", err)
+	}
+	response.Allowed = true
```
cmurphy, line 60: wonder how this could have worked before :thinking:

pkg/admission/admission.go
```diff
@@ -0,0 +1,244 @@
...
+// Request is a simple wrapper for an AdmissionRequest that includes the context from the original http.Request.
+type Request struct {
+	admissionv1.AdmissionRequest
+	Context context.Context
+}
+
+// NewDefaultValidationWebhook creates a new ValidatingWebhook based on the WebhookHandler provided.
+// The path set on the client config will be appended with the webhooks path.
+// The return webhook will not be nil.
+func NewDefaultValidationWebhook(handler WebhookHandler, clientConfig v1.WebhookClientConfig, scope v1.ScopeType) *v1.ValidatingWebhook {
```
cmurphy, line 78: nit - NewDefaultValidatingWebhook

pkg/admission/admission.go
```diff
@@ -0,0 +1,244 @@
...
+		SideEffects:             Ptr(v1.SideEffectClassNone),
+		TimeoutSeconds:          nil,
+		AdmissionReviewVersions: []string{"v1", "v1beta1"},
+	}
+}
+
+// NewDefaultMutationWebhook creates a new MutatingWebhook based on the WebhookHandler provided.
+// The path set on the client config will be appended with the webhooks path.
+// The return webhook will not be nil.
+func NewDefaultMutationWebhook(handler WebhookHandler, clientConfig v1.WebhookClientConfig, scope v1.ScopeType) *v1.MutatingWebhook {
```
cmurphy, line 95: nit - NewDefaultMutatingWebhook

pkg/admission/admission.go
```diff
@@ -0,0 +1,244 @@
...
+			sendError(responseWriter, review, err)
+			return
+		}
+
+		review.Response.UID = review.Request.UID
+		writeResponse(responseWriter, review)
+	}
+}
+
+// Ptr is a generic function that returns the pointer of a string.
```
cmurphy, line 204: :rocket:

pkg/resources/validation/feature/feature.go
```diff
@@ -3,42 +3,67 @@ package feature
...
+
+// Operations returns list of operations handled by this validator.
+func (v *Validator) Operations() []admissionregistrationv1.OperationType {
+	return []admissionregistrationv1.OperationType{admissionregistrationv1.Update}
+}
+
+// ValidatingWebhook returns the ValidatingWebhook used for this CRD.
+func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) *admissionregistrationv1.ValidatingWebhook {
+	valWebhook := admission.NewDefaultValidationWebhook(v, clientConfig, admissionregistrationv1.ClusterScope)
+	valWebhook.FailurePolicy = admission.Ptr(admissionregistrationv1.Ignore)
```
cmurphy, line 38: Was this the failure policy before this? I'm not finding it

cmurphy (commented):
One teensy question on the failure policy on features.management, otherwise this lgtm

pkg/resources/validation/feature/feature.go
KevinJoiner: yeah so, before we had two webhook configurations, one had a failure policy of ignore [here](https://github.com/rancher/webhook/blob/14683a296710e23bbf805a304be0241467d73c58/pkg/server/server.go#L168-#L176) and the rules which it applied to (cluster & features) are defined [here](https://github.com/rancher/webhook/blob/14683a296710e23bbf805a304be0241467d73c58/pkg/server/rules.go#L5-#L29).

pkg/resources/mutation/cluster/provisioningcluster.go
KevinJoiner: Before, SetCreatorIDAnnotation would also set `response.Allowed`, but this was moved out since setting an annotation does not guarantee that the request is valid for anyone else using the function.

KevinJoiner:
The latest push contains updates to function names.

pkg/resources/validation/cluster/cluster.go
```diff
@@ -3,37 +3,65 @@ package cluster
...
+}
+
+// Operations returns list of operations handled by this validator.
+func (v *Validator) Operations() []admissionregistrationv1.OperationType {
+	return []admissionregistrationv1.OperationType{admissionregistrationv1.Create, admissionregistrationv1.Update, admissionregistrationv1.Delete}
+}
+
+// ValidatingWebhook returns the ValidatingWebhook used for this CRD.
+func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) *admissionregistrationv1.ValidatingWebhook {
+	return admission.NewDefaultValidatingWebhook(v, clientConfig, admissionregistrationv1.ClusterScope)
```
cmurphy, line 47: Does this need to set the failure policy to ignore?

KevinJoiner: yes, good catch.

cmurphy (approved):

crobby (approved):

PR #151, opened 2022-11-18
Integration test

KevinJoiner:
This adds integration tests to our webhook.

Test steps
1. Build integration test binary and the rancher-webhook Docker image
2. Save the newly created docker image into `dist/artifacts`
3. Start a Rancher & k8s instance with the latest released rancher image.
4. Import the newly created rancher-webhook image into k8's containerd registry
5. Update the rancher webhook deployment to use the new rancher-webhook image.
6. Run the integration test.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,130 +1,144 @@
 ---
 kind: pipeline
 name: amd64
+type: docker
 
 platform:
   os: linux
   arch: amd64
 
 steps:
-- name: build
-  image: rancher/dapper:v0.4.1
-  commands:
-  - dapper ci
-  volumes:
-  - name: docker
-    path: /var/run/docker.sock
+  - name: build
+    image: rancher/dapper:v0.4.1
+    commands:
+      - dapper ci
+    volumes:
+      - name: docker
+        path: /var/run/docker.sock
 
-- name: github_binary_release
-  image: plugins/github-release
-  settings:
-    api_key:
-      from_secret: github_token
-    prerelease: true
-    checksum:
-    - sha256
-    checksum_file: CHECKSUMsum-amd64.txt
-    checksum_flatten: true
-    files:
-    - "dist/artifacts/*"
-  when:
-    instance:
-    - drone-publish.rancher.io
-    ref:
-    - refs/head/master
-    - refs/tags/*
-    event:
-    - tag
+  - name: integration-test
+    image: rancher/rancher:v2.7.0
+    privileged: true
+    commands:
+      - scripts/integration-test
 
-- name: docker-publish
-  image: plugins/docker
-  settings:
-    dockerfile: package/Dockerfile
-    password:
-      from_secret: docker_password
-    repo: "rancher/rancher-webhook"
-    tag: "${DRONE_TAG}-amd64"
-    username:
-      from_secret: docker_username
-  when:
-    instance:
-    - drone-publish.rancher.io
-    ref:
-    - refs/head/master
-    - refs/tags/*
-    event:
-    - tag
+  - name: github_binary_release
+    image: plugins/github-release
+    settings:
+      api_key:
+        from_secret: github_token
+      prerelease: true
+      checksum:
+        - sha256
+      checksum_file: CHECKSUMsum-amd64.txt
+      checksum_flatten: true
+      files:
+        - "dist/artifacts/*"
+    when:
+      instance:
+        - drone-publish.rancher.io
+      ref:
+        - refs/head/master
+        - refs/tags/*
+      event:
+        - tag
 
+  - name: docker-publish
+    image: plugins/docker
+    settings:
+      dockerfile: package/Dockerfile
+      password:
+        from_secret: docker_password
+      repo: "rancher/rancher-webhook"
+      tag: "${DRONE_TAG}-amd64"
+      username:
+        from_secret: docker_username
+    when:
+      instance:
+        - drone-publish.rancher.io
+      ref:
+        - refs/head/master
+        - refs/tags/*
+      event:
+        - tag
 volumes:
-- name: docker
-  host:
-    path: /var/run/docker.sock
+  - name: docker
+    host:
+      path: /var/run/docker.sock
 
 ---
 kind: pipeline
 name: arm64
+type: docker
 
 platform:
   os: linux
   arch: arm64
 
 steps:
-- name: build
-  image: rancher/dapper:v0.4.1
-  commands:
-  - dapper ci
-  volumes:
-  - name: docker
-    path: /var/run/docker.sock
+  - name: build
+    image: rancher/dapper:v0.4.1
+    commands:
+      - dapper ci
+    volumes:
+      - name: docker
+        path: /var/run/docker.sock
 
-- name: github_binary_release
-  image: plugins/github-release
-  settings:
-    api_key:
-      from_secret: github_token
-    prerelease: true
-    checksum:
-    - sha256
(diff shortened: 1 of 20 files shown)


scripts/integration-test
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+set -e
+
+export KUBECONFIG=
+export CATTLE_DEV_MODE=yes
+export CATTLE_SERVER_URL="https://$(ip route get 8.8.8.8 | awk '{print $7}'):443"
```
MbolotSuse, line 6: Nit: If you want to get rid of the awk I think you could use cut which appears to be installed by default with `bci:golang:1.19`

tests/integration/main_test.go
```diff
@@ -0,0 +1,592 @@
...
+		newObj:         newObj,
+		validCreateObj: validCreateObj,
+		invalidUpdate:  invalidUpdate,
+		validUpdate:    validUpdate,
+		validDelete:    validDelete,
+	}
+	validateEndpoints(m.T(), endPoints, m.clientFactory)
+}
+
+func (m *IntegrationSuite) TestGlobalRole() {
```
MbolotSuse, line 133: I would recommend splitting the test for various CRDs into separate files/packages of the integration test directory. The current structure sets a bad precedent for users who add integration tests in the future and may just expect to add more to a single file.

scripts/test
```diff
@@ -4,4 +4,9 @@ set -e
 cd $(dirname $0)/..
 
 echo Running tests
-go test -cover -tags=test ./...
+go test --coverpkg=./pkg/... -coverprofile=coverage.out --count=1 ./pkg/...
```
MbolotSuse, line 7: Near as I can tell, if you run `go test -cover` without specifying the other args it will ouptut the results in a pretty readable format (see [here](https://go.dev/blog/cover)). Can you elaborate on why you chose to output this to a file and then use awk to parse the file?

.drone.yml
```diff
@@ -1,130 +1,133 @@
...
 kind: pipeline
 name: amd64
+type: docker
 
 platform:
   os: linux
   arch: amd64
 
 steps:
-- name: build
```
MbolotSuse, line 10: Most of the changes to this file (excepting the `type: docker` and the `integration-test` pipeline) seem to be formatting changes. Was that the intention? If so, it may be helpful in the future to separate out larger formatting changes like this into their own commit - when they get mixed in with real changes like this it can be hard to understand the changes.

MbolotSuse (changes requested):
Really solid start on the integration tests - thanks for putting this together. Have a few minor suggestions .

.drone.yml
KevinJoiner: `Type: docker` was added because the old pipeline was working based on undefined drone behavior.
The [docs](https://docs.drone.io/pipeline/configuration/) don't state anywhere that `Type` can be set to nil and that docker will be used by default.

I am unsure if having a commit that just corrects our inconsistent file indentations is useful. Since GitHub can show the spacing change in the diff.

scripts/test
KevinJoiner: `go test -cover` only show the coverage of each individual package which isn't that useful. The changes here show the total coverage across the entire repo.

I am unsure if you are also asking why we are using a file rather than piping to awk, but if you are, that is because a coverprofile is not the same as stdout and holds more statistical information about the test that ran.
https://go.dev/blog/cover#viewing-the-results

scripts/integration-test
KevinJoiner: This instance of awk is run inside the rancher/rancher image, which already has awk installed. This code was copied from [rancher](https://github.com/rancher/rancher/blob/b8a90235e0fbba17fce53e0a03019a98325d93f9/scripts/run#L17).
 
I will leave the other usage of awk since it can not easily be done with cut due to the arithmetic operations used to calculate code coverage.

KevinJoiner:
Update the integration test to update the rancher-webhook deployments image with the one created during ci.

.drone.yml
MbolotSuse: > Type: docker was added because the old pipeline was working based on undefined drone behavior.
The [docs](https://docs.drone.io/pipeline/configuration/) don't state anywhere that Type can be set to nil and that docker will be used by default.

Ok - I assume that you are referring to the logic outlined [here](https://discourse.drone.io/t/type-parameter-is-not-required/9029). Seems like these may have been made in older versions of drone and needed an update. Thank you for doing that. 

> I am unsure if having a commit that just corrects our inconsistent file indentations is useful. Since GitHub can show the spacing change in the diff.

You don't need to make a change to accommodate me here. That being said, when reviewing it doesn't show line by line indentations very well. It shows massive blocks of read and green (e.x. the first block in this PR is 40 lines) - and I have to compare line by line to make sure that indentation was the only thing that changed. This is made especially important since you changed other things in this file, so I have no indicator of if anything in this section was a spacing/linting change or a functionality change. This takes more time, and makes it harder to review.

scripts/test
MbolotSuse: Resolving - complete coverage could have it's uses eventually and since we don't want to introduce go/perl into the pipeline we are going to stick with awk.

.drone.yml
KevinJoiner: ahh, I see maybe I can split for a review and then squash the commit after?

MbolotSuse (commented):
One significant concern on the saved docker image being included as part of the pre-release. Outside of that, just a couple of nits (up to you on if you want to fix those).

pkg/health/health.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Name returns the Name of the checker.
+func (e *ErrorChecker) Name() string { return e.name }
+
+// Check returns the last error stored.
+func (e *ErrorChecker) Check(_ *http.Request) error {
+	e.mutex.RLock()
+	defer e.mutex.RUnlock()
+	retError := e.lastError
```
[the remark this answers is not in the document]
KevinJoiner: yeah, we can I was using atomic.Value before which caused the needed this extra line.

tests/integration/main_test.go
```diff
@@ -0,0 +1,193 @@
...
+	defer cancel()
+
+	if objs.invalidCreate != nil {
+		invalidObj := objs.invalidCreate()
+		err = client.Create(ctx, invalidObj.GetNamespace(), invalidObj, nil, v1.CreateOptions{})
+		assert.Error(t, err, "No error returned during the creation of an invalid Object")
+	}
+	err = client.Create(ctx, objs.validCreateObj.GetNamespace(), objs.validCreateObj, result, v1.CreateOptions{})
+	assert.NoError(t, err, "Error returned during the creation of a valid Object")
+	if objs.invalidUpdate != nil {
```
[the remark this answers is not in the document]
KevinJoiner: My thought was that you can't test and update without an object being there, so you will always need a create call of some sort. The other thought is that other tests in the future do not have to use the `validateEndpoints` function it is only a nice to have function for common endpoint testing. If you look at the Secret test, it does not use the validate function.

I can still leave a comment to indicate that there is an order of operations for future users.

.drone.yml
MbolotSuse: Yeah - you could do that. I'm would also be ok with you just leaving the commit as separate. All good though, we don't need to do anything for this PR right now.

MbolotSuse (approved):
LGTM

crobby (approved):
lgtm, runs great for me

PR #156, opened 2022-12-01
Fix for 2.5 pipeline on github release phase

MbolotSuse:
Drone CI is currently broken on the github release phase. It looks like the github release plugin removed support for arm v32 (see https://github.com/drone-plugins/drone-github-release/pull/143), so we are pinning to a version which recently worked, for this phase.


KevinJoiner (approved):
LGTM

PR #157, opened 2022-12-14
PodSecurityAdmissionConfigurationTemplate validation

HarrisonWAffel:
### Problem:
A new CRD was introduced (`PodSecurityAdmissionConfigurationTemplate`), but due to the current transition from the `v3` API to the `v1` API no validation could be written within Norman. 

### Solution: 
Add the missing validation to the web-hook inline with how other `V1` API objects perform validation.

In situations where a resource may have multiple issues with its configuration, only one error will be returned at a time. Including the description of many errors at once in the same response resulted in a less than readable result which would not display well in the UI. 

In situations where a template may be in use by either a `v1.Cluster` or `v3.Cluster` the web-hook will prevent the template from being deleted. Additionally, you cannot delete the default built-in templates `restricted` & `privileged` 

### Testing: 
A large amount of unit tests have been written, but these steps can be used for manual validation as well. 

A. Object Validation testing 
1. Create a new yaml file with the following contents 
```yaml
kind: PodSecurityAdmissionConfigurationTemplate
metadata:
  name: testing
  namespace: testing
configuration:
  defaults:
    enforce: "privileged"
    enforce-version: "latest"
    audit: "restricted"
    audit-version: "latest"
    warn: "privileged"
    warn-version: "latest"
  exemptions:
    usernames: ["abc123"]
    runtimeClasses: ["abs"]
    namespaces: []
```
2. Create the YAML file using `kubectl`
3. Alter the YAML file in various ways to ensure validation occurs
    a. All defaults must be either 'privileged', 'baseline', or 'restricted' 
    b. All versions must be semver (`v1.x`) or `latest`
    c. All Exemptions cannot contain duplicate values 
    d. runtime classes and namespaces must be a RFC 1123 subdomain (cannot start with `-`)

B. Object deletion testing 

1. Create a new PodSecurityAdmissionConfigurationTemplate using the above YAML
2. Create an RKE1 cluster and an RKE2 cluster 
3. For both RKE1 and RKE2, Modify the cluster YAML to include the `defaultPodSecurityAdmissionConfigurationTemplateName` field directly under the spec and set its value equal to `testing` (or the name of the PSACT you created)
4. Attempt to delete the template and verify that the web-hook prevents you due to it being used by a cluster


jiaqiluo:
> Should we prevent users from deleting the default PSACT's?

Yes, we should.

Also, we should prevent the deletion of the PSACT if any downstream cluster uses it.

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,295 @@
...
+const ByPodSecurityAdmissionConfigurationName = "podSecurityAdmissionConfigurationName"
+
+func NewPodSecurityAdmissionConfigurationTemplateValidator(clients *clients.Clients) *Validator {
+	return &Validator{
+		ManagementClusterCache:   clients.Management.Cluster().Cache(),
+		provisioningClusterCache: clients.Provisioning.Cluster().Cache(),
+	}
+}
+
+func (v Validator) ValidatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) *admissionregistrationv1.ValidatingWebhook {
```
jiaqiluo, line 47:
```suggestion
func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) *admissionregistrationv1.ValidatingWebhook {
```
This change needs to be expanded to all other methods associated to  the struct

jiaqiluo (changes requested):

jiaqiluo (approved):
LGTM 💯

pkg/generated/controllers/management.cattle.io/v3/podsecurityadmissionconfigurationtemplate.go
```diff
@@ -0,0 +1,256 @@
...
+
+	Delete(namespace, name string, options *metav1.DeleteOptions) error
+	Get(namespace, name string, options metav1.GetOptions) (*v3.PodSecurityAdmissionConfigurationTemplate, error)
+	List(namespace string, opts metav1.ListOptions) (*v3.PodSecurityAdmissionConfigurationTemplateList, error)
+	Watch(namespace string, opts metav1.ListOptions) (watch.Interface, error)
+	Patch(namespace, name string, pt types.PatchType, data []byte, subresources ...string) (result *v3.PodSecurityAdmissionConfigurationTemplate, err error)
+}
+
+type PodSecurityAdmissionConfigurationTemplateCache interface {
+	Get(namespace, name string) (*v3.PodSecurityAdmissionConfigurationTemplate, error)
```
jiaqiluo, line 67: @HarrisonWAffel 
I just realized that `PodSecurityAdmissionConfigurationTemplate` should be a **non-namespaced** resource type. Is it right? 😄 

The fix should be in rancher/rancher: we need to add `// +genclient:nonNamespaced` to where the struct is defined. 

example: https://github.com/jiaqiluo/rancher/blob/ae44c398377516fcd150b8fc2a81a9fd6d4d8fb5/pkg/apis/management.cattle.io/v3/authz_types.go#L128-L128

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	provClusters, err := v.provisioningClusterCache.GetByIndex(ByPodSecurityAdmissionConfigurationName, oldTemplate.Name)
+	if reportDeletionError(len(provClusters), err, "provisioning", oldTemplate.Name, resp) {
+		return
+	}
+
+}
+
+func (v *Validator) ValidateConfiguration(configurationTemplate *mgmtv3.PodSecurityAdmissionConfigurationTemplate, resp *admissionv1.AdmissionResponse) {
```
crobby, line 133: does it need to be exported?

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact_test.go
```diff
@@ -0,0 +1,631 @@
...
+		},
+	}
+
+	m.Run()
+}
+
+func TestHandleDeletion(t *testing.T) {
+	for _, testcase := range deleteTests {
+		t.Run(testcase.testName, func(t *testing.T) {
+			// normally the `Admit` handler sets the final allowed
```
crobby, line 470: Why not test Admit directly?  It is the primary exported function.

crobby (changes requested):
A couple comments, a few questions.  Seems pretty sane.

crobby (changes requested):
Also, look into adding an integration test.  The framework for them is pretty new, but it works pretty well.

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,295 @@
...
+	oldTemplate, newTemplate, err := objectsv3.PodSecurityAdmissionConfigurationTemplateOldAndNewFromRequest(&req.AdmissionRequest)
+	if err != nil {
+		return resp, err
+	}
+
+	if req.Operation == admissionv1.Create || req.Operation == admissionv1.Update {
+		v.ValidateConfiguration(newTemplate, resp)
+		if resp.Result != nil {
+			return resp, nil
+		}
```
KevinJoiner, line 92: It is not a great paradigm to pass in a variable `response` to be mutated and then checked for specific mutations. `ValidateConfiguration` should just return an error if any and then the response can be created with that error and returned. 

```suggestion
		errList := v.ValidateConfiguration(newTemplate)
		if errList.ToAggregate() != nil {
			return &admissionv1.AdmissionResponse{
				Result: &metav1.Status{
					Status:  "Failure",
					Message: errList.ToAggregate().Error(),
					Reason:  metav1.StatusReasonBadRequest,
					Code:    http.StatusUnprocessableEntity,
				},
			}, nil
		}
```

An alternative approach if you do not want the `ErrorList` logic this high up in `Admit(...)` is to rename `v.ValidateConfiguration(newTemplate)` to `handleCreateAndUpdate` and just return the response and and error.
```suggestion
		return v.handleCreateAndUpdate(newTemplate)
```

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	if req.Operation == admissionv1.Delete {
```
KevinJoiner, line 95: `else if `or covert this to a `select` statement.
```suggestion
	} else if req.Operation == admissionv1.Delete {
```

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,295 @@
...
+			Code:    http.StatusInternalServerError,
+		}
+		resp.Allowed = false
+		return true
+	}
+
+	if clustersUsingTemplate > 0 {
+		message := fmt.Sprintf("Cannot delete template '%s' as it is being used by %d %s clusters", templateName, clustersUsingTemplate, clusterType)
+		if clustersUsingTemplate == 1 {
+			message = fmt.Sprintf("Cannot delete template '%s' as it is being used by a %s cluster", templateName, clusterType)
```
KevinJoiner, line 187:
```suggestion
			message = fmt.Sprintf("Cannot delete template '%s' as it is being used by the '%s' cluster", templateName, clusterType)
```

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,295 @@
...
+	return false
+}
+
+func reportError(err field.ErrorList, resp *admissionv1.AdmissionResponse) bool {
+	if err.ToAggregate() == nil {
+		return false
+	}
+	resp.Result = &metav1.Status{
+		Status:  "Failure",
+		Message: fmt.Sprintf("%s", err.ToAggregate()),
```
KevinJoiner, line 208:
```suggestion
		Message: err.ToAggregate().Error(),
```

KevinJoiner (changes requested):

pkg/clients/clients.go
```diff
@@ -51,6 +59,7 @@ func New(ctx context.Context, rest *rest.Config, mcmEnabled bool) (*Clients, err
 	result := &Clients{
 		Clients:                *clients,
 		Management:             mgmt.Management().V3(),
+		Provisioning:           prov.Provisioning().V1(),
```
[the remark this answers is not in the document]
KevinJoiner: I checked on this and provisioning CRDs are present so this should be fine.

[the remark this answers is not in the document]
HarrisonWAffel: @KevinJoiner Thanks! I was looking into this last night and was thinking the same. The provisioning CRD's are created when the `ProvisioningV2` flag is enabled, which is it by default. That flag also cannot be turned off in the UI. Just for my understanding, when we are working with CRD's that _can_ be disabled by the UI we should add a new argument to the webhook & its chart to specify if its enabled or not, right?

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
HarrisonWAffel: There can be many management and provisioning clusters, and because we don't include the name of the cluster I think `a` is the correct wording.

pkg/generated/controllers/management.cattle.io/v3/podsecurityadmissionconfigurationtemplate.go
HarrisonWAffel: This struct actually isn't needed to accomplish this validation, so it has been removed

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,301 @@
...
+	resp := &admissionv1.AdmissionResponse{}
+	oldTemplate, newTemplate, err := objectsv3.PodSecurityAdmissionConfigurationTemplateOldAndNewFromRequest(&req.AdmissionRequest)
+	if err != nil {
+		return resp, fmt.Errorf("failed to parse PodSecurityAdmissionConfigurationTemplate object from request:%w", err)
+	}
+
+	switch req.Operation {
+	case admissionv1.Update:
+		fallthrough
+	case admissionv1.Create:
```
jiaqiluo, line 97: I think this is what you mean here: do the same set of validations for both the event of create and update. 
```suggestion
	case admissionv1.Create, admissionv1.Update:
```

jiaqiluo (changes requested):

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
HarrisonWAffel: Thanks for the tip! I've changed it to no longer use the `fallthrough`

jiaqiluo (approved):

crobby (approved):
It would still be good to add integration tests, but what's here lgtm.

pkg/clients/clients.go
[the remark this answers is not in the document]
KevinJoiner: I think you are right, and this should be supported. We should probably create a separate issue for that feature.

pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go
```diff
@@ -0,0 +1,299 @@
...
+		err = v.validateConfiguration(newTemplate)
+		if err != nil {
+			resp.Result = &metav1.Status{
+				Status:  "Failure",
+				Message: err.Error(),
+				Reason:  metav1.StatusReasonBadRequest,
+				Code:    http.StatusUnprocessableEntity,
+			}
+			resp.Allowed = false
+			break
```
KevinJoiner, line 105: nit: for some of these cases it appears we can just return early instead of breaking.
```suggestion
			return resp, nil
```

KevinJoiner (approved):

PR #160, opened 2022-12-20
Bump kube-version to support k8s 1.25

cmurphy:
https://github.com/rancher/rancher/issues/39854


cbron (approved):
lgtm,  error is unrelated

PR #163, opened 2022-12-21
Update webhook deployment to require rancher-webhook-tls only when capi is enabled

crobby:
Since rancher-webhook will take care of its own cert generation, we only need to include the secret/volume for use by capi when capi is enabled.

### DIFF
--- a/charts/rancher-webhook/templates/deployment.yaml
+++ b/charts/rancher-webhook/templates/deployment.yaml
@@ -11,10 +11,12 @@ spec:
       labels:
         app: rancher-webhook
     spec:
+      {{- if .Values.capi.enabled }}
       volumes:
       - name: tls
         secret:
           secretName: rancher-webhook-tls
+      {{- end }}
       {{- if .Values.global.hostNetwork }}
       hostNetwork: true
       {{- end }}
@@ -46,9 +48,11 @@ spec:
           containerPort: 9443
         - name: capi-https
           containerPort: 8777
+        {{- if .Values.capi.enabled }}
         volumeMounts:
         - name: tls
           mountPath: /tmp/k8s-webhook-server/serving-certs
+        {{- end }}
       serviceAccountName: rancher-webhook
       {{- if .Values.priorityClassName }}
       priorityClassName: "{{.Values.priorityClassName}}"


cmurphy (commented):
It looks like this is mostly a revert of a11cc91efb51d9e062229c7e5341df752095f05d, could you explain more why the secret name was changed?

charts/rancher-webhook/templates/deployment.yaml
```diff
@@ -14,7 +14,7 @@ spec:
       volumes:
       - name: tls
         secret:
-          secretName: rancher-webhook-tls
+          secretName: cattle-webhook-tls
```
maxsokolovsky, line 17: Why the `cattle` in the secret name? I am not seeing it in previous webhook versions, it's always `rancher-webhook-tls`.

crobby:
> It looks like this is mostly a revert of [a11cc91](https://github.com/rancher/webhook/commit/a11cc91efb51d9e062229c7e5341df752095f05d), could you explain more why the secret name was changed?

I think my previous attempt was misguided and my testing of the change wound-up not really testing the change.  Previously, the rancher-webhook-tls was only used by the capi service...I thought it could also be repurposed for rancher-webhook....I was wrong.  Looking at the rancher-webhook source revealed the following reveals what the rancher-webhook is using:  https://github.com/rancher/webhook/blob/release/v0.3/pkg/server/server.go#L30

I think the previous code was wrong, but not in a way that interfered with things working correctly.  The rancher-webhook deployment was set up to expect the rancher-webhook-tls secret, but never actually used it (only used by capi afaict).  It wasn't a problem because the webhook never used to run downstream (and therefore ran separately from capi).

It all went bad when I moved the need-a-cert annotation to the rancher-webhook service because down in the wrangler code that does the secret/cert generation, it also changes the caBundle  https://github.com/rancher/wrangler/blob/master/pkg/needacert/needacert.go#L168

charts/rancher-webhook/templates/deployment.yaml
crobby: https://github.com/rancher/webhook/blob/release/v0.3/pkg/server/server.go#L30

crobby:
Ok, I've now tested this on both local and downstream clusters, with and without capi enabled and I think this is correct.

maxsokolovsky (approved):

cmurphy (approved):

PR #165, opened 2022-12-23
add the support for using PodSecurityAdmissionConfigurationTemplate in RKE1 cluster 1.25 and above

jiaqiluo:
**Issue:** 
https://github.com/rancher/rancher/issues/39992
https://github.com/rancher/rancher/issues/40009

Additionally, this also partially addresses https://github.com/rancher/rancher/issues/39995 by implementing the check on webhook side (the deeper check on RKE1 side is implemented as part of https://github.com/rancher/rke/pull/3132).

**Description:**
This PR is a part of the effort to support using PSACT (Pod Security Admission Configuration Template) in the RKE1 1.25 cluster. 

The PR introduces the following major changes:
- add a new mutator for the v3 management Cluster  ( see the file `pkg/resources/mutation/cluster/cluster.go`) 
- update the existing validator for the v3 management Cluster to validate the PSP-related and PSACT-related fields in the cluster  (see the file `pkg/resources/validation/cluster/cluster.go`)

 Related PRs:
- https://github.com/rancher/rancher/pull/40006 to fix a panic error  in rancher/rancher caused by changes in PR


go.mod
```diff
@@ -59,14 +61,12 @@ require (
 	github.com/Masterminds/sprig/v3 v3.2.2 // indirect
 	github.com/PuerkitoBio/purell v1.1.1 // indirect
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
-	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20210826220005-b48c857c3a0e // indirect
```
sbstp, line 62: Surprised to see antlr in here, I wonder which component needs this

sbstp: I did some digging and found that it comes from https://github.com/google/cel-go

pkg/resources/mutation/cluster/cluster.go
```diff
@@ -0,0 +1,132 @@
...
+		return psa.AdmissionResponseAllowed(), nil
+	}
+	newTemplateName := newCluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+	oldTemplateName := oldCluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+
+	// If the template is set(or changed), update the cluster with the new template's content
+	if newTemplateName != "" {
+		err := m.setPSAConfig(newCluster)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to set PSAconfig: %w", err)
```
crobby, line 76: Any chance this error message could be more specific than "failed to set..."?

crobby (commented):
Might be useful to add unit tests for the functions in podsecurityadmission.go

pkg/resources/mutation/cluster/cluster.go
```diff
@@ -0,0 +1,132 @@
...
+	plugin, err := psa.GetPluginConfigFromTemplate(template)
+	if err != nil {
+		return fmt.Errorf("failed to get plugin config from template: %w", err)
+	}
+	admissionConfig := psa.GetAdmissionConfigFromCluster(cluster)
+	found := false
+	for i, item := range admissionConfig.Plugins {
+		if item.Name == "PodSecurity" {
+			admissionConfig.Plugins[i] = plugin
+			found = true
```
crobby, line 123: maybe break from the loop after it's found?

jiaqiluo: the `break` is added to save some iterations 😄

pkg/patch/patch.go
```diff
@@ -23,7 +23,11 @@ func CreatePatch(oldJSON []byte, newObj interface{}, response *v1.AdmissionRespo
 	if err != nil {
 		return fmt.Errorf("failed to marshal generated patch to JSON: %w", err)
 	}
-
+	// json.Marshal() function treats an empty slice as "[]" and a nil pointer as "null" when performing the marshal,
```
snasovich, line 26: I think it would be cleaner/more efficient to just ensure `patch.Patches` is non-nil and non-empty slice by returning nil if `len(patch.Patches) == 0` before even marshalling it and checking string responses.

pkg/podsecurityadmission/podsecurityadmission.go
```diff
@@ -0,0 +1,114 @@
...
+		return plugin, fmt.Errorf("failed to marshal PodSecurityConfiguration: %w", err)
+	}
+	plugin.Configuration.Raw = cBytes
+	return plugin, nil
+}
+
+// GetPlugConfigFromCluster generates a PluginConfig for PodSecurity from a Cluster,
+// or a new one with default values if the cluster does not have one.
+// True is returned if a PluginConfig is found in the cluster.
+func GetPlugConfigFromCluster(cluster *apisv3.Cluster) (apiserverv1.AdmissionPluginConfiguration, bool) {
```
snasovich, line 67: `GetPlugConfigFromCluster` -> `GetPluginConfigFromCluster`

pkg/podsecurityadmission/podsecurityadmission.go
```diff
@@ -0,0 +1,114 @@
...
+	return apiserverv1.AdmissionPluginConfiguration{
+		Name: "PodSecurity",
+		Configuration: &runtime.Unknown{
+			ContentType: "application/json",
+		},
+	}, false
+}
+
+// DropPSAPlugConfigFromAdmissionConfig removes the PluginConfig for PodSecurity from a Cluster if it has one.
+func DropPSAPlugConfigFromAdmissionConfig(cluster *apisv3.Cluster) {
```
snasovich, line 83: `DropPSAPlugConfigFromAdmissionConfig` -> `DropPSAPluginConfigFromAdmissionConfig`

snasovich (commented):
Couple minor comments, started reviewing it and now need to take care of something else. Planning to resume reviewing later.

pkg/patch/patch.go
jiaqiluo: Thanks for this good idea! 
Code is updated.

pkg/podsecurityadmission/podsecurityadmission.go
jiaqiluo: fixed

pkg/podsecurityadmission/podsecurityadmission.go
jiaqiluo: fixed

pkg/resources/validation/cluster/cluster.go
```diff
@@ -93,16 +154,136 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
+			// under kube-api.admission_configuration at the same time.
+			if oldTemplateName != "" {
+				newConfig, found := psa.GetPluginConfigFromCluster(newCluster)
+				if !found {
+					// not found means the kube-api.admission_configuration section is also removed, which is good
+					return psa.AdmissionResponseAllowed(), nil
+				}
+				oldConfig, _ := psa.GetPluginConfigFromCluster(oldCluster)
+				if reflect.DeepEqual(newConfig, oldConfig) {
+					msg := "The Plugin Config for PodSecurity under kube-api.admission_configuration is the same as the previously-set PodSecurityAdmissionConfigurationTemplate." +
```
snasovich, line 202: I'm not sure why we should not allow setting `kube-api.admission_configuration` explicitly to the same value as the configuration per previously set PSACT. Maybe users just want to "untie" from the configuration template.
Maybe I'm missing a requirement here.

pkg/resources/validation/cluster/cluster.go
```diff
@@ -93,16 +154,136 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
-func toExtra(extra map[string]authenticationv1.ExtraValue) map[string]v1.ExtraValue {
-	result := map[string]v1.ExtraValue{}
-	for k, v := range extra {
-		result[k] = v1.ExtraValue(v)
+// checkPSAConfigOnCluster validates the cluster spec when DefaultPodSecurityAdmissionConfigurationTemplateName is set
+func (v *Validator) checkPSAConfigOnCluster(cluster *apisv3.Cluster) (*admissionv1.AdmissionResponse, error) {
+	// validate that extra_args.admission-control-config-file is not set at the same time
+	_, found := cluster.Spec.RancherKubernetesEngineConfig.Services.KubeAPI.ExtraArgs["admission-control-config-file"]
+	if found {
+		msg := "could not use external admission control configuration file when using PodSecurityAdmissionConfigurationTemplate"
```
snasovich, line 217: Nit, but why do we set `msg` var instead of just returning `psa.AdmissionResponseBadRequest("...message...")`?

pkg/resources/validation/cluster/cluster.go
```diff
@@ -93,16 +154,136 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
+		return psa.AdmissionResponseBadRequest(msg), nil
 	}
-	return result
+	// validate that the configuration for PodSecurityAdmission under the kube-api.admission_configuration section
+	// matches the content of the PodSecurityAdmissionConfigurationTemplate specified in the cluster
+	name := cluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+	template, err := v.psact.Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
snasovich, line 225: What if PSACT used by this cluster got deleted? This would prevent any updates to the cluster.
Maybe we need to further discuss it, but I thought that after PSACT is specified we basically copy its contents down and that relationship to PSACT is severed and we no longer synchronize updates to PSACT down to all clusters that use it - but maybe I'm wrong here.

pkg/resources/validation/cluster/cluster.go
```diff
@@ -93,16 +154,136 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
+		return nil, fmt.Errorf("failed to unmarshal PodSecurityConfiguration: %w", err)
+	}
+
+	psaConfig2 := &psav1.PodSecurityConfiguration{}
+	err = json.Unmarshal(fromAdmissionConfig.Configuration.Raw, psaConfig2)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal PodSecurityConfiguration: %w", err)
+	}
+	if !reflect.DeepEqual(psaConfig, psaConfig2) {
+		msg := "PodSecurity Configuration under kube-api.admission_configuration does not match " +
```
snasovich, line 251: Same as previous comment - what if PSACT got modified _after_ it was applied to the cluster - we will start hitting this error on any update to the cluster (unless mutating webhook is guaranteed to act first and ensures the contents match?).

pkg/resources/validation/cluster/cluster.go
```diff
@@ -93,16 +154,136 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
+		return psa.AdmissionResponseBadRequest(msg), nil
+	}
+
+	return psa.AdmissionResponseAllowed(), nil
+}
+
+func getClusterVersion(version string) (semver.Version, error) {
+	var parsedVersion semver.Version
+	if len(version) <= 1 || !strings.HasPrefix(version, "v") {
+		return parsedVersion, fmt.Errorf("%s is not valid version", version)
```
snasovich, line 282: Nit:
```suggestion
		return parsedVersion, fmt.Errorf("%s is not valid k8s version", version)
```

pkg/generated/objects/management.cattle.io/v3/objects.go
```diff
@@ -220,6 +220,59 @@ func FleetWorkspaceFromRequest(request *admissionv1.AdmissionRequest) (*v3.Fleet
 	return object, nil
 }
 
+// PodSecurityAdmissionConfigurationTemplateOldAndNewFromRequest gets the old and new PodSecurityAdmissionConfigurationTemplate objects, respectively, from the webhook request.
+// If the request is a Delete operation, then the new object is the zero value for PodSecurityAdmissionConfigurationTemplate.
+// Similarly, if the request is a Create operation, then the old object is the zero value for PodSecurityAdmissionConfigurationTemplate.
+func PodSecurityAdmissionConfigurationTemplateOldAndNewFromRequest(request *admissionv1.AdmissionRequest) (*v3.PodSecurityAdmissionConfigurationTemplate, *v3.PodSecurityAdmissionConfigurationTemplate, error) {
```
snasovich, line 226: I'm not seeing usage of this (or next) func. Am I missing implementation for validation/mutation for PSACTs?

pkg/resources/mutation/cluster/cluster.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to set PSAconfig: %w", err)
+		}
+	} else {
+		switch request.Operation {
+		case admissionv1.Create:
+			return psa.AdmissionResponseAllowed(), nil
+		case admissionv1.Update:
+			// It is a valid use case where user switches from using PSACT to putting a PluginConfig for PSA under kube-api.AdmissionConfiguration,
+			// but it is not a valid use case where the PluginConfig for PSA has the same content as the one in the previous-set PSACT,
```
snasovich, line 84: Same as in the other comment, I'm not sure why _"it is not a valid use case where the PluginConfig for PSA has the same content as the one in the previous-set PSACT"_. Are not allowing the users to "untie" from PSACT and just specify the same content under `kube-api.AdmissionConfiguration`?

snasovich (changes requested):
Added some comments/questions/suggestions.

pkg/generated/objects/management.cattle.io/v3/objects.go
jiaqiluo: It is the next function not being used. 
This is a generated file, so we can not do much here.

snasovich: Ugh, my bad about generated code. 😞

pkg/resources/mutation/cluster/cluster.go
jiaqiluo: Yes, the current design is that users can not specify the same content under `kube-api.AdmissionConfiguration` as the previously-set PSACT. This is because we cannot tell if the content under `kube-api.AdmissionConfiguration` is from the previously-set PSACT or users.  Either keeping it when it is from the PSACT or dropping it when it is from the users is not good. 

To solve this problem, we can take the idea from how we support PSA today: adding a setting for enabling/disabling using PSACT in the cluster. However, this also does not provide a much better user experience: users have to configure one more field.

pkg/resources/validation/cluster/cluster.go
jiaqiluo: nothing special but to avoid a long return line, but I feel it does not help much.

pkg/resources/validation/cluster/cluster.go
jiaqiluo: 1. This check confirms that the content under the `kube-api.admission_configuration` matches the specified PSACT; therefore, we need to get the PSACT. 
2. To avoid the event you concerned, we do not allow deletion of a PSACT if any cluster uses it, which has been added to the webhook recently [here](https://github.com/rancher/webhook/blob/2b9e6c6301e489122b034101b9e7ed9ae8f7106e/pkg/resources/validation/podsecurityadmissionconfigurationtemplate/podsecurityact.go#L134-L148)

pkg/resources/validation/cluster/cluster.go
jiaqiluo: > unless mutating webhook is guaranteed to act first and ensures the contents match
Yes, this is guaranteed

pkg/resources/validation/cluster/cluster.go
```diff
@@ -51,16 +68,60 @@ func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.Webho
...
+		return nil, fmt.Errorf("failed to get old and new clusters from request: %w", err)
 	}
 
-	if newCluster.Spec.FleetWorkspaceName == "" ||
-		oldCluster.Spec.FleetWorkspaceName == newCluster.Spec.FleetWorkspaceName {
-		return &admissionv1.AdmissionResponse{
-			Allowed: true,
-		}, nil
+	if newCluster.Spec.FleetWorkspaceName == "" || oldCluster.Spec.FleetWorkspaceName == newCluster.Spec.FleetWorkspaceName {
+		return psa.AdmissionResponseAllowed(), nil
```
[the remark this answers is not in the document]
jiaqiluo: the function `AdmissionResponseAllowed` and `AdmissionResponseBadRequest` are moved to a new file that holds all the utility functions.

pkg/resources/validation/cluster/cluster.go
```diff
@@ -51,16 +68,60 @@ func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.Webho
...
+	response, err := v.validateFleetPermissions(request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to validate fleet permissions: %w", err)
+	}
+	if !response.Allowed {
+		return response, nil
+	}
+
+	if request.Operation == admissionv1.Create || request.Operation == admissionv1.Update {
+		_, newCluster, err := objectsv3.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
```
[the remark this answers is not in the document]
jiaqiluo: Yes, we can. The code is uppdated.

pkg/resources/validation/cluster/cluster.go
```diff
@@ -93,16 +154,136 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get old and new clusters from request: %w", err)
+	}
+	newTemplateName := newCluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+	oldTemplateName := oldCluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+	parsedVersion, err := getClusterVersion(newCluster.Spec.RancherKubernetesEngineConfig.Version)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse cluster version: %w", err)
+	}
+	if parsedRangeLessThan125(parsedVersion) && newTemplateName != "" {
```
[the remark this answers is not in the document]
jiaqiluo: fixed!

pkg/resources/validation/cluster/cluster.go
jiaqiluo: fixed.

pkg/resources/validation/cluster/cluster.go
jiaqiluo: please see my reply [here](https://github.com/rancher/webhook/pull/165#discussion_r1061045288)

snasovich (approved):
@jiaqiluo , thank you for addressing my comments. LGTM based on the previous review.
I'm approving it with the assumption that the second review will not be a rubber-stamp one. :)

pkg/resources/utilities.go
```diff
@@ -0,0 +1,30 @@
+// Package resources contains utility functions which are useful for both mutator for validator
```
KevinJoiner, line 1: These functions fit well in the pkg/admission package.
And the names could then be updated to ResponseAllowed and ResponseBadRequest

pkg/resources/validation/cluster/cluster.go
```diff
@@ -51,13 +71,59 @@ func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.Webho
...
+		return nil, fmt.Errorf("failed to validate fleet permissions: %w", err)
+	}
+	if !response.Allowed {
+		return response, nil
+	}
+
+	if request.Operation == admissionv1.Create || request.Operation == admissionv1.Update {
+		cluster, err := objectsv3.ClusterFromRequest(&request.AdmissionRequest)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get old and new clusters from request: %w", err)
```
KevinJoiner, line 85:
```suggestion
			return nil, fmt.Errorf("failed to get new clusters from request: %w", err)
```

pkg/resources/utilities.go
jiaqiluo: good idea! 
fixed

pkg/resources/validation/cluster/cluster.go
jiaqiluo: fixed

KevinJoiner (approved):

PR #166, opened 2023-01-04
Go generate.

KevinJoiner:
Updates the Copywrite header in our generated files to 2023.


crobby (approved):
lgtm.  Is there something we can tweak so that we don't run into this every 1/1/XX?

KevinJoiner:
The only thing would be to add an automated pipeline that runs at the end of the year that creates this PR, but I think that may be more trouble than it's worth.

MbolotSuse (approved):
lgtm. Thanks for changing this.

PR #169, opened 2023-01-05
fix the mutation and validation in v3 mgmt cluster

jiaqiluo:


KevinJoiner (approved):

pkg/resources/validation/cluster/cluster.go
```diff
@@ -82,12 +82,10 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
 		if err != nil {
 			return nil, fmt.Errorf("failed to get cluster from request: %w", err)
 		}
-		if cluster.Name == "local" {
+		// no need to validate the local cluster, or imported cluster which represents a KEv2 cluster (GKE/EKS/AKS) or v1 Provisioning Cluster
+		if cluster.Name == "local" || cluster.Spec.RancherKubernetesEngineConfig == nil {
 			return admission.ResponseAllowed(), nil
 		}
-		if cluster.Spec.RancherKubernetesEngineConfig == nil {
```
jakefhyde, line 88: Is it possible to check the cluster driver here or is it not populated yet? I'm going to approve so as to unblock CI that's failing because of this.

jakefhyde (approved):

pkg/resources/validation/cluster/cluster.go
jiaqiluo: It is doable but unnecessary because we need to skip it anyway.

PR #171, opened 2023-01-06
Remove the PSP and references to it from the chart

maxsokolovsky:
Issue: https://github.com/rancher/rancher/issues/39733


MbolotSuse (commented):
It appears that the pre-delete job is going to attempt to run as user 0 (see [here](https://github.com/rancher/webhook/blob/72d2039f1472822b20afa685a6add252c583f2d9/charts/rancher-webhook/templates/pre-delete-hook-job.yaml#L36-L37)). With this PSP being removed, this job will not be able to run in hardened clusters. 

Do we know how this is intended to work for those scenarios?

crobby:
> It appears that the pre-delete job is going to attempt to run as user 0 (see [here](https://github.com/rancher/webhook/blob/72d2039f1472822b20afa685a6add252c583f2d9/charts/rancher-webhook/templates/pre-delete-hook-job.yaml#L36-L37)). With this PSP being removed, this job will not be able to run in hardened clusters.
> 
> Do we know how this is intended to work for those scenarios?

Are clusters going to be set to include an exemption for the cattle-system namespace?  I thought that was the plan to still allow rancher to do things that might be out of bounds otherwise.

crobby (approved):
lgtm

macedogm:
> @MbolotSuse, that depends on how the hardened cluster's setup changes with PSPs being removed, in general.
> I don't know which exact mechanisms the use will have to be hardened. PSAs or something else? @macedogm

From v1.25 on we will have to rely on PSS/PSA now, which will require an update to our hardening guides.

MbolotSuse (approved):
Based on @macedogm 's comment about `cattle-system` needing an exception, I think we can move ahead with this.

PR #176, opened 2023-01-11
Run the handler for PSACT only when MCM is enabled

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/39365

Problem:
The handler for PSACT(PodSecurityAdmissionConfiguraTiontemplate) runs when the rancher-webhook app is installed in a downstream cluster, which causes the degradation of RBAC-related validation.  

By design, The handler for PSACT(PodSecurityAdmissionConfiguraTiontemplate) should run only in the local cluster, and it needs to get the v1 provisioning Cluster objects to perform its validations.  

Fix:
Move the handler to the correct section, so it runs only when MultiClusterManagement is enabled which is in the local cluster. 

Tests:

1. The fix is tested by running the webhook locally but pointing it to a downstream RKE2 cluster in Rancher. 
When I am trying to create a namespace and set PSA labels  as the project member, I see the following error as expected
```
admission webhook "rancher.cattle.io.namespaces" denied the request: Unauthorized
```
screencapture-137-184-181-209-dashboard-c-c-m-s59jql5k-explorer-namespace-create-2023-01-11-12_50_08

2. Also, @crobby  confirmed the fix working by building and using the container image with the same fix.

### DIFF
--- a/pkg/server/handlers.go
+++ b/pkg/server/handlers.go
@@ -27,11 +27,10 @@ func Validation(clients *clients.Clients) ([]admission.ValidatingAdmissionHandle
 		cluster.NewProvisioningClusterValidator(clients),
 		&machineconfig.Validator{},
 		nshandler.NewValidator(clients.K8s.AuthorizationV1().SubjectAccessReviews()),
-		podsecurityadmissionconfigurationtemplate.NewValidator(clients.Management.Cluster().Cache(),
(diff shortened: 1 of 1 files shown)


crobby (approved):
I've built and tested this fix.  lgtm

HarrisonWAffel (approved):

PR #178, opened 2023-01-12
Go dependencies bumps release/v0.3

macedogm:
Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>


KevinJoiner (approved):
Are the dependencies still considered indirect if we update them to a specific version?

macedogm:
Yes, they are still indirect.

PR #179, opened 2023-01-13
Fix handling of PSA label deletion

crobby:
Addresses issue found in testing of https://github.com/rancher/rancher/issues/38701

Updated handling of potential PSA-related label deletion.
Tweaked 1 test and added 2 new unit tests to cover the behaviors.

### DIFF
--- a/pkg/resources/validation/psa-validation.go
+++ b/pkg/resources/validation/psa-validation.go
@@ -24,11 +24,9 @@ var psaLabels = []string{
 // IsUpdatingPSAConfig will indicate whether or not the labels being passed in
 // are attempting to update PSA-related configuration.
 func IsUpdatingPSAConfig(old map[string]string, new map[string]string) bool {
-	for label := range new {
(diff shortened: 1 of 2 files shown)


pkg/resources/validation/psa-validation.go
```diff
@@ -31,6 +31,14 @@ func IsUpdatingPSAConfig(old map[string]string, new map[string]string) bool {
 			}
 		}
 	}
+	// Check to see if a psa-related label has been deleted in the update
```
KevinJoiner, line 34: You can simplify `IsUpdatingPSAConfig` to simply loop over `psaLabels` once
``` golang
func IsUpdatingPSAConfig(old map[string]string, new map[string]string) bool {
    for _, label := range psaLabels {
        if old[label] != new[label] {
	        return true
        }
    }
    return false
}
```

KevinJoiner (changes requested):

maxsokolovsky (approved):

PR #184, opened 2023-01-26
Add PSP references along with setting and logic to handle PSPs

crobby:
https://github.com/rancher/rancher/issues/40326

### DIFF
--- a/charts/rancher-webhook/templates/pre-delete-hook-psp.yaml
+++ b/charts/rancher-webhook/templates/pre-delete-hook-psp.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.preDelete.enabled (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
+{{- if .Values.global.cattle.psp.enabled  -}}
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
--- a/charts/rancher-webhook/templates/pre-delete-hook-psp.yaml
+++ b/charts/rancher-webhook/templates/pre-delete-hook-psp.yaml
@@ -1,0 +1,33 @@
+{{- if and .Values.preDelete.enabled (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: rancher-webhook-pre-delete
+  labels: {{ include "rancher-webhook.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
+spec:
+  privileged: false
+  hostNetwork: false
+  hostIPC: false
+  hostPID: false
+  runAsUser:
+    rule: 'RunAsAny'
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'MustRunAs'
+    ranges:
+      - min: 1
+        max: 65535
+  fsGroup:
+    rule: 'MustRunAs'
+    ranges:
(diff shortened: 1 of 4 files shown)


MbolotSuse (approved):
LGTM. Hold off on merging for now, there might need to be a slight change needed in validate-psp-install.yaml

maxsokolovsky (dismissed):

charts/rancher-webhook/templates/validate-psp-install.yaml
```diff
@@ -0,0 +1,5 @@
+#{{- if .Values.global.cattle.psp.enabled }}
+#{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
+#{{- required "The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding." "" -}}
```
MbolotSuse, line 3: Requirements for this changed to require `fail` instead of `required`. Would you mind changing this to be in line with our other charts?

MbolotSuse (changes requested):

charts/rancher-webhook/templates/validate-psp-install.yaml
```diff
@@ -0,0 +1,5 @@
+#{{- if .Values.global.cattle.psp.enabled }}
+#{{- if not (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
+#{{- fail "The target cluster does not have the PodSecurityPolicy API resource. Please disable PSPs in this chart before proceeding." "" -}}
```
MbolotSuse, line 3: Sorry, wasn't super clear here. `fail` will only take one arg, where as required took two. This is causing your CI to fail. I would recommend taking the content of this file directly from the issue.

MbolotSuse (changes requested):

MbolotSuse (approved):
LGTM

charts/rancher-webhook/values.yaml
```diff
@@ -6,6 +6,8 @@ image:
 global:
   cattle:
     systemDefaultRegistry: ""
+    psp:
+      enabled: true #PSPs enablement should default to true
```
maxsokolovsky, line 10: Do we need this comment?

charts/rancher-webhook/templates/pre-delete-hook-psp.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.preDelete.enabled (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}
+{{- if .Values.global.cattle.psp.enabled  -}}
```
maxsokolovsky, line 1: I am not a Helm expert - why is there a dash before the ending braces here but not on line 19 in the `pre-delete-cluster-role.yaml`?

charts/rancher-webhook/values.yaml
MbolotSuse: The comment was in the issue template and was adapted from the docs on this format. Given that it does not affect the functionality of this in anyway, I don't think it needs to change.

charts/rancher-webhook/templates/pre-delete-hook-psp.yaml
MbolotSuse: Helm uses the - to control the use of whitespace. In this case, AFAIK this tells helm to "eat" the whitespace. In this case, it does not appear to make a difference testing on my local with the trailing `-` and without the trailing `-` using `helm install --dry-run rancher-webhook -n cattle-system` did not produce a difference in whitespace.

Now this does not hold for the change in the `ClusterRole.yaml`. If you include the `-` on the if statement there you get a rendering error.

maxsokolovsky: Interesting, thanks for testing this out!

maxsokolovsky (approved):

PR #187, opened 2023-01-27
add the support for using PSACT on RKE2/K3S clusters

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/39994

This PR adds support for using PSACT on RKE2/K3S clusters.
Mainly it does the following two things when  creating/updating/deleting an RKE2/K3S cluster:
- create/update/delete a Secret which holds the generated admission configuration file from the PSACT for the target cluster 
- update/remove the following fields  from the cluster's Spec: 
	- a MachineSelectorFiles entry that mounts the admission configuration file to the control plane node 
	- the extra arg for kube-apisever to use the admission configuration file

Testing:
This PR is tested locally to confirm it is working as expected.


pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -2,44 +2,89 @@ package cluster
...
 	"github.com/rancher/webhook/pkg/admission"
+	"github.com/rancher/webhook/pkg/auth"
+	"github.com/rancher/webhook/pkg/clients"
+	v3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
 	objectsv1 "github.com/rancher/webhook/pkg/generated/objects/provisioning.cattle.io/v1"
-	"github.com/rancher/webhook/pkg/resources/mutation"
+	"github.com/rancher/webhook/pkg/patch"
+	psa "github.com/rancher/webhook/pkg/podsecurityadmission"
+	"github.com/rancher/wrangler/pkg/data/convert"
+	v1controller "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
```
jakefhyde, line 19:
```suggestion
	corecontrollers "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
```

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -2,44 +2,89 @@ package cluster
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/trace"
 )
 
+const (
+	KubeAPIAdmissionConfigOption = "admission-control-config-file"
+	ControlPlaneRoleLabel        = "rke.cattle.io/control-plane-role"
+	SecretAnnotation             = "rke.cattle.io/object-authorized-for-clusters"
+	SecretNamespace              = "fleet-default"
```
jakefhyde, line 35: I don't think the concept of a dedicated secret namespace is applicable here: The secret namespace should be equivalent to the namespace of the associated provisioning cluster. This is currently `"fleet-default"` in practice, but there is active work being done to change this.

jakefhyde (changes requested):
I'm still reviewing the overall logic, but I'll send my review in prematurely to give some time to address some of the comments.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +98,258 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	}
+	return map[string]string{}
+}
+
+func toMap(input interface{}) map[string]string {
+	args := map[string]string{}
+	parsed := convert.ToInterfaceSlice(input)
+	if parsed != nil {
+		for _, arg := range parsed {
+			pair := strings.Split(convert.ToString(arg), "=")
```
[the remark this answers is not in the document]
jiaqiluo: @jakefhyde  Can you explain why strings.Cut is a better choice here?  
One possible reason I can think of is for the case of "foo=bar=baz", strings.Cut makes sure we can get the "bar=baz" part. But I doubt if it is a valid example, as least I do not see any option that allows the value contains "=" for kube-api-server.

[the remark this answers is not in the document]
jakefhyde: `Strings.Cut` is a good choice when you are trying to get the left and right sides of a string delimited by the first instance of a token, especially when there may be other instances of the token on the right.

I think restricting this to only strings with a single `"="` is probably too optimistic about the arguments we have to support, as even if we don't support any flags like that, I don't think we can guarantee that a `"="` anywhere in the flag is always invalid.

These are the possible flags I have identified as being potentially broken with the existing logic (from https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/, I'm not sure if these are all applicable).

```
--allow-metric-labels (Example: metric1,label1='v1,v2,v3', metric1,label2='v1,v2,v3' metric2,label1='v1,v2,v3'
--feature-gates <comma-separated 'key=True|False' pairs>
--oidc-required-claim <comma-separated 'key=value' pairs>
--runtime-config <comma-separated 'key=value' pairs>
--strict-transport-security-directives strings (Example: 'max-age=31536000,includeSubDomains,preload')
--vmodule pattern=N,... (comma-separated list of pattern=N settings for file-filtered logging (only works for text log format)
```

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +98,258 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+		Type:       corev1.SecretTypeOpaque,
+	}
+
+	if annotations != nil {
+		secret.Annotations = annotations
+	}
+	if existing == nil {
+		return m.secret.Create(secret)
+	}
+	if !reflect.DeepEqual(existing.StringData, secret.StringData) {
```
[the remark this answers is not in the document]
jiaqiluo: > StringData is a write-only field, so this comparison needs to be updated, as it will always be false if data != nil or {}. 

Based on my test, this is not true. Because we set the  `StringData` field when creating the Secret, it is accessible in the returned Secret object and can be used to compare. For testing purposes, I add a printing in the if clause and only see the message in the log when the StringData is changed (by changing the defaultPSACT assigned to the cluster).

pkg/resources/mutation/cluster/provisioningcluster.go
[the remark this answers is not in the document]
jiaqiluo: This is a good point!
I have updated the code.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: Good point! 
I updated the code to get the namespace from the cluster object to create the secret in the same namespace as the cluster. Currently, its value is fleet-default.

pkg/resources/mutation/cluster/provisioningcluster.go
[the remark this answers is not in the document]
jakefhyde: I don't think that logic holds true if the webhook restarts. Additionally, StringData is probably only populated for the secret until it is updated somewhere else. Can you verify this is working as intended after a restart?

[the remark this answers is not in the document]
jiaqiluo: Here is what I found:  
Suppose others update the Secret outside of rancher-webhook, for example, by adding an annotation. In that case, the field `StringData` will be reset to its default empty value, later on, when rancher-webhook compares the existing one with the new one, it will detect the change and update the Secret even though the value is unchanged. 

Based on the above finding and your advice, I changed to logic to update and compare the `Data` field in the Secret.

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -164,6 +178,93 @@ func (p *ProvisioningClusterValidator) validateClusterName(request *admission.Re
...
+
+	name := fmt.Sprintf(mutator.SecretName, cluster.Name)
+	mountPath := fmt.Sprintf(mutator.MountPath, mutator.GetRuntime(cluster.Spec.KubernetesVersion))
+	templateName := cluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+
+	switch request.Operation {
+	case admissionv1.Delete:
+		_, err := p.secretCache.Get(cluster.Namespace, name)
+		if err == nil {
+			return fmt.Errorf("[provsioning cluster validator] the secret %s still exisit in the cluster", name)
```
jakefhyde, line 195:
```suggestion
			return fmt.Errorf("[provsioning cluster validator] the secret %s still exists in the cluster", name)
```

jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -2,44 +2,88 @@ package cluster
...
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/trace"
 )
 
+const (
+	KubeAPIAdmissionConfigOption = "admission-control-config-file"
+	ControlPlaneRoleLabel        = "rke.cattle.io/control-plane-role"
+	SecretAnnotation             = "rke.cattle.io/object-authorized-for-clusters"
+	SecretName                   = "%s-admission-configuration-psact"
+	Secretkey                    = "admission-config-psact"
```
jakefhyde, line 36:
```suggestion
	SecretKey                    = "admission-config-psact"
```

jakefhyde (changes requested):
Just a few nits, rest of the logic looks sound.

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -164,6 +178,93 @@ func (p *ProvisioningClusterValidator) validateClusterName(request *admission.Re
...
+		}
+	case admissionv1.Create, admissionv1.Update:
+		if cluster.DeletionTimestamp != nil {
+			return nil
+		}
+		if templateName == "" {
+			// validate that the secret does not exist
+			_, err := p.secretCache.Get(cluster.Namespace, name)
+			if err == nil {
+				return fmt.Errorf("[provsioning cluster validator] the secret %s still exisit in the cluster", name)
```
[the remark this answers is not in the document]
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -164,6 +178,93 @@ func (p *ProvisioningClusterValidator) validateClusterName(request *admission.Re
...
+
+	name := fmt.Sprintf(mutator.SecretName, cluster.Name)
+	mountPath := fmt.Sprintf(mutator.MountPath, mutator.GetRuntime(cluster.Spec.KubernetesVersion))
+	templateName := cluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+
+	switch request.Operation {
+	case admissionv1.Delete:
+		_, err := p.secretCache.Get(cluster.Namespace, name)
+		if err == nil {
+			return fmt.Errorf("[provsioning cluster validator] the secret %s still exisits in the cluster", name)
```
jakefhyde, line 195:
```suggestion
			return fmt.Errorf("[provsioning cluster validator] the secret %s still exists in the cluster", name)
```

jakefhyde (approved):
1 Nit, I'll approve though so if nothing else changes just assume you have my approval

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,258 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	}
+
+	secretName := fmt.Sprintf(SecretName, cluster.Name)
+	mountPath := fmt.Sprintf(MountPath, GetRuntime(cluster.Spec.KubernetesVersion))
+	templateName := cluster.Spec.DefaultPodSecurityAdmissionConfigurationTemplateName
+
+	switch request.Operation {
+	case admissionv1.Delete:
+		err := m.secret.Delete(cluster.Namespace, secretName, &metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
```
annablender, line 149: What if the error is not nil but it's also not an `IsNotFound` API error? Couldn't the mutator fail to delete an existing secret? If so, I'd suggest updating this to the error handling pattern on line 171

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,258 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	case admissionv1.Create, admissionv1.Update:
+		if cluster.DeletionTimestamp != nil {
+			return admission.ResponseAllowed(), nil
+		}
+		if templateName == "" {
+			err := m.secret.Delete(cluster.Namespace, secretName, &metav1.DeleteOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				return nil, fmt.Errorf("[provsioning cluster mutator] failed to delete the secret: %w", err)
+			}
+			// drop relevant fields if they exist in the cluster
```
annablender, line 161: Shouldn't the logic for dropping relevant fields from the cluster management object also be under the admission.v1 Delete case? If those fields still exist, and you're just deleting the secret they would be incorrectly left behind on the object.. but lmk if I just missed something because I haven't had enough Monday coffee yet

In general, I think a few more `Now, handle the case of ` comments in this section like the one on line 169 would be helpful

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,258 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	}
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Data: data,
+		Type: corev1.SecretTypeOpaque,
+	}
+
```
annablender, line 220: Nit: remove extra line

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,258 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// AddMachineSelectorFile adds the provided RKESystemFiles to the cluster's MachineSelectorFiles list.
+// It is a no-op if the provided RKESystemFiles already exists in the cluster.
```
annablender, line 303: Nit: `already exist in the cluster` - nicely worded

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -164,6 +178,93 @@ func (p *ProvisioningClusterValidator) validateClusterName(request *admission.Re
 	return nil
 }
 
+// validatePSACT validate if the cluster and underlying secret are configured properly when PSACT is used in or dropped from the cluster
```
annablender, line 181: Suggestion: `validates if the cluster and underlying secret are configured properly when PSACT is enabled or disabled`

annablender (changes requested):

pkg/podsecurityadmission/podsecurityadmission.go
```diff
@@ -123,3 +125,38 @@ func GetClusterVersion(version string) (semver.Version, error) {
 	}
 	return parsedVersion, nil
 }
+
+// GenerateAdmissionConfigFile generate the admission configuration file for PodSecurity based on the provided PodSecurityAdmissionConfigurationTemplate.
```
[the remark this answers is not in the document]
jiaqiluo: fixed.

pkg/podsecurityadmission/podsecurityadmission.go
```diff
@@ -123,3 +125,38 @@ func GetClusterVersion(version string) (semver.Version, error) {
...
+	if err != nil {
+		return []byte(""), fmt.Errorf("failed to parse cluster version: %w", err)
+	}
+	switch {
+	case parsedRangeAtLeast125(parsedVersion):
+		input["Version"] = "v1"
+	case parsedRange123to124(parsedVersion):
+		input["Version"] = "v1beta1"
+	default:
+		return []byte(""), fmt.Errorf("k8s version need to be at least 1.23")
```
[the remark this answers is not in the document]
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,258 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+		return nil, fmt.Errorf("failed to create patch: %w", err)
+	}
 	return response, nil
 }
+
+// handlePSACT updates the cluster and an underlying secret to support PSACT.
+// If a PSACT is set in the cluster, handlePSACT generates an admission configuration file, put it into a secret,
+// updates the cluster's spec to mount the secret to the control plane nodes, and configures kube-apisever to use the admission configuration file;
+// If a PSACT is unset in the cluster, handlePSACT does the cleanup for both the secret and cluster's spec.
+func (m *ProvisioningClusterMutator) handlePSACT(request *admission.Request, cluster *v1.Cluster) (*admissionv1.AdmissionResponse, error) {
```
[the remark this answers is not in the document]
jiaqiluo: Hmm, I have a hard time thinking of a better name that can reflect what the function does on the cluster object. Any suggestion?  😄

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: In this case, the user will see his operation is denied in the Rancher UI(the HTTP request fails), and he needs to check the rancher-webhook's logs if he wants to see the error message. 
I do not return the error to the user, like on line 171, because it is an internal error that should not be exposed to the user directly. As a compassion, on line 171, the error happens because the user sets a non-exist PSACT, in such case, we should return the error to the user to let him know. 
I am open to more discussion on this.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed

pkg/resources/validation/cluster/provisioningcluster.go
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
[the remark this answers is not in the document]
annablender: I just meant capitalize for cohesion

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+			if err != nil {
+				return nil, fmt.Errorf("[provsioning cluster mutator] failed to generate admission configuration from PSACT [%s]: %w", templateName, err)
+			}
+			anno := map[string]string{
+				SecretAnnotation: cluster.Name,
+			}
+			data := map[string][]byte{
+				SecretKey: fileContent,
+			}
+			_, err = m.createOrUpdateSecret(cluster.Namespace, secretName, data, anno)
```
KevinJoiner, line 187: nit: If we do not need the variable we should not return it since this is the only call to our private function.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+					},
+				},
+			},
+		},
+	}
+}
+
+// AddMachineSelectorFile adds the provided RKESystemFiles to the cluster's MachineSelectorFiles list.
+// It is a no-op if the provided RKESystemFiles already exist in the cluster.
+func AddMachineSelectorFile(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster) *v1.Cluster {
```
KevinJoiner, line 303: Same as the DropMachineSelectorFile. 
```suggestion
func addMachineSelectorFile(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster) {
```

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	for _, item := range cluster.Spec.RKEConfig.MachineSelectorFiles {
+		if equality.Semantic.DeepEqual(file, item) {
+			return cluster
+		}
+	}
+	// adding the new one to the beginning of the list will theoretically make the lookup quicker by reducing iteration
+	cluster.Spec.RKEConfig.MachineSelectorFiles = append([]rkev1.RKEProvisioningFiles{file}, cluster.Spec.RKEConfig.MachineSelectorFiles...)
```
KevinJoiner, line 310: Note: if we really care about speed, it would be faster to append to the end and just change our lookup to start searching back to front since this requires `append` to make a new slice and copy the entries.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+			args[key] = val
+		}
+	}
+	return args
+}
+
+// SetKubeAPIServerArg uses the provided arg to overwrite the value of kube-apiserver-arg under the cluster's MachineGlobalConfig.
+// If the provided arg is an empty map, SetKubeAPIServerArg removes the existing kube-apiserver-arg from the cluster's MachineGlobalConfig.
+func SetKubeAPIServerArg(arg map[string]string, cluster *v1.Cluster) *v1.Cluster {
+	var parsed []interface{}
```
KevinJoiner, line 261: We can preallocate our slice since we know what size it will be.
nit: I think any is easier to read and is an alias for `interface{}`
```suggestion
	parsed := make([]any, 0, len(arg))
```

pkg/resources/mutation/cluster/provisioningcluster.go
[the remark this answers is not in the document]
jiaqiluo: Oh, I see what you mean now. 
I made it private because I thought this function would be used only internally in the struct.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+			}
+			args[key] = val
+		}
+	}
+	return args
+}
+
+// SetKubeAPIServerArg uses the provided arg to overwrite the value of kube-apiserver-arg under the cluster's MachineGlobalConfig.
+// If the provided arg is an empty map, SetKubeAPIServerArg removes the existing kube-apiserver-arg from the cluster's MachineGlobalConfig.
+func SetKubeAPIServerArg(arg map[string]string, cluster *v1.Cluster) *v1.Cluster {
```
KevinJoiner, line 260: Same issue as DropMachineSelectorFile(...)
```suggestion
func setKubeAPIServerArg(arg map[string]string, cluster *v1.Cluster) {
```

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	var parsed []interface{}
+	for key, val := range arg {
+		parsed = append(parsed, interface{}(fmt.Sprintf("%s=%s", key, val)))
+	}
+	if len(parsed) == 0 {
+		delete(cluster.Spec.RKEConfig.MachineGlobalConfig.Data, "kube-apiserver-arg")
+	} else {
```
KevinJoiner, line 267: Can this be moved to the top of the function?
```suggestion
	if len(arg) == 0 {
		delete(cluster.Spec.RKEConfig.MachineGlobalConfig.Data, "kube-apiserver-arg")
		return
	} 
```

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	if cluster.Spec.RKEConfig.MachineGlobalConfig.Data != nil {
+		return toMap(cluster.Spec.RKEConfig.MachineGlobalConfig.Data["kube-apiserver-arg"])
+	}
+	return map[string]string{}
+}
+
+func toMap(input interface{}) map[string]string {
+	args := map[string]string{}
+	parsed := convert.ToInterfaceSlice(input)
+	if parsed != nil {
```
KevinJoiner, line 245: We do not need a nil check since range can handle a nil slice.
```suggestion
```

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -164,6 +178,93 @@ func (p *ProvisioningClusterValidator) validateClusterName(request *admission.Re
...
+	switch request.Operation {
+	case admissionv1.Delete:
+		_, err := p.secretCache.Get(cluster.Namespace, name)
+		if err == nil {
+			return fmt.Errorf("[provsioning cluster validator] the secret %s still exists in the cluster", name)
+		} else if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("[provsioning cluster validator] failed to validate if the secret exists: %w", err)
+		} else {
+			return nil
+		}
```
KevinJoiner, line 200: We don't need else with early returns.
```suggestion
		if err == nil {
			return fmt.Errorf("[provsioning cluster validator] the secret %s still exists in the cluster", name)
		} 
		if !apierrors.IsNotFound(err) {
			return fmt.Errorf("[provsioning cluster validator] failed to validate if the secret exists: %w", err)
		} 
		return nil
		
```

pkg/resources/validation/cluster/provisioningcluster.go
```diff
@@ -164,6 +178,93 @@ func (p *ProvisioningClusterValidator) validateClusterName(request *admission.Re
...
+	case admissionv1.Create, admissionv1.Update:
+		if cluster.DeletionTimestamp != nil {
+			return nil
+		}
+		if templateName == "" {
+			// validate that the secret does not exist
+			_, err := p.secretCache.Get(cluster.Namespace, name)
+			if err == nil {
+				return fmt.Errorf("[provsioning cluster validator] the secret %s still exists in the cluster", name)
+			} else if !apierrors.IsNotFound(err) {
```
KevinJoiner, line 210:
```suggestion
			}
			if !apierrors.IsNotFound(err) {
```

KevinJoiner (changes requested):
I don't know what the time frame is on this, but if not in this PR we will need to come back and add unit testa for provisioningcluster validation.

pkg/server/handlers.go
```diff
@@ -47,7 +47,7 @@ func Validation(clients *clients.Clients) ([]admission.ValidatingAdmissionHandle
 // Mutation returns a list of all MutatingAdmissionHandlers used by the webhook.
 func Mutation(clients *clients.Clients) ([]admission.MutatingAdmissionHandler, error) {
 	return []admission.MutatingAdmissionHandler{
-		&mutationCluster.Mutator{},
+		mutationCluster.NewProvisioningClusterMutator(clients),
```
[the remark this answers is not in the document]
jiaqiluo: How can I update my PR to the pattern you described here? Can you point me to an example in the code base? 
Or do you mean that is something we will do?

pkg/podsecurityadmission/podsecurityadmission.go
```diff
@@ -123,3 +125,38 @@ func GetClusterVersion(version string) (semver.Version, error) {
...
+func GenerateAdmissionConfigFile(configurationTemplate *apisv3.PodSecurityAdmissionConfigurationTemplate, k8sVersion string) ([]byte, error) {
+	input := map[string]interface{}{
+		"Version":    "",
+		"Defaults":   configurationTemplate.Configuration.Defaults,
+		"Exemptions": configurationTemplate.Configuration.Exemptions,
+	}
+
+	parsedVersion, err := GetClusterVersion(k8sVersion)
+	if err != nil {
+		return []byte(""), fmt.Errorf("failed to parse cluster version: %w", err)
```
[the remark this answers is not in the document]
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -2,44 +2,88 @@ package cluster
...
+const (
+	KubeAPIAdmissionConfigOption = "admission-control-config-file"
+	ControlPlaneRoleLabel        = "rke.cattle.io/control-plane-role"
+	SecretAnnotation             = "rke.cattle.io/object-authorized-for-clusters"
+	SecretName                   = "%s-admission-configuration-psact"
+	SecretKey                    = "admission-config-psact"
+	MountPath                    = "/etc/rancher/%s/config/rancher-psact.yaml"
+	RuntimeK3S                   = "k3s"
+	RuntimeRKE2                  = "rke2"
+	RuntimeRKE                   = "rke"
```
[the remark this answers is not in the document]
jiaqiluo: fixed, now only the necessary ones are exported.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+// If a PSACT is set in the cluster, handlePSACT generates an admission configuration file, mounts the file into a secret,
+// updates the cluster's spec to mount the secret to the control plane nodes, and configures kube-apisever to use the admission configuration file;
+// If a PSACT is unset in the cluster, handlePSACT does the cleanup for both the secret and cluster's spec.
+func (m *ProvisioningClusterMutator) handlePSACT(request *admission.Request, cluster *v1.Cluster) (*admissionv1.AdmissionResponse, error) {
+	if cluster.Name == "local" || cluster.Spec.RKEConfig == nil {
+		return admission.ResponseAllowed(), nil
+	}
+	parsedVersion, err := psa.GetClusterVersion(cluster.Spec.KubernetesVersion)
+	if err != nil {
+		return nil, fmt.Errorf("[provsioning cluster mutator] failed to parse cluster version: %w", err)
```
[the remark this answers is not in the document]
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/validation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/validation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: good point! 
fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+}
+
+// createOrUpdateSecret creates or updates a secret based on the provided information.
+func (m *ProvisioningClusterMutator) createOrUpdateSecret(namespace, name string, data map[string][]byte, annotations map[string]string) (*corev1.Secret, error) {
+	var existing *corev1.Secret
+	var err error
+	if namespace == "" || name == "" {
+		return existing, fmt.Errorf("namespace and name are required")
+	}
+	existing, err = m.secret.Cache().Get(namespace, name)
```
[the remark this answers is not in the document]
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+			args := GetKubeAPIServerArg(cluster)
+			args[KubeAPIAdmissionConfigOption] = mountPath
+			cluster = SetKubeAPIServerArg(args, cluster)
+		}
+	}
+	return admission.ResponseAllowed(), nil
+}
+
+// createOrUpdateSecret creates or updates a secret based on the provided information.
+func (m *ProvisioningClusterMutator) createOrUpdateSecret(namespace, name string, data map[string][]byte, annotations map[string]string) (*corev1.Secret, error) {
```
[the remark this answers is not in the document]
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+			return cluster
+		}
+	}
+	// adding the new one to the beginning of the list will theoretically make the lookup quicker by reducing iteration
+	cluster.Spec.RKEConfig.MachineSelectorFiles = append([]rkev1.RKEProvisioningFiles{file}, cluster.Spec.RKEConfig.MachineSelectorFiles...)
+	return cluster
+}
+
+// DropMachineSelectorFile removes the provided RKESystemFile from the cluster if it is found in the cluster.
+func DropMachineSelectorFile(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster) *v1.Cluster {
```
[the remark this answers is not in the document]
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +97,257 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+	source := cluster.Spec.RKEConfig.MachineSelectorFiles
+	for i, item := range source {
+		if equality.Semantic.DeepEqual(file, item) {
+			if len(source) == 1 {
+				cluster.Spec.RKEConfig.MachineSelectorFiles = nil
+				break
+			}
+			copy(source[i:], source[i+1:])
+			cluster.Spec.RKEConfig.MachineSelectorFiles = source[:len(source)-1]
```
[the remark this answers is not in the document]
jiaqiluo: fixed.  BTW, since `source` here is a slice of the object of type rkev1.RKEProvisioningFiles,  not a point, skip `source[i] = nil` step.

pkg/podsecurityadmission/podsecurityadmission.go
```diff
@@ -123,3 +125,38 @@ func GetClusterVersion(version string) (semver.Version, error) {
 	}
 	return parsedVersion, nil
 }
+
+// GenerateAdmissionConfigFile generates the admission configuration file for PodSecurity based on the provided PodSecurityAdmissionConfigurationTemplate.
+// The k8sVersion is required for determining the API version.
+func GenerateAdmissionConfigFile(configurationTemplate *apisv3.PodSecurityAdmissionConfigurationTemplate, k8sVersion string) ([]byte, error) {
```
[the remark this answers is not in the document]
jiaqiluo: The implementation is updated!

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -2,44 +2,91 @@ package cluster
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/trace"
 )
 
+const (
+	KubeAPIAdmissionConfigOption = "admission-control-config-file"
```
crobby, line 34: Exported constants should have a comment

pkg/podsecurityadmission/podsecurityadmission.go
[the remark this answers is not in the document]
KevinJoiner: Nice job! Reusing `GetPluginConfigFromTemplate` makes this much cleaner.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -53,10 +100,250 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+			return
+		}
+	}
+	cluster.Spec.RKEConfig.MachineSelectorFiles = append(cluster.Spec.RKEConfig.MachineSelectorFiles, file)
+}
+
+// dropMachineSelectorFile removes the provided RKEProvisioningFiles from the cluster if it is found in the cluster.
+func dropMachineSelectorFile(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster) {
+	source := cluster.Spec.RKEConfig.MachineSelectorFiles
+	for i := len(source) - 1; i >= 0; i-- {
```
KevinJoiner, line 315: Leave a comment like before that we are doing this for faster lookups.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -2,44 +2,91 @@ package cluster
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/trace"
 )
 
+const (
+	KubeAPIAdmissionConfigOption = "admission-control-config-file"
+	SecretName                   = "%s-admission-configuration-psact"
+	MountPath                    = "/etc/rancher/%s/config/rancher-psact.yaml"
```
KevinJoiner, line 36: these exported variables need comments too.

KevinJoiner (changes requested):
Small things, but overall looks good! Before you merge, we can leave this into 2 commits, one for PSCAT support and one for the drone fixes.

pkg/server/handlers.go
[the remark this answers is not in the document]
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: There will not be a conflict even If the secret is left in the cluster because Rancher will either update the existing secret next time user updates the cluster or tries to delete the secret again next time user deletes the cluster.

KevinJoiner (approved):
LGTM

pkg/resources/mutation/cluster/provisioningcluster.go
annablender: Ahh ok, thank you for clarifying. I think that is fine then

annablender (approved):

PR #189, opened 2023-02-08
Removing PSP uses/references and adding MutatingWebhook to the Chart

MbolotSuse:
Related to rancher/rancher#40326.

## Overview
This PR removes the psps from the rancher-webhook chart. These were only used previously to allow the pre-delete job to run, which removed the MutatingValidationWebhook that the webhook created as a part of this code.

In order to not orphan these resources and re-introduce an earlier issue, the MutatingWebhook is now initially declared in the chart. This will cause helm to remove the resource when a user uninstalls the webhook through the use of the chart. 

## Changes
- Moves the MutatingWebhookConfiguration into the chart
- Removes the pre- delete job and the associated resources, which was only used to remove the MutatingValidationWebhook
- Removes the pre-delete configuration from values.yaml and the psp configuration variables from values.yaml

## Notes
This will require a change in rancher to be properly used. In upgrade scenarios, there will already be a MutatingWebhookConfiguration which is not managed by helm but has the same name as the one declared by the chart. To fix this, we will either need to force helm to "adopt" this resource (taking the existing resource and adding labels/annotations to it), adopt the resource into helm manually, or remove the resource and let the chart/webhook re-create it. In any case, these changes will need to be made in a separate PR to rancher (which should also remove the logic to pass down PSP settings to the webhook).


jiaqiluo:
@MbolotSuse  The changes to the chart LGTM.  But it is better to have Rancher ready first before merging this PR.   
Regarding the rancher side, it sounds rational to delete this MutatingValidationWebhook object right before updating the webhook.

MbolotSuse:
@jiaqiluo Thanks for the input. I've created the rancher side pr rancher/rancher#40480. I decided to use the force-adopt approach instead - which basically registers existing resources (including the existing mutating webhook) as part of this chart.

charts/rancher-webhook/templates/webhook.yaml
```diff
@@ -17,3 +17,23 @@ webhooks:
...
+    service:
+      name: rancher-webhook
+      namespace: cattle-system
+      path: /v1/webhook/validation
+      port: 443
+  failurePolicy: Ignore
+  matchPolicy: Equivalent
+  name: rancher.cattle.io
+  sideEffects: None
+  timeoutSeconds: 10
```
KevinJoiner, line 39: The webhooks list is optional.
``` suggestion
```

KevinJoiner (changes requested):

charts/rancher-webhook/templates/webhook.yaml
MbolotSuse: I'm trying to match what the chart declares for the "ValidatingAdmissionWebhook", which this does.

So my options are:
- Both have this irrelevant value
- Only one has the irrelevant value (don't like this)
- Both don't have the irrelevant value

What is your preference?

crobby: I'd prefer that the solutions match-up.  I wonder if there was a reason for including the irrelevant value in validating?

KevinJoiner: We should remove the values from both if we want them to be the same. Especially since it can be misleading if someone doesn't know they get overwritten on startup.

crobby: Yeah, that would probably be best (remove from both)

MbolotSuse: I've removed the `webhooks:` portion from both.

crobby (approved):
lgtm

KevinJoiner (approved):

MbolotSuse:
Rancher-side change has merged, so I'm going to merge this one now.

PR #196, opened 2023-02-27
Adds group checking to ruleResolvers.

KevinJoiner:
Forward Port: https://github.com/rancher/webhook/pull/195
This commit updates the CRTB and PRTB ruleResolvers to use CRTBs and PRTBs with matching groups as well as matching usernames. This commit also updates the caches used by the resolvers to have a new indexer that index on a composite key of the binding subject and the clusterName or ProjectName of the binding.

### DIFF
--- a/pkg/resolvers/aggregateResolver_test.go
+++ b/pkg/resolvers/aggregateResolver_test.go
@@ -1,4 +1,4 @@
-package resolvers_test
+package resolvers
 
 import (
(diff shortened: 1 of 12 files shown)


KevinJoiner:
https://github.com/rancherlabs/rancher-security/issues/1130

MbolotSuse (approved):

PR #197, opened 2023-03-06
Marshal before patching

jakefhyde:
https://github.com/rancher/rancher/issues/40826

Every time we update the provisioning cluster struct definition we are going to have to update the webhook's pkg/apis module dependency (like we did for https://github.com/rancher/webhook/pull/193) including for testing locally, which is probably going to result in a lot of churn for team 2 as we develop, and team 1 for merging webhook PRs and cutting releases. This changes the current behavior to only create a patch based on the marshalled json to prevent dropping fields unintentionally. I tested this locally, and it's now working as expected (I can add fields to the provisioning cluster without having to do a release of the webhook).

### DIFF
--- a/pkg/resources/mutation/cluster/provisioningcluster.go
+++ b/pkg/resources/mutation/cluster/provisioningcluster.go
@@ -3,6 +3,7 @@ package cluster
 import (
 	"crypto/sha256"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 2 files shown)


jakefhyde:
Holding off on merging until Q2 (but feel free to review in the meantime)

HarrisonWAffel (dismissed):

KevinJoiner (commented):
Can you add some unit test for this PR

KevinJoiner (approved):
Thanks for adding tests.

HarrisonWAffel (approved):

PR #198, opened 2023-03-07
Update the mutator and validator to support the content checking on RKEProvisioningFiles objects

jiaqiluo:
## Notes to Reviews

~This PR is a draft because we need the [rancher/rancher PR](https://github.com/rancher/rancher/pull/40789) to be merged first, then update the Go mod file.~
~Except for the Go mod file, the rest of the PR is ready for review.~

This PR is ready for review. 

## Issue: 

 
https://github.com/rancher/rancher/issues/40755

## Problem

 
When PSACT is used in the rke2 cluster and we upgrade the cluster from k8s 1.24 to 1.25, the Secret holding the generated admission config file is updated by rancher-webhook because the API version inside the config file is changed from v1beta1 to v1; meanwhile, the planner in Rancher keeps monitoring the RKEControlPlane and updating the Plan.   But there is a time lag between Rancher getting the updated Secret and the Plan. As a result, the planner sends the old Plan and the content from the new Secret to the control plane nodes and triggers the upgrade. Therefore the cluster fails to upgrade and logs the following errors in kube-apiserver:

`failed to initialize admission: couldn't init admission plugin \"PodSecurity\": no kind \"PodSecurityConfiguration\" is registered for version \"pod-security.admission.config.k8s.io/v1\" in scheme \"k8s.io/pod-security-admission/admission/api/scheme/scheme.go:30\"`

 
## Solution

 
To make sure the planner always uses the correct version of Secret and Plan to trigger the upgrade on the RKE2 cluster, we introduce a new field named `hash` to the underlying data struct, and its value is the base64 encoded value of the sha256 sum of the admission config file.  The planner compares the `hash` value with the one calculated based on the current Secret, and only processes to upgrade the cluster if those two values match.   

This PR is to update rancher-webhooks to set the `hash` value when it mutations the request for creating/updating the cluster.  

## Testing

This PR and the changes in rancher/rancher are tested locally.  Following the same steps in the original issue,  the REK2 cluster is successfully upgraded from 1.24 to 1.25.   Also, editing the cluster to change PSACT works as expected.


HarrisonWAffel (dismissed):

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -187,14 +190,16 @@ func (m *ProvisioningClusterMutator) handlePSACT(request *admission.Request, clu
...
 			}
 			err = m.ensureSecret(cluster.Namespace, secretName, data, anno)
 			if err != nil {
 				return nil, fmt.Errorf("[provisioning cluster mutator] failed to create or update the secret for the admission configuration file: %w", err)
 			}
-			// set relevant fields in the cluster
-			addMachineSelectorFile(MachineSelectorFileForPSA(secretName, mountPath), cluster)
+			// drop then set relevant fields if they exist in the cluster
+			dropMachineSelectorFile(MachineSelectorFileForPSA(secretName, mountPath, ""), cluster, true)
+			hash := sha256.Sum256(fileContent)
```
KevinJoiner, line 201: Nit: sha256 is fine here but for future uses, if you are only doing a checksum and nothing security related md5 is a faster hashing algorithm.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -328,9 +339,14 @@ func dropMachineSelectorFile(file rkev1.RKEProvisioningFiles, cluster *v1.Cluste
...
 
 // MachineSelectorFileExists returns a boolean to indicate if the provided RKEProvisioningFiles exist in the provided cluster.
-func MachineSelectorFileExists(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster) bool {
+func MachineSelectorFileExists(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster, ignoreValueCheck bool) bool {
 	for _, item := range cluster.Spec.RKEConfig.MachineSelectorFiles {
-		if equality.Semantic.DeepEqual(file, item) {
+		fromCluster := item.DeepCopy()
+		if ignoreValueCheck {
+			cleanupHash(fromCluster)
+			cleanupHash(&file)
```
KevinJoiner, line 347: `cleanupHash(&file)` Only needs to be called once so this could be moved out of the loop.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -328,9 +339,14 @@ func dropMachineSelectorFile(file rkev1.RKEProvisioningFiles, cluster *v1.Cluste
 }
 
 // MachineSelectorFileExists returns a boolean to indicate if the provided RKEProvisioningFiles exist in the provided cluster.
-func MachineSelectorFileExists(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster) bool {
+func MachineSelectorFileExists(file rkev1.RKEProvisioningFiles, cluster *v1.Cluster, ignoreValueCheck bool) bool {
 	for _, item := range cluster.Spec.RKEConfig.MachineSelectorFiles {
-		if equality.Semantic.DeepEqual(file, item) {
+		fromCluster := item.DeepCopy()
+		if ignoreValueCheck {
+			cleanupHash(fromCluster)
```
KevinJoiner, line 346: `DeepCopy()` is expensive so let's not call it unless we need to modify the object.
```suggestion
		fromCluster := item
		if ignoreValueCheck {
		    fromCluster = item.DeepCopy()
			cleanupHash(fromCluster)
```

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -187,14 +190,16 @@ func (m *ProvisioningClusterMutator) handlePSACT(request *admission.Request, clu
...
 			err = m.ensureSecret(cluster.Namespace, secretName, data, anno)
 			if err != nil {
 				return nil, fmt.Errorf("[provisioning cluster mutator] failed to create or update the secret for the admission configuration file: %w", err)
 			}
-			// set relevant fields in the cluster
-			addMachineSelectorFile(MachineSelectorFileForPSA(secretName, mountPath), cluster)
+			// drop then set relevant fields if they exist in the cluster
+			dropMachineSelectorFile(MachineSelectorFileForPSA(secretName, mountPath, ""), cluster, true)
+			hash := sha256.Sum256(fileContent)
+			addMachineSelectorFile(MachineSelectorFileForPSA(secretName, mountPath, base64.StdEncoding.EncodeToString(hash[:])), cluster)
```
KevinJoiner, line 202: `hash[:]` this is pretty cool, I never thought about doing this to convert an array to a slice.

pkg/resources/mutation/cluster/provisioningcluster.go
```diff
@@ -277,7 +282,7 @@ func setKubeAPIServerArg(arg map[string]string, cluster *v1.Cluster) {
 
 // MachineSelectorFileForPSA generates an RKEProvisioningFiles that mounts the secret which contains
 // the generated admission configuration file to the control plane node
-func MachineSelectorFileForPSA(secretName, mountPath string) rkev1.RKEProvisioningFiles {
+func MachineSelectorFileForPSA(secretName, mountPath, hash string) rkev1.RKEProvisioningFiles {
```
KevinJoiner, line 285: nit: You don't need to change anything, but could we have used pointers for all our calls using `rkev1.RKEProvisioningFiles` purely asking out of curiosity definitely not a big deal since the struct only has two pointers in it.

KevinJoiner (changes requested):

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: Good to know that 😄  
SHA256 is used here because it is commonly used in rancher. 
I will keep it because changes need to be done both here and in rancher/rancher.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed.

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: fixed

pkg/resources/mutation/cluster/provisioningcluster.go
jiaqiluo: The only reason for returning an instance is because the Cluster contains a list of `rkev1.RKEProvisioningFiles`, not pointer to `rkev1.RKEProvisioningFiles`, so it makes some operations easy. 

However, as the implementation iterates, I feel it makes more sense to return a pointer.  

The code is updated now.

pkg/resources/mutation/cluster/provisioningcluster.go
pjbgf: On the hash algorithmn selection, we just need to be mindful of the implications on users that use Rancher in regulated environments. Ideally sticking to NIST approved options (e.g. `sha256`, `sha512`).

But looking on the bright side, `SHA-NI` [improvements](https://go-review.googlesource.com/c/go/+/408795/1) are coming for Go's `sha256` implementation, which may even mean it beat `md5`'s performance. :smile:

KevinJoiner (approved):
LGTM! 👌🏿  don't forget to squash when merging.

HarrisonWAffel (approved):

PR #206, opened 2023-03-16
Allows namespace Webhook to provide multiple configurations.

KevinJoiner:
Issue: https://github.com/rancher/rancher/issues/40843

This commit allows attempts to create the kube-system namespace while the webhook is down to pass through rancher-webhook.


pkg/resources/validation/namespace/namespace.go
```diff
@@ -38,17 +40,63 @@ func (v *Validator) GVR() schema.GroupVersionResource {
...
+	for i, rule := range standardWebhook.Rules {
+		idx := slices.Index(rule.Operations, admv1.Create)
+		if idx != -1 {
+			standardWebhook.Rules[i].Operations = slices.Delete(rule.Operations, idx, idx+1)
+		}
+	}
+
+	// default configuration for all create operations except those belonging to the kube-system namespace
+	createWebhook := admission.NewDefaultValidatingWebhook(v, clientConfig, admv1.ClusterScope)
+	createWebhook.Name = "rancher.cattle.io.namespace.create.non.kubesystem"
```
MbolotSuse, line 67: This value seems to me like something that should only partially be defined in this package. The `rancher.cattle.io` part of it isn't determined by a specific validator/mutator wide. I think it might be better to have a function (something like `admission.SetWebhookName(resourceType string, suffix string)` which takes in a specified suffix and crafts the webhook's name, so we don't end up with different re-definitions of the validator's name.

pkg/resources/validation/namespace/namespace.go
```diff
@@ -38,17 +40,63 @@ func (v *Validator) GVR() schema.GroupVersionResource {
 }
 
 // Operations returns list of operations handled by this validator.
-func (v *Validator) Operations() []admissionregistrationv1.OperationType {
```
MbolotSuse, line 41: I don't think that this makes sense as part of the interface any longer. Each `adm1.ValidatingWebhook` can cover it's own set of operations, and there's no reason why they need to share the same types of operations. 

In this implementation, which is the first actual use of using this interface to produce multiple  `adm1.ValidatingWebhook`s, we produce two different webhooks with two different sets of operations. If this produces some off logic in the very first use case, I think it's better to reconsider it.

MbolotSuse (changes requested):
I think that the design of the ValidatingAdmissionHandler interface needs some work. More specifically, I think that the following approach would work better:

- Keep the existing ValidatingWebhook method as (i.e. don't return a slice, just return one value)
- Add a new method to the `WebhookHandler` interface allowing a user to specify objectSelectors
  - We can also just modify the optionSelectors as part of the return from the `ValidatingWebhook` method from the namespace package
- Add a new Validator to the `namespace` package which validates `kube-system` on create only, and has the appropriate object selector and fail policy of ignore
- Add a new Validator to the `namespace` pacakge which validates non-kube-system on create only, and has the appropriate object selector and fail policy of fail
- Adjust the existing validator to only handle update operations and exclude no namespaces
- Since these 3 validators call out to the same admit, we can put that in it's own file and just have each of the public-facing validators call that function when triggered.
- If we don't want to construct each of the 3 validators individually, we can have a package level `namespaces.GetValidators()` which does this - that way we keep the interface intact and keep the ergonomics of the call. 


I think that the proposed approach (in this comment) has the following advantages over the current approach (the approach in the PR):
- The current approach uses the `Operation` call to determine which operations work for the webhoook, requiring a slice-modification when we need to construct multiple webhooks for the same validator which have different operations
- The proposed approach keeps our interface (which is well-designed) intact, and makes webhook formulation explicit.

tests/integration/failPolicy_test.go
```diff
@@ -0,0 +1,79 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
+	err = client.Create(ctx, "", validCreateObj, nil, v1.CreateOptions{})
+	m.Require().True(errors.IsInternalError(err), "Webhook should fail with service unavailable when the webhook is down instead we received :%v", err)
+
+	// attempt to clean up namespace if the create went through
+	defer client.Delete(ctx, "", testNamespace, v1.DeleteOptions{})
+
+	err = client.Update(ctx, "", validCreateObj, nil, v1.UpdateOptions{})
+	m.Require().True(errors.IsInternalError(err) || errors.IsNotFound(err), "Webhook should fail with service unavailable when the webhook is down instead we received :%v", err)
```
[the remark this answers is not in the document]
KevinJoiner: This works. It is just failing because of a race condition in CI where the webhook is still up when the test is run.

scripts/integration-test
```diff
@@ -47,4 +47,9 @@ while ! kubectl rollout status -w -n cattle-system deploy/rancher-webhook; do
     sleep 2
 done
 
-./bin/rancher-webhook-integration.test -test.v
+./bin/rancher-webhook-integration.test -test.v -test.run IntegrationTest
+
+kubectl scale deploy rancher-webhook -n cattle-system --replicas=0
+sleep 1
```
MbolotSuse, line 53: Nit:

You can wait for the webhook to be fully down by using `kubectl wait`, which is a bit more of a "sure thing" than this sleep command.

This command should do the trick:

```bash
kubectl wait pods -l app=rancher-webhook --for=delete -n cattle-system
```

tests/integration/failPolicy_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type FailurePolicySuite struct {
+	suite.Suite
+	clientFactory client.SharedClientFactory
+}
+
+// TestFailurePolicyTest should be run only when the webhook is not running.
```
MbolotSuse, line 27: Nit: You could enforce this condition with a require statement somewhere which fails the test (with an explicit message) if any of the webhook pods are still up.

Totally optional though.

tests/integration/failPolicy_test.go
MbolotSuse, line 70: Nit: I don't think that `NotFound` would be a valid condition for this. That would indicate to me that the webhook did not reject the request, and instead this went through. Meaning that the fail condition was `ignore`, when we want it to be `fail` on update calls.

MbolotSuse (approved):
I like the new design, it looks good.

I have a few test nits which would be good to fix, but nothing that will hold the PR up.

scripts/integration-test
KevinJoiner: Thank you. I couldn't remember what to wait on. I tried looking at the deployment status, but that returned before this.

tests/integration/failPolicy_test.go
KevinJoiner: Updated.

MbolotSuse (dismissed):

pkg/admission/admission.go
```diff
@@ -264,3 +273,12 @@ func ResponseBadRequest(message string) *admissionv1.AdmissionResponse {
 		Allowed: false,
 	}
 }
+
+// CreateWebhookName returns a new name for the give webhook with the given suffix.
```
pjbgf, line 277:
```suggestion
// CreateWebhookName returns a new name for the given webhook with the given suffix.
```

pkg/admission/admission.go
```diff
@@ -51,20 +55,24 @@ type WebhookHandler interface {
...
 
 // MutatingAdmissionHandler is a handler used for creating a MutatingAdmission Webhook.
 type MutatingAdmissionHandler interface {
 	WebhookHandler
 
-	// MutatingWebhook returns the configuration information for a ValidatingWebhook.
-	// This functions allows MutatingAdmissionHandler to perform and modifications to the default configuration if needed.
+	// MutatingWebhook returns a list of configurations to route to this handler.
+	//
+	// MutatingWebhook functions allows MutatingAdmissionHandler to perform and modifications to the default configuration if needed.
```
pjbgf, line 72: Maybe?
```suggestion
	// MutatingWebhook functions allows MutatingAdmissionHandler to perform modifications to the default configuration if needed.
```

tests/integration/failPolicy_test.go
```diff
@@ -0,0 +1,96 @@
...
+	logrus.Infof("Setting up test with KUBECONFIG=%s", kubeconfigPath)
+	restCfg, err := kubeconfig.GetNonInteractiveClientConfig(kubeconfigPath).ClientConfig()
+	m.Require().NoError(err, "Failed to clientFactory config")
+	m.clientFactory, err = client.NewSharedClientFactoryForConfig(restCfg)
+	m.Require().NoError(err, "Failed to create clientFactory Interface")
+
+	schemes.Register(v3.AddToScheme)
+	schemes.Register(provisioningv1.AddToScheme)
+	schemes.Register(corev1.AddToScheme)
+
```
pjbgf, line 45:
```suggestion
```

tests/integration/failPolicy_test.go
```diff
@@ -0,0 +1,96 @@
...
+	err = client.Update(ctx, "", validCreateObj, nil, v1.UpdateOptions{})
+	m.Require().True(errors.IsInternalError(err), "Webhook should fail with service unavailable when the webhook is down instead we received :%v", err)
+
+	validCreateObj.Name = "kube-system"
+	err = client.Create(ctx, "", validCreateObj, nil, v1.CreateOptions{})
+	m.Require().True(errors.IsAlreadyExists(err), "Webhook should fail to create kube-system with an already exist error instead we received :%v", err)
+
+	err = client.Update(ctx, "", validCreateObj, nil, v1.UpdateOptions{})
+	m.Require().True(errors.IsInternalError(err), "Webhook should fail to update kube-system namespace with service unavailable when the webhook is down instead we received :%v", err)
+
```
pjbgf, line 95:
```suggestion
```

pkg/resources/validation/namespace/namespace.go
```diff
@@ -38,17 +39,49 @@ func (v *Validator) GVR() schema.GroupVersionResource {
...
+			{
+				Key:      corev1.LabelMetadataName,
+				Operator: metav1.LabelSelectorOpNotIn,
+				Values:   []string{"kube-system"},
+			},
+		},
+	}
+
+	// kubeSystemOnlyWebhook is a separate webhook configuration that routes to this handler only if the namespace is equal to kube-system.
+	// This configuration differs from above because it allows create request to go through while the webhook is down if and only if the namespace is kube-system
```
pjbgf, line 70:
```suggestion
	// This configuration differs from above because it allows create request to go through while the webhook is down if and only if the namespace is kube-system.
```

pkg/resources/validation/namespace/namespace.go
```diff
@@ -38,17 +39,49 @@ func (v *Validator) GVR() schema.GroupVersionResource {
...
 }
 
 // ValidatingWebhook returns the ValidatingWebhook used for this CRD.
-func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) *admissionregistrationv1.ValidatingWebhook {
+func (v *Validator) ValidatingWebhook(clientConfig admv1.WebhookClientConfig) []admv1.ValidatingWebhook {
 	// Note that namespaces are actually CLUSTER scoped
-	return admission.NewDefaultValidatingWebhook(v, clientConfig, admissionregistrationv1.ClusterScope)
+
+	// standardWebhook validates all operations other than create on all namespaces.
+	standardWebhook := admission.NewDefaultValidatingWebhook(v, clientConfig, admv1.ClusterScope, []admv1.OperationType{admv1.Update})
```
pjbgf, line 54: The comment says "all operations other than create", but the operation type is only `Update`. Is that intentional? Don't we need to also add `Delete` here? Do we need to add a test case to cover this?

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/admissionregistration/v1/types.go#L536-L542

pjbgf (commented):
A few nits and questions, but overall lgtm.

pkg/resources/validation/namespace/namespace.go
KevinJoiner: I can update the comment. I meant all operations the handler cares about, not all available operations. This handler does not care about the deletion of a namespace since its goal is to check the PSA labels on active namespaces. 

We don't need a test case for deletion since this handler will never be sent a deletion request. This is controlled by another package that filters requests based on the specified options in this function https://github.com/rancher/webhook/blob/release/v0.3/pkg/resources/validation/namespace/namespace.go#L41-L44

tests/integration/failPolicy_test.go
KevinJoiner: Fixed.

pkg/admission/admission.go
KevinJoiner: Fixed.

pkg/resources/validation/namespace/namespace.go
KevinJoiner: Fixed.

pjbgf (approved):

PR #215, opened 2023-04-18
feat(mcm): deny change from true to false

FrankYang0529:
issue: https://github.com/harvester/harvester/issues/2679

Prevent users from changing the `multi-cluster-management` feature from `true` to `false`.

### DIFF
--- a/pkg/resources/validation/feature/feature.go
+++ b/pkg/resources/validation/feature/feature.go
@@ -49,6 +49,18 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
 		return nil, err
 	}
 
+	if oldFeature.Name == "multi-cluster-management" && !isValidMCMFeatureValue(oldFeature.Status.Default, oldFeature.Spec.Value, newFeature.Spec.Value) {
+		return &admissionv1.AdmissionResponse{
+			Result: &metav1.Status{
+				Status:  "Failure",
+				Message: "mcm feature flag cannot be changed from true to false",
+				Reason:  metav1.StatusReasonInvalid,
+				Code:    http.StatusBadRequest,
+			},
+			Allowed: false,
+		}, nil
+	}
+
 	if !isValidFeatureValue(newFeature.Status.LockedValue, oldFeature.Spec.Value, newFeature.Spec.Value) {
 		return &admissionv1.AdmissionResponse{
 			Result: &metav1.Status{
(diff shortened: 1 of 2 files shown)


pkg/resources/validation/feature/feature.go
```diff
@@ -49,6 +49,18 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
 		return nil, err
 	}
 
+	if oldFeature.Name == "multi-cluster-management" && !isValidMCMFeatureValue(oldFeature.Status.Default, oldFeature.Spec.Value, newFeature.Spec.Value) {
```
bk201, line 52: Is there a constant definition of this flag?

bk201:
Part of SURE-6068

pkg/resources/validation/feature/feature.go
FrankYang0529: I can only find this line in Rancher repo. Other files are YAML.

https://github.com/rancher/rancher/blob/dc282bc553b17277c5e7ae1b871857f049842445/pkg/features/feature.go#L33

pkg/resources/validation/feature/feature.go
```diff
@@ -49,6 +49,18 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
+				Status:  "Failure",
+				Message: "mcm feature flag cannot be changed from true to false",
+				Reason:  metav1.StatusReasonInvalid,
+				Code:    http.StatusBadRequest,
+			},
+			Allowed: false,
+		}, nil
+	}
+
 	if !isValidFeatureValue(newFeature.Status.LockedValue, oldFeature.Spec.Value, newFeature.Spec.Value) {
```
MbolotSuse, line 64: Can you elaborate more on why this check (preventing a locked value from changing) does not suit your use case? I.e., can harvester simply lock the mcm feature with the `true` value to meet this use case?

MbolotSuse (changes requested):
One overall comment:

Please add tests to your PR. We have been trying to add more focus to the webhook, so we'd like tests for any new additions. You can see an example of how to write these here: https://github.com/rancher/webhook/blob/release/v0.3/pkg/resources/validation/namespace/namespace_test.go. 

If you can also elaborate on why locking the feature doesn't work for your use case, that would be helpful to me.

pkg/resources/validation/feature/feature.go
FrankYang0529: From this [comment](https://suse.slack.com/archives/C02CX064EBX/p1680123611416709?thread_ts=1679454772.476669&cid=C02CX064EBX), I think changing the `multi-cluster-management` feature from `true` to `false` is an invalid operation. Also, users can still change `lockedValue` by editing YAML, so it's better to use a webhook to deny the operation.

I also added test cases. Thanks.

KevinJoiner (changes requested):
@FrankYang0529 
> Also, users can still change lockedValue by editing YAML, so it's better to use a webhook to deny the operation.

I may disagree with this conclusion. Based on your situation, `lockedValue` seems to fit the use case of "Stopping a user from accidentally toggling a feature that should not be changed."
If my assumption of your use case is wrong, please give an example of how your needs differ.

If `lockedValue` is set, the user would have to change both `lockedValue` and the value of the feature. This would be an intentional action performed outside of the UI, which is currently not a recommended workflow for users.

If we are concerned with users editing locked features via kubectl, we should create a separate issue to cover that discussion and consider adding a check to make `lockedValue` immutable.

Overall this is not a bad idea. My main concern is that this check feels too specific for something that is a separate issue outside the scope of Harvester and Rancher integration.

FrankYang0529:
@KevinJoiner Thanks for the suggestion. I agree they are different issues. I will close this one. Thank you.

PR #217, opened 2023-04-19
Prevent unset of fleetWorkspaceName in mgmt cluster validator

bfbachmann:
## Problem
If users remove the `fleetWorkspaceName` field on management cluster objects, those clusters get deleted. This is generally an unintended consequence for the user.
 
## Solution
Update the management cluster validation webhook logic to return a failure if the `fleetWorkspaceName` is being unset.
 
## Testing
I tested this manually by building a new webhook image locally, pushing it to a personal Docker image repo, and updating the deployment on my local webhook deployment to point to my new image. Then I tried removing `fleetWorkspaceName` on management cluster objects in my local cluster and verified that the edits were rejected.


pkg/resources/validation/cluster/cluster.go
```diff
@@ -115,7 +115,19 @@ func (v *Validator) validateFleetPermissions(request *admission.Request) (*admis
...
+				Message: "field FleetWorkspaceName cannot be unset",
+				Reason:  metav1.StatusReasonInvalid,
+				Code:    http.StatusBadRequest,
+			},
+			Allowed: false,
+		}, nil
+	} else if oldCluster.Spec.FleetWorkspaceName == newCluster.Spec.FleetWorkspaceName {
 		return &admissionv1.AdmissionResponse{
 			Allowed: true,
 		}, nil
```
KevinJoiner, line 133: Since we are returning in the above condition, we do not need an else statement.
```suggestion
	if newCluster.Spec.FleetWorkspaceName == "" && oldCluster.Spec.FleetWorkspaceName != "" {
		return &admissionv1.AdmissionResponse{
			Result: &metav1.Status{
				Status:  "Failure",
				Message: "field FleetWorkspaceName cannot be unset",
				Reason:  metav1.StatusReasonInvalid,
				Code:    http.StatusBadRequest,
			},
			Allowed: false,
		}, nil
	}
	if oldCluster.Spec.FleetWorkspaceName == newCluster.Spec.FleetWorkspaceName {
		return admission.ResponseAllowed(), nil
```

KevinJoiner, line 133: Since we are returning in the above condition, we do not need an else statement.
```suggestion
	if newCluster.Spec.FleetWorkspaceName == "" && oldCluster.Spec.FleetWorkspaceName != "" {
		return &admissionv1.AdmissionResponse{
			Result: &metav1.Status{
				Status:  "Failure",
				Message: "field FleetWorkspaceName cannot be unset",
				Reason:  metav1.StatusReasonInvalid,
				Code:    http.StatusBadRequest,
			},
			Allowed: false,
		}, nil
	}
	if oldCluster.Spec.FleetWorkspaceName == newCluster.Spec.FleetWorkspaceName {
		return admission.ResponseAllowed(), nil
```

KevinJoiner (changes requested):
Thanks for adding the unit test.

KevinJoiner (dismissed):

snasovich (changes requested):
Putting a "Request changes" while we're looking into the implications of this change to the functionality of Fleet Workspaces (more specifically deleting Fleet Workspace with existing RKE1 clusters in it).

pkg/resources/validation/cluster/cluster_test.go
```diff
@@ -0,0 +1,101 @@
...
+			newCluster:    v3.Cluster{Spec: v3.ClusterSpec{FleetWorkspaceName: "fleet-default"}},
+			operation:     admissionv1.Update,
+			expectAllowed: true,
+		},
+		"Delete": {
+			oldCluster:    v3.Cluster{Spec: v3.ClusterSpec{FleetWorkspaceName: "fleet-default"}},
+			operation:     admissionv1.Delete,
+			expectAllowed: true,
+		},
+	} {
```
KevinJoiner, line 71: For table-driven tests, the standard format is first to define a slice of tests and then loop over that slice. There are some examples in our code, but here is an article to reference as well, Though you don't need to define an anonymous struct the like the example.

https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
``` golang
func TestSplit(t *testing.T) {
    tests := []struct {
        input string
        sep   string
        want  []string
    }{
        {input: "a/b/c", sep: "/", want: []string{"a", "b", "c"}},
        {input: "a/b/c", sep: ",", want: []string{"a/b/c"}},
        {input: "abc", sep: "/", want: []string{"abc"}},
    } 

    for _, tc := range tests {
        got := Split(tc.input, tc.sep)
        if !reflect.DeepEqual(tc.want, got) {
            t.Fatalf("expected: %v, got: %v", tc.want, got)
        }
    }
}
```

KevinJoiner (changes requested):

snasovich (approved):

snasovich:
> Putting a "Request changes" while we're looking into the implications of this change to the functionality of Fleet Workspaces (more specifically deleting Fleet Workspace with existing RKE1 clusters in it).

These have been addressed by ensuring validation only runs on Update (it was unintentionally running on Delete).

KevinJoiner (changes requested):
The code looks good just need to squash the commits.
It looks like you were trying to pull in changes from what is on HEAD. In the future, you can use a command like `git rebase release/v0.3` to accomplish this without the merge commits.

snasovich:
> The code looks good just need to squash the commits. It looks like you were trying to pull in changes from what is on HEAD. In the future, you can use a command like `git rebase release/v0.3` to accomplish this without the merge commits.

@KevinJoiner ,  can we just "Squash and merge" instead?

KevinJoiner:
@snasovich Yeah, we can do that. I generally ask to squash beforehand because it is not uncommon for people to forget when merging.

KevinJoiner (approved):

PR #221, opened 2023-04-21
Updating PR template to match Rancher's

bfbachmann:
## Problem
PR template doesn't match Rancher's standard template.
 
## Solution
Updated PR template


KevinJoiner:
I don't think we need to match Rancher's template at this time. The current base should be sufficient.

PR #225, opened 2023-04-28
Changing package structure to reflect group name

MbolotSuse:
Changes the package structure to separate by group/version rather than mutator vs validator. Also moves some code to a new "common" package for easy re-use.

## Issue: 
N/A
 
## Problem
The webhook currently separates its package structure into `mutators` and `validators`. This makes it difficult to share functionality across mutating/validating webhooks which use the same type, and makes it difficult to identify which files are altering a particular CRD.
 
## Solution
The package structure has been changed to group by resource, and common code used by validators/migrators have been moved to the `common` package. 
 
## Testing
N/A


JonCrowther (approved):
LGTM, just made one comment about updating the README

KevinJoiner (approved):

JonCrowther (approved):

KevinJoiner (approved):

PR #228, opened 2023-05-03
Port remaining PRTB validation rules from Norman

maxsokolovsky:
## Problem
Remaining validation logic from Norman needs to be ported over to the Webhook.
 
## Solution
1. Validators
exists = Role template must exist.
exists = Role template must be unlocked.
**new  = Role template must belong to its appropriate context (project).**
exists = Either user/userPrincipal OR group/groupPrincipal must exist.

2. Struct tags
        UserName           norman:"noupdate,type=reference[user]"
        UserPrincipalName  norman:"noupdate,type=reference[principal]"
        GroupName          norman:"noupdate,type=reference[group]"
        GroupPrincipalName norman:"noupdate,type=reference[principal]"
        ProjectName        norman:"required,noupdate,type=reference[project]"
        RoleTemplateName   norman:"required,noupdate,type=reference[roleTemplate]"
**new =   ServiceAccount     norman:"nocreate,noupdate"** - will only implement noupdate

3. Store
rtb_store.go Create method - will use a controller
 
4. Formatters: none

### DIFF
--- a/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
+++ b/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
@@ -1,9 +1,9 @@
-// Package projectroletemplatebinding is used for validating projectroletemplatebinding admission request.
+// Package projectroletemplatebinding is used for validating ProjectRoleTemplateBinding admission requests.
 package projectroletemplatebinding
 
 import (
+	"errors"
 	"fmt"
-	"net/http"
 	"strings"
 
 	apisv3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
@@ -14,8 +14,8 @@ import (
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	k8validation "k8s.io/kubernetes/pkg/registry/rbac/validation"
 	"k8s.io/utils/trace"
 )
@@ -71,28 +71,21 @@ type admitter struct {
 	roleTemplateResolver *auth.RoleTemplateResolver
 }
 
-// Admit is the entrypoint for the validator. Admit will return an error if it unable to process the request.
-// If this function is called without NewValidator(..) calls will panic.
+// Admit is the entrypoint for the validator. Admit will return an error if it's unable to process the request.
+// If this method is called on a nil Validator, it panics.
 func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResponse, error) {
 	listTrace := trace.New("projectRoleTemplateBindingValidator Admit", trace.Field{Key: "user", Value: request.UserInfo.Username})
 	defer listTrace.LogIfLong(admission.SlowTraceDuration)
 
+	fieldPath := field.NewPath("projectroletemplatebinding")
+
 	if request.Operation == admissionv1.Update {
 		oldPRTB, newPRTB, err := objectsv3.ProjectRoleTemplateBindingOldAndNewFromRequest(&request.AdmissionRequest)
 		if err != nil {
-			return nil, fmt.Errorf("failed to decode PRTB objects from request: %w", err)
+			return nil, fmt.Errorf("failed to decode old and new PRTB objects from request: %w", err)
 		}
-
-		if err = validateUpdateFields(oldPRTB, newPRTB); err != nil {
-			return &admissionv1.AdmissionResponse{
-				Result: &metav1.Status{
-					Status:  "Failure",
-					Message: err.Error(),
-					Reason:  metav1.StatusReasonBadRequest,
-					Code:    http.StatusBadRequest,
-				},
-				Allowed: false,
-			}, nil
+		if err := validateUpdateFields(oldPRTB, newPRTB, fieldPath); err != nil {
+			return admission.ResponseBadRequest(err.Error()), nil
 		}
 	}
 
@@ -102,16 +95,12 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 	}
 
 	if request.Operation == admissionv1.Create {
-		if err = a.validateCreateFields(prtb); err != nil {
-			return &admissionv1.AdmissionResponse{
-				Result: &metav1.Status{
-					Status:  "Failure",
-					Message: err.Error(),
-					Reason:  metav1.StatusReasonBadRequest,
-					Code:    http.StatusBadRequest,
-				},
-				Allowed: false,
-			}, nil
+		var fieldErr *field.Error
+		if err := a.validateCreateFields(prtb, fieldPath); err != nil {
+			if errors.As(err, &fieldErr) {
+				return admission.ResponseBadRequest(err.Error()), nil
+			}
+			return nil, fmt.Errorf("failed to validate fields on create: %w", err)
 		}
 	}
 
@@ -152,51 +141,73 @@ func clusterFromProject(project string) (string, string) {
 }
 
 // validUpdateFields checks if the fields being changed are valid update fields.
-func validateUpdateFields(oldPRTB, newPRTB *apisv3.ProjectRoleTemplateBinding) error {
-	var invalidFieldName string
+func validateUpdateFields(oldPRTB, newPRTB *apisv3.ProjectRoleTemplateBinding, fieldPath *field.Path) *field.Error {
+	const reason = "field is immutable"
 	switch {
 	case oldPRTB.RoleTemplateName != newPRTB.RoleTemplateName:
-		invalidFieldName = "referenced roleTemplate"
+		return field.Invalid(fieldPath.Child("roleTemplateName"), newPRTB.RoleTemplateName, reason)
 	case oldPRTB.ProjectName != newPRTB.ProjectName:
-		invalidFieldName = "projectName"
+		return field.Invalid(fieldPath.Child("projectName"), newPRTB.ProjectName, reason)
 	case oldPRTB.UserName != newPRTB.UserName && oldPRTB.UserName != "":
-		invalidFieldName = "userName"
+		return field.Invalid(fieldPath.Child("userName"), newPRTB.UserName, reason)
 	case oldPRTB.UserPrincipalName != newPRTB.UserPrincipalName && oldPRTB.UserPrincipalName != "":
-		invalidFieldName = "userPrincipalName"
+		return field.Invalid(fieldPath.Child("userPrincipalName"), newPRTB.UserPrincipalName, reason)
 	case oldPRTB.GroupName != newPRTB.GroupName && oldPRTB.GroupName != "":
-		invalidFieldName = "groupName"
+		return field.Invalid(fieldPath.Child("groupName"), newPRTB.GroupName, reason)
 	case oldPRTB.GroupPrincipalName != newPRTB.GroupPrincipalName && oldPRTB.GroupPrincipalName != "":
-		invalidFieldName = "groupPrincipalName"
+		return field.Invalid(fieldPath.Child("groupPrincipalName"), newPRTB.GroupPrincipalName, reason)
 	case (newPRTB.GroupName != "" || oldPRTB.GroupPrincipalName != "") && (newPRTB.UserName != "" || oldPRTB.UserPrincipalName != ""):
-		invalidFieldName = "both user and group"
+		return field.Forbidden(fieldPath,
+			"binding must target either a user [userName]/[userPrincipalName] OR a group [groupName]/[groupPrincipalName]")
+	case newPRTB.ServiceAccount != "":
+		return field.Forbidden(fieldPath.Child("serviceAccount"), "update is not allowed")
 	default:
 		return nil
 	}
-
-	return fmt.Errorf("cannot update %s for clusterRoleTemplateBinding %s: %w", invalidFieldName, oldPRTB.Name, admission.ErrInvalidRequest)
 }
 
 // validateCreateFields checks if all required fields are present and valid.
-func (a *admitter) validateCreateFields(newPRTB *apisv3.ProjectRoleTemplateBinding) error {
+func (a *admitter) validateCreateFields(newPRTB *apisv3.ProjectRoleTemplateBinding, fieldPath *field.Path) error {
 	hasUserTarget := newPRTB.UserName != "" || newPRTB.UserPrincipalName != ""
 	hasGroupTarget := newPRTB.GroupName != "" || newPRTB.GroupPrincipalName != ""
+	hasServiceAccountTarget := newPRTB.ServiceAccount != ""
 
-	if (hasUserTarget && hasGroupTarget) || (!hasUserTarget && !hasGroupTarget) {
-		return fmt.Errorf("binding must target either a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]: %w", admission.ErrInvalidRequest)
+	if !onlyOneTrue(hasUserTarget, hasGroupTarget, hasServiceAccountTarget) {
+		return field.Forbidden(fieldPath,
+			"binding must target only a user [userName]/[userPrincipalName] OR a group [groupName]/[groupPrincipalName] OR a [serviceAccount]")
 	}
 
 	if newPRTB.ProjectName == "" {
-		return fmt.Errorf("binding must have field projectName set: %w", admission.ErrInvalidRequest)
+		return field.Required(fieldPath.Child("projectName"), "")
+	}
+
+	if newPRTB.RoleTemplateName == "" {
+		return field.Required(fieldPath.Child("roleTemplateName"), "")
 	}
 
 	roleTemplate, err := a.roleTemplateResolver.RoleTemplateCache().Get(newPRTB.RoleTemplateName)
 	if err != nil {
-		return fmt.Errorf("unknown reference roleTemplate '%s': %w", newPRTB.RoleTemplateName, err)
+		return err
 	}
 
 	if roleTemplate.Locked {
-		return fmt.Errorf("referenced role '%s' is locked and cannot be assigned: %w", roleTemplate.DisplayName, admission.ErrInvalidRequest)
+		return field.Forbidden(fieldPath.Child("roleTemplate"), fmt.Sprintf("referenced role '%s' is locked and cannot be assigned", roleTemplate.DisplayName))
+	}
+
+	const projectContext = "project"
+	if roleTemplate.Context != projectContext {
+		return field.NotSupported(fieldPath.Child("roleTemplate", "context"), roleTemplate.Context, []string{projectContext})
 	}
 
 	return nil
+}
+
+func onlyOneTrue(values ...bool) bool {
+	var trueCount int
+	for _, v := range values {
+		if v {
+			trueCount++
+		}
+	}
+	return trueCount == 1
 }
--- a/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator_test.go
+++ b/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator_test.go
@@ -33,12 +33,13 @@ const (
 
 type ProjectRoleTemplateBindingSuite struct {
 	suite.Suite
-	adminRT         *apisv3.RoleTemplate
-	readNodesRT     *apisv3.RoleTemplate
-	lockedRT        *apisv3.RoleTemplate
-	adminCR         *rbacv1.ClusterRole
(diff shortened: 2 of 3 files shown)


KevinJoiner:
Can you update the validator to use field errors for validation? similar to what is done in this file.
https://github.com/rancher/webhook/blob/67a765634fd67be256a3249f1d16b4c23a19dbcc/pkg/resources/validation/roletemplate/roletemplate.go#L188-L197
and then use something like this to create the response with the validation error
``` golang
if fieldErr != nil {
}
```

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -141,50 +145,60 @@ func clusterFromProject(project string) (string, string) {
 }
 
 // validUpdateFields checks if the fields being changed are valid update fields.
-func validateUpdateFields(oldPRTB, newPRTB *apisv3.ProjectRoleTemplateBinding) error {
-	var invalidFieldName string
+func validateUpdateFields(oldPRTB, newPRTB *apisv3.ProjectRoleTemplateBinding, fieldPath *field.Path) *field.Error {
 	switch {
 	case oldPRTB.RoleTemplateName != newPRTB.RoleTemplateName:
-		invalidFieldName = "referenced roleTemplate"
+		return field.Forbidden(fieldPath.Child("roleTemplateName"), "")
```
KevinJoiner, line 151: Add a message for your `Forbidden` errors telling the user what they did wrong.
```suggestion
		return field.Forbidden(fieldPath.Child("roleTemplateName"), "can not update")
```

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -141,50 +145,60 @@ func clusterFromProject(project string) (string, string) {
...
 
 	roleTemplate, err := v.roleTemplateResolver.RoleTemplateCache().Get(newPRTB.RoleTemplateName)
 	if err != nil {
-		return fmt.Errorf("unknown reference roleTemplate '%s': %w", newPRTB.RoleTemplateName, err)
+		return field.Invalid(fieldPath.Child("roleTemplate"), roleTemplate, "")
 	}
 
 	if roleTemplate.Locked {
-		return fmt.Errorf("referenced role '%s' is locked and cannot be assigned: %w", roleTemplate.DisplayName, admission.ErrInvalidRequest)
+		return field.NotSupported(fieldPath.Child("roleTemplate"), roleTemplate.Locked, []string{"false"})
```
KevinJoiner, line 196: I think we should use forbidden here.
> // ErrorTypeForbidden is used to report valid (as per formatting rules)
	// values which would be accepted under some conditions, but which are not
	// permitted by the current conditions (such as security policy)

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -141,50 +145,60 @@ func clusterFromProject(project string) (string, string) {
...
 
 	if newPRTB.ProjectName == "" {
-		return fmt.Errorf("binding must have field projectName set: %w", admission.ErrInvalidRequest)
+		return field.Required(fieldPath.Child("projectName"), "")
 	}
 
 	roleTemplate, err := v.roleTemplateResolver.RoleTemplateCache().Get(newPRTB.RoleTemplateName)
 	if err != nil {
-		return fmt.Errorf("unknown reference roleTemplate '%s': %w", newPRTB.RoleTemplateName, err)
+		return field.Invalid(fieldPath.Child("roleTemplate"), roleTemplate, "")
```
KevinJoiner, line 192: I was not handling this error case correctly in my first iteration. I think we should modify the function to return an error if the error is anything other than `ErrorNotFound`. Then, the returning function can use `errors.Is(...)` to determine if it is a `field.Error` that should be sent back in a denied response or an error in the handler.

KevinJoiner (changes requested):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
maxsokolovsky: What if we keep the logic and add the error text as the detail:
`return field.Invalid(fieldPath.Child("roleTemplate"), roleTemplate, err.Error())`

maxsokolovsky:
@KevinJoiner, is this part OK? I am slightly dubious on the `value` (the "combination" verbiage rather than some concrete value):
```
if (hasUserTarget && hasGroupTarget) || (!hasUserTarget && !hasGroupTarget) {
		return field.Invalid(fieldPath.Child("groupName", "groupPrincipalName", "userName", "userPrincipalName"),
			"combination of [userId]/[userPrincipalId] and [groupId]/[groupPrincipalId]",
			"binding must target either a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]")
}
```

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: Actually, after looking at the K8s code this is a more standard way of rejecting immutable fields
 ``` golang
 field.Invalid(fieldPath.Child("roleTemplateName"), newPRTB.RoleTemplateName, "field is immutable"))
 ```

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: This would still be undesired because we want the Admit function to return an error without a response object.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -141,50 +145,60 @@ func clusterFromProject(project string) (string, string) {
...
+	}
+
 	hasUserTarget := newPRTB.UserName != "" || newPRTB.UserPrincipalName != ""
 	hasGroupTarget := newPRTB.GroupName != "" || newPRTB.GroupPrincipalName != ""
 
 	if (hasUserTarget && hasGroupTarget) || (!hasUserTarget && !hasGroupTarget) {
-		return fmt.Errorf("binding must target either a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]: %w", admission.ErrInvalidRequest)
+		return field.Invalid(fieldPath.Child("groupName", "groupPrincipalName", "userName", "userPrincipalName"),
+			"combination of [userId]/[userPrincipalId] and [groupId]/[groupPrincipalId]",
+			"binding must target either a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]")
```
KevinJoiner, line 183: If you are validating a combination of fields we can just use the parent as the fieldPath
```suggestion
		return field.Forbidden(fieldPath, "binding must target either a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]")
```

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -91,16 +96,19 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
-					Code:    http.StatusBadRequest,
-				},
-				Allowed: false,
-			}, nil
+		if err := v.validateCreateFields(prtb, fieldPath); err != nil {
+			if errors.As(err, &fieldErr) {
+				return &admissionv1.AdmissionResponse{
+					Result: &metav1.Status{
+						Status:  "Failure",
+						Message: err.Error(),
```
JonCrowther, line 104: I think this should be
```suggestion
						Message: fieldErr.Error(),
```

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -60,23 +62,26 @@ func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.Webho
...
 
-		if err = validateUpdateFields(oldPRTB, newPRTB); err != nil {
+		if fieldErr = validateUpdateFields(oldPRTB, newPRTB, fieldPath); fieldErr != nil {
 			return &admissionv1.AdmissionResponse{
 				Result: &metav1.Status{
 					Status:  "Failure",
-					Message: err.Error(),
+					Message: fieldErr.Error(),
 					Reason:  metav1.StatusReasonBadRequest,
 					Code:    http.StatusBadRequest,
```
JonCrowther, line 86: Looking at the [validator.go](https://github.com/rancher/rancher/blob/1f014b7fcd734e9321555b613cc28b1b12bf8eeb/pkg/api/norman/customization/roletemplatebinding/validator.go#L77) from Norman, the various errors have specific error codes. @KevinJoiner Should we be passing these along to the admission control response?

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -141,50 +149,65 @@ func clusterFromProject(project string) (string, string) {
...
+	}
+
+	if newPRTB.RoleTemplateName == "" {
+		return field.Required(fieldPath.Child("roleTemplateName"), "")
 	}
 
 	roleTemplate, err := v.roleTemplateResolver.RoleTemplateCache().Get(newPRTB.RoleTemplateName)
 	if err != nil {
-		return fmt.Errorf("unknown reference roleTemplate '%s': %w", newPRTB.RoleTemplateName, err)
+		return err
```
JonCrowther, line 201: I don't follow why this can't also be a `fieldErr` returned? Something like:
```
return field.InternalError(fieldPath.Child("roleTemplate"), err)
```

That way we can avoid the `errors.As` check above.

JonCrowther (commented):
Reviewed validator.go. Will go over the tests at a later time

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: either works in the current form `err` would be more verbose if the field.Error was wrapped.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: We don't need to follow Norman errors, but we should take a look at using specific http error codes.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -152,50 +142,65 @@ func clusterFromProject(project string) (string, string) {
...
-
-	return fmt.Errorf("cannot update %s for clusterRoleTemplateBinding %s: %w", invalidFieldName, oldPRTB.Name, admission.ErrInvalidRequest)
 }
 
 // validateCreateFields checks if all required fields are present and valid.
-func (a *admitter) validateCreateFields(newPRTB *apisv3.ProjectRoleTemplateBinding) error {
+func (a *admitter) validateCreateFields(newPRTB *apisv3.ProjectRoleTemplateBinding, fieldPath *field.Path) error {
+	if newPRTB.ServiceAccount != "" {
+		return field.Forbidden(fieldPath.Child("serviceAccount"), "")
+	}
```
KevinJoiner, line 174: Why can't this be empty, and what is it being used for today?

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -152,50 +142,65 @@ func clusterFromProject(project string) (string, string) {
...
+			"binding must target either a user [userName]/[userPrincipalName] OR a group [groupName]/[groupPrincipalName]")
 	}
 
 	if newPRTB.ProjectName == "" {
-		return fmt.Errorf("binding must have field projectName set: %w", admission.ErrInvalidRequest)
+		return field.Required(fieldPath.Child("projectName"), "")
+	}
+
+	if newPRTB.RoleTemplateName == "" {
+		return field.Required(fieldPath.Child("roleTemplateName"), "")
```
KevinJoiner, line 189: I think Norman may allow for these values to be empty string.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -152,50 +142,65 @@ func clusterFromProject(project string) (string, string) {
...
 	}
 
 	if roleTemplate.Locked {
-		return fmt.Errorf("referenced role '%s' is locked and cannot be assigned: %w", roleTemplate.DisplayName, admission.ErrInvalidRequest)
+		return field.Forbidden(fieldPath.Child("roleTemplate"), "role template cannot be locked")
+	}
+
+	const projectContext = "project"
+	if roleTemplate.Context != projectContext {
+		return field.NotSupported(fieldPath.Child("roleTemplate", "context"), roleTemplate.Context, []string{projectContext})
```
KevinJoiner, line 203: I am unsure if we can do this today with the existing rancher behavior.

KevinJoiner (changes requested):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
maxsokolovsky: Norman has `required` in tags.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
maxsokolovsky: It's got a Norman tag: `norman:"nocreate,noupdate"`, so I am just emulating this here. Let me see what value it normally has with a real setup...

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
maxsokolovsky: That's fine. Again, it's just transferring over the current behavior...

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
maxsokolovsky: In a PRTB for a user, that `serviceAccount` field is missing. I am seeing only a couple of places where the property is read. 
https://github.com/maxsokolovsky/rancher/blob/96f5ed3c4faefa8580ab9a5ceae204c74c058f86/pkg/controllers/managementuserlegacy/monitoring/projectHandler.go#L351
Here, for example, it's used in managementuserlegacy as an indexer func.

I am not sure if this field is really used anymore.

maxsokolovsky: Maybe it used to be important for projects? @cmurphy, do you happen to know?

cmurphy: > Why can't this be empty, and what is it being used for today?

@KevinJoiner I think it's saying the opposite, it's saying it has to be empty (so it's useless)

> Maybe it used to be important for projects?

It's so you can create a binding with a service account as a subject: https://github.com/rancher/rancher/blob/5f98b0f28bc8a48366e9ced8f0a0fea3d826d90d/pkg/rbac/common.go#L46

It was introduced so that the monitoring app could create a service account and use rancher RBAC to manage its permissions: https://github.com/rancher/rancher/pull/19766

I'm guessing it used `norman:nocreate` in order to prevent users from using it, they just wanted the monitoring controller using it: https://github.com/rancher/rancher/pull/19766/files#diff-34975bed09fad198a231f28780f8194ba1129f987d56770a1cd79a5b00a15f85R329-R338

If the webhook forbids setting this field, that would actually break that controller.

I don't see a good reason to keep enforcing `nocreate` for this, just `noupdate` same as the other subject fields.

Seeing as this seems to only be needed for monitoring v1, maybe the new PRTB CRD could just drop it, and  delete that controller? But in the mean time I don't think the webhook needs to do anything special for it.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: @cmurphy 
> I think once it's marshalled into an object like it is at this point, it's ambiguous whether it was originally set or not.

 Yeah, I think we can ignore the required tag for anything that is not a pointer since everything else should have a valid default value after it is marshaled. The Norman code was slightly different because it was validating against the raw input, not the marshaled struct, which seems different from how upstream does validation.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: Sorry if this is current behavior. We can leave it. Based on the PR comment, I thought this was new.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: Interesting 👀  That all sounds good to me.

maxsokolovsky: Only ported the noupdate in this PR. But I am curious about how the controller isn't already broken, since Norman with its struct tags must currently enforce nocreate...

KevinJoiner: The controllers do not use Norman endpoints, only users.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -152,50 +142,65 @@ func clusterFromProject(project string) (string, string) {
 }
 
 // validUpdateFields checks if the fields being changed are valid update fields.
-func validateUpdateFields(oldPRTB, newPRTB *apisv3.ProjectRoleTemplateBinding) error {
-	var invalidFieldName string
+func validateUpdateFields(oldPRTB, newPRTB *apisv3.ProjectRoleTemplateBinding, fieldPath *field.Path) *field.Error {
+	const reason = "field is immutable"
```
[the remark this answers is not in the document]
KevinJoiner: Due to the support matrix reasons, we can't use it now.
Separately I don't love putting actual logic inside code comments. The immutability check is simple and not harmful, but I fear that will cause more people to leverage [CEL](https://github.com/google/cel-spec) to write more involved logic in comments which is obviously more error-prone. So it seems like it is easier to just say no CEL for validation.

But I am curious if you have any counter thoughts.

KevinJoiner (dismissed):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
maxsokolovsky: @cmurphy, @KevinJoiner - the webhook rejects the request if a PRTB doesn't have a user subject (user or user principal) AND doesn't have a group subject (group or group principal). What about ServiceAccount? It's also a kind of subject. This means there could be a user subject set AND a serviceAccount. Is this bad?

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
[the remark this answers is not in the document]
cmurphy: It would be nice to push as much validation as we can out of the webhook and into the API server. But with no way to write unit tests for code comments I agree with you that it's more error prone and probably not worth it.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -91,16 +96,19 @@ func (v *Validator) Admit(request *admission.Request) (*admissionv1.AdmissionRes
...
-				Result: &metav1.Status{
-					Status:  "Failure",
-					Message: err.Error(),
-					Reason:  metav1.StatusReasonBadRequest,
-					Code:    http.StatusBadRequest,
-				},
-				Allowed: false,
-			}, nil
+		if err := v.validateCreateFields(prtb, fieldPath); err != nil {
+			if errors.As(err, &fieldErr) {
```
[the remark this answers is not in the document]
cmurphy: Unresolving to highlight latest comment

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
cmurphy: Yes, ServiceAccount is a kind of subject and it is in conflict with User and Group as a subject. This needs to be fixed to validate that only one of the three types is set.

cmurphy (changes requested):
Requesting these changes

https://github.com/rancher/webhook/pull/228#discussion_r1264009761
https://github.com/rancher/webhook/pull/228/files#r1264004308

cmurphy (approved):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
KevinJoiner: good catch.

KevinJoiner (approved):

PR #247, opened 2023-06-06
Handle nil response in send error.

KevinJoiner:


KevinJoiner:
This was already fixed

PR #250, opened 2023-06-22
Add Validation on NodeDriver update/delete to prevent machine deletion

thatmidwesterncoder:
## Problem

Currently there is no validation on the `NodeDriver` object and when a node driver is disabled/deleted it currently removes the related CRDs which can result in machines being deleted and data loss.
 
## Solution

By adding a webhook for the `NodeDriver` resource we can validate that there are 0 rke1/rke2 machines using that driver before allowing the request to go through.
 
## Testing

1. Create a cluster with a node driver (digitalocean is what i used)
2. Go to the drivers page and try and disable the driver -> it will fail after this patch. 
3. Remove the cluster
4. Disabling the driver will go through as expected.


pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,185 @@
...
+
+// interface that holds the methods necessary to hit the k8s api and fetch
+// whether rke1/2 resources are deleted
+type k8sAPI interface {
+	rke1ResourcesDeleted(ctx context.Context, driver *v3.NodeDriver) (bool, error)
+	rke2ResourcesDeleted(ctx context.Context, driver *v3.NodeDriver) (bool, error)
+}
+
+type k8sAPIImpl struct {
+	rest rest.Interface
```
KevinJoiner, line 51: We don't need to use a raw `rest.Interface` for our lookups. Instead, we should use one of our generated clients and caches. You can generate new ones for RKE1 and RKE2 machines here https://github.com/rancher/webhook/blob/release/v0.3/pkg/codegen/main.go#L30

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// the check to see if the driver is being disabled is either when we're
+	// running a delete operation OR an update operation where the active flag
+	// toggles from true -> false
+	var disablingDriver bool
+	if request.Operation == admissionv1.Delete && old.Spec.Active {
+		disablingDriver = true
+	} else if !new.Spec.Active && old.Spec.Active {
+		disablingDriver = true
+	}
```
KevinJoiner, line 98: Add a bit of future-proofing here if someone were to go and add a new operation for separate logic, we want to make sure this evaluates the same.
```suggestion
	if request.Operation == admissionv1.Delete && old.Spec.Active ||  request.Operation == admissionv1.Update && !new.Spec.Active && old.Spec.Active {
		disablingDriver = true
	}
```

KevinJoiner (changes requested):
Nice job on this, I appreciate you taking the time to match the structure of the other webhook Validators. Could you also add a readme for this validator? This is something new we are doing to add documentation on all of our webhook validations. Here are some references on what that looks like https://github.com/rancher/webhook/blob/dev/v0.3/README.md#docs & https://github.com/rancher/webhook/blob/dev/v0.3/pkg/resources/management.cattle.io/v3/roletemplate/RoleTemplate.md

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
jakefhyde: Since these machines are dynamic, we cannot generate new clients for the rke1 or rke2 machines here. You can look here for an example of how we do this in rancher/rancher: https://github.com/rancher/rancher/blob/release/v2.7/pkg/controllers/capr/machineprovision/controller.go#L502

tl;dr we use `unstructured.Unstructured` with a dynamic client.

jakefhyde (changes requested):

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
KevinJoiner: Ahh, my bad I always forget we have those dynamic CRDs. I think we should replicate the same behavior you linked and use [dynamic](https://github.com/rancher/lasso/tree/master/pkg/dynamic) to keep our clients consistent throughout Rancher.

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,185 @@
...
+	old, new, err := objectsv3.NodeDriverOldAndNewFromRequest(&request.AdmissionRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	// hardcoded bypass - if the annotation is present we want to allow the
+	// driver to be disabled.
+	if old.Annotations[overrideAnnotation] == "true" || new.Annotations != nil && new.Annotations[overrideAnnotation] == "true" {
+		return admission.ResponseAllowed(), nil
+	}
```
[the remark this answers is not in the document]
thatmidwesterncoder: Sounds good, I removed the check for now.

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
thatmidwesterncoder: Updated to use the `dynamic.Controller` class, injected through the `NewValidator(...)` function.

thatmidwesterncoder:
I added a commit to fix the golangci-lint errors it found - hopefully thats ok. If it isn't I can split that out into another PR.

pkg/resources/management.cattle.io/v3/nodedriver/NodeDriver.md
```diff
@@ -0,0 +1,11 @@
...
+
+Note: checks only run if a node driver is being disabled or deleted
+
+### RKE1 Nodes
+
+RKE1 Nodes are under the `nodes.management.cattle.io` CRD and watched via a controller+cache. When a request for disable/deletion of the driver comes in the validation will list all of the nodes and look through the `status.NodeTemplateSpec.Driver` field to see if any of the nodes are using the NodeDriver that is being requested to be disabled, if any are found the request will be denied.
+
+### RKE2 Nodes via DynamicDriver
+
+RKE2 Nodes can be created via their own CRD's (think `digitaloceanmachine.rke-machine.cattle.io` for the DigitalOcean NodeDriver) and thus need to be looked up with the dynamic client. If there are _any_ instances of the CRD for that NodeDriver the request will be denied.
```
KevinJoiner, line 11: Apologies, we just added these docs requirements, so the expected format may need a better explanation. The goal of these docs is to let users know what the webhook is checking in their cluster. So we do not want implementation details only high-level explanations. We are trying to emulate this page that is provided by K8s https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#what-does-each-admission-controller-do
We also only need one section per Admitter. So this file would give an overview of how the validation stops users from disabling or deleting a driver that is in use.

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,163 @@
...
+	rke1ResourcesDeleted(driver *v3.NodeDriver) (bool, error)
+	rke2ResourcesDeleted(driver *v3.NodeDriver) (bool, error)
+}
+
+type k8sAPIImpl struct {
+	nodeCache controllersv3.NodeCache
+	dynamic   *dynamic.Controller
+}
+
+func NewValidator(nodeCache controllersv3.NodeCache, dynamic *dynamic.Controller) admission.ValidatingAdmissionHandler {
```
KevinJoiner, line 55: comment

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,163 @@
+package nodedriver
```
KevinJoiner, line 1: package comment
it can be something simple like
```suggestion
// Package nodedriver handles validation and creation for rke1 and rke2 nodedrivers.
package nodedriver
```

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,163 @@
...
+// whether rke1/2 resources are deleted
+type k8sAPI interface {
+	rke1ResourcesDeleted(driver *v3.NodeDriver) (bool, error)
+	rke2ResourcesDeleted(driver *v3.NodeDriver) (bool, error)
+}
+
+type k8sAPIImpl struct {
+	nodeCache controllersv3.NodeCache
+	dynamic   *dynamic.Controller
+}
```
KevinJoiner, line 53: Generally, we will put our required caches and clients directly on the object and then mock out the cache and the clients in our tests. `controllersv3.NodeCache` is already, so we would just need one for `*dynamic.Controller`. This may seem like just another way to implement what you already had, but when we mock the caches and clients we get better test coverage since the current implementation leaves `rke1ResourcesDeleted` and `rke2ResourcesDeleted` untested.
```suggestion
type admitter struct {
	nodeCache controllersv3.NodeCache
	dynamic   dynamicLister
}

type dynamicLister struct {
    List(gvk schema.GroupVersionKind, namespace string, selector labels.Selector) ([]runtime.Object, error) 
}
```

pkg/server/handlers.go
```diff
@@ -27,6 +28,7 @@ func Validation(clients *clients.Clients) ([]admission.ValidatingAdmissionHandle
 		provisioningCluster.NewProvisioningClusterValidator(clients),
 		machineconfig.NewValidator(),
 		nshandler.NewValidator(clients.K8s.AuthorizationV1().SubjectAccessReviews()),
+		nodedriver.NewValidator(clients.Management.Node().Cache(), clients.Dynamic),
```
KevinJoiner, line 31: @jakefhyde Do we use noddrivers in the downstream? If not, then this statement would need to be moved in the condition check on line 34.

pkg/resources/core/v1/secret/mutator.go
```diff
@@ -126,7 +126,7 @@ func (m *Mutator) admitCreate(secret *corev1.Secret, request *admission.Request)
 // admitDelete checks to see if there are any roleBindings owned by this secret which provide access to a role granting access to this secret
 // if so, it redacts the role so that it only grants delete access. This handles cases where users were given owner access to an individual secret
 // through a controller (like cloud-credentials), and delete the secret but keep the rbac
-func (m *Mutator) admitDelete(secret *corev1.Secret, request *admission.Request) (*admissionv1.AdmissionResponse, error) {
+func (m *Mutator) admitDelete(secret *corev1.Secret, _ *admission.Request) (*admissionv1.AdmissionResponse, error) {
```
KevinJoiner, line 129: If this is not implementing an interface, you can just delete the argument altogether.

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,163 @@
...
+// looking from the displayName of the driver.
+func (k *k8sAPIImpl) rke2ResourcesDeleted(driver *v3.NodeDriver) (bool, error) {
+	gvk := schema.GroupVersionKind{
+		Group:   "rke-machine.cattle.io",
+		Version: "v1",
+		Kind:    driver.Spec.DisplayName + "machine",
+	}
+	machines, err := k.dynamic.List(gvk, "", labels.Everything())
+	if err != nil {
+		return false, err
```
KevinJoiner, line 155: wrap error from external function call

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,163 @@
...
+	return admission.ResponseAllowed(), nil
+}
+
+// // RKE1
+// this one is a bit more clean since we're just looking at nodes with
+// the <displayname> provider
+func (k *k8sAPIImpl) rke1ResourcesDeleted(driver *v3.NodeDriver) (bool, error) {
+	nodes, err := k.nodeCache.List("", labels.Everything())
+	if err != nil {
+		return true, err
```
KevinJoiner, line 128: 1. wrap errors from external function calls.
2. return the default value of a bool when err is not nil.
```suggestion
		return false, err
```

KevinJoiner (changes requested):

pkg/server/handlers.go
jakefhyde: not in the downstream no

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -0,0 +1,158 @@
...
+	// toggles from true -> false
+	var disablingDriver bool
+	if (request.Operation == admissionv1.Delete && oldObject.Spec.Active) ||
+		(request.Operation == admissionv1.Update && !newObject.Spec.Active && oldObject.Spec.Active) {
+		disablingDriver = true
+	}
+
+	if !disablingDriver {
+		return admission.ResponseAllowed(), nil
+	}
```
KevinJoiner, line 98: nit: could be simplified but this is more styltic so if you think it is easier to read the way you had it there is no need to change.
```suggestion
	if !(request.Operation == admissionv1.Delete && oldObject.Spec.Active) &&
		!(request.Operation == admissionv1.Update && !newObject.Spec.Active && oldObject.Spec.Active) {
		return admission.ResponseAllowed(), nil
	}
```

pkg/server/handlers.go
KevinJoiner: Cool! @thatmidwesterncoder an you move this inside the `if clients.MultiClusterManagement ` condition since those validators only run in the local cluster.

KevinJoiner (changes requested):

KevinJoiner (approved):

jakefhyde (approved):

PR #253, opened 2023-06-23
Add docs and tests

maxsokolovsky:
There are open PRs with changes to PRTB, CRTB, Token and other resource validation and mutation. I will update the docs when they are merged.

### DIFF
--- a/pkg/resources/core/v1/secret/mutator_test.go
+++ b/pkg/resources/core/v1/secret/mutator_test.go
@@ -19,6 +19,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+var (
+	secretGVR = metav1.GroupVersionResource{Group: "", Version: "v1", Resource: "secrets"}
+	secretGVK = metav1.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"}
+)
+
 func Test_roleBindingIndexer(t *testing.T) {
 	testNamespace := "test-ns"
 	createBinding := func(roleRefKind string, ownerRefs ...metav1.OwnerReference) rbacv1.RoleBinding {
@@ -116,9 +121,7 @@ func Test_roleBindingIndexer(t *testing.T) {
 	}
 }
 
-func TestMutatorAdmit(t *testing.T) {
-	secretGVR := metav1.GroupVersionResource{Group: "", Version: "v1", Resource: "secrets"}
-	secretGVK := metav1.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"}
+func TestMutatorAdmitOnDelete(t *testing.T) {
 	const secretName = "test-secret"
 	const secretNamespace = "test-ns"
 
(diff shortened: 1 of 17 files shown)


pkg/resources/core/v1/secret/mutator_test.go
```diff
@@ -363,6 +366,70 @@ func TestMutatorAdmit(t *testing.T) {
...
+					UserInfo:        authenicationv1.UserInfo{Username: "test-user", UID: ""},
+					Object:          runtime.RawExtension{},
+					OldObject:       runtime.RawExtension{},
+				},
+			}
+			var err error
+			req.Object.Raw, err = json.Marshal(test.secret)
+			require.NoError(t, err)
+
+			ctrl := gomock.NewController(t)
```
JonCrowther, line 414: Minor, but I believe you could move all this ctrl portion above the for loop since none of it relies on the tests themselves

pkg/resources/management.cattle.io/v3/feature/validator_test.go
```diff
@@ -0,0 +1,144 @@
...
+					OldObject:       runtime.RawExtension{},
+				},
+			}
+			var err error
+			req.Object.Raw, err = json.Marshal(test.newFeature)
+			assert.NoError(t, err, "Failed to marshal new Feature while creating request")
+			req.OldObject.Raw, err = json.Marshal(test.oldFeature)
+			assert.NoError(t, err, "Failed to marshal old Feature while creating request")
+
+			if test.name == "bad request" {
```
JonCrowther, line 131: I'm not particularly fond of having this custom logic for one of the tests. Maybe add the `AdmissionRequest` as one of the test fields?

JonCrowther (changes requested):
Everything looks good, just some suggestions for the structure of the tests

pkg/resources/core/v1/secret/mutator_test.go
maxsokolovsky: Sure, I'll move this above the test loop, since each test case here requires the mutator to be constructed in the same way.

pkg/resources/management.cattle.io/v3/feature/validator_test.go
maxsokolovsky: Agreed, but the setup will be less clear in how the inputs are defined. I am testing features, so my inputs are v3.Feature values. Otherwise, I'd have to define `admissionv1.AdmissionRequest` for each test instead, which would be fine, since the method under test, Admit, actually takes a value of that type. But, again, readability might suffer a bit, since I really care about Features.
It's not ideal, but the test case itself is explicitly declared with the name and expected result.

JonCrowther (dismissed):

KevinJoiner (changes requested):

KevinJoiner (dismissed):
You should be able to update this to merge into the release branch once we get approvals and merge the dev branch into the release. https://github.com/rancher/webhook/pull/256

KevinJoiner (approved):

JonCrowther (approved):

PR #254, opened 2023-06-27
Validate `machinePool` names during Cluster creation

thatmidwesterncoder:
## Problem

When creating a cluster if the pool name is longer than 63 characters the machines never get deployed due to annotation length limits.
 
## Solution

Validating the machinePool name(s) on the RKE Config on the cluster object to be less than 64 characters will allow provisioning to continue.  
 
## Testing

1. Create a cluster with a pool name that is quite long
2. The cluster will fail to provision, the rancher logs will say the machineset resource is invalid due to an annotation name longer than 63 characters
3. Apply patch
4. The create request will be denied during the create process.


pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -131,3 +136,30 @@ func Test_isValidName(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateMachinePoolName(t *testing.T) {
```
jakefhyde, line 140: Everything else is LGTM, but could we add a few more test cases?

thatmidwesterncoder: Yep! Added a few more and changed to the struct approach like the test above.

jakefhyde (approved):
LGTM, nice work!

jiaqiluo (approved):
LGTM 💯

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -184,6 +188,29 @@ func (p *provisioningAdmitter) validateClusterName(request *admission.Request, r
...
+		return nil
+	}
+
+	if cluster.Spec.RKEConfig == nil {
+		return nil
+	}
+
+	for _, pool := range cluster.Spec.RKEConfig.MachinePools {
+		if len(pool.Name) > 63 {
+			response.Result = &metav1.Status{
```
maxsokolovsky, line 202: You might want to `break` after finding the first long pool name? There is no need to keep looping...

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -131,3 +136,60 @@ func Test_isValidName(t *testing.T) {
...
```
maxsokolovsky, line 140: Feel free to start the test with `t.Parallel()`.

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -131,3 +136,60 @@ func Test_isValidName(t *testing.T) {
...
+		{
+			name: "regular-string-test",
+			fail: false,
+		},
+	}
+
+	a := provisioningAdmitter{}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
maxsokolovsky, line 162:
```suggestion
                for _, tt := range tests {
		    tt := tt
		    t.Run(tt.name, func(t *testing.T) {
	                t.Parallel()
```

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -131,3 +136,60 @@ func Test_isValidName(t *testing.T) {
...
 	}
 }
+
+func TestValidateMachinePoolName(t *testing.T) {
+	tests := []struct {
+		name string
+		fail bool
+	}{
+		{
+			name: strings.Repeat("12345678", 8),
```
maxsokolovsky, line 146: Nit: I'd choose more descriptive names for subtests, and the value to test I'd make a separate field.

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
maxsokolovsky: This allows you to also run the subtests in parallel.

thatmidwesterncoder: Thanks for the code snippet - I wasn't quite sure where to put the parallel piece (above at the top of the test or here).

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
maxsokolovsky: You can run the test AND subtests in parallel in this case. The first line of the test can also be `t.Parallel()`.

thatmidwesterncoder: Ahhh fun! Updated.

maxsokolovsky (approved):
One thing to do later is to add documentation (in this case, a single sentence noting the validation added), but the docs infrastructure is on the dev branch, which is yet to merged into release.

PR #257, opened 2023-07-06
Remove cluster name local restriction

raulcabello:
## Issue: 
https://github.com/rancher/fleet/issues/1489
 
## Problem
Users can't migrate the local cluster to the `fleet-default` workspace. Cluster creation is rejected by the webhook. This is the error displayed:
```
[ERROR] error syncing ‘local’: handler provisioning-cluster-create: failed to create fleet-default/local [provisioning.cattle.io/v1](http://provisioning.cattle.io/v1), Kind=Cluster for provisioning-cluster-create local: admission webhook “[rancherauth.cattle.io](http://rancherauth.cattle.io/)” denied the requeset: cluster name must be 63 characters or fewer, cannot be “local” nor of the form “c-xxxxx”, requeuing
```
 
## Solution
Remove the local cluster name validation, so users are able to move the local cluster to a different workspace.

### DIFF
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
@@ -172,7 +172,7 @@ func (p *provisioningAdmitter) validateClusterName(request *admission.Request, r
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


MbolotSuse (changes requested):
Blocking for offline discussion.

raulcabello:
Moving the local cluster should not be supported. We will clearly state that moving the local cluster is not supported and hide the move button from the UI

PR #259, opened 2023-07-11
Add validation bypass to the webhook.

KevinJoiner:
Requests made by rancher-webhook's sudo service account will bypass the webhook's validation.


charts/rancher-webhook/templates/rbac.yaml
```diff
@@ -10,3 +10,16 @@ subjects:
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: rancher-webhook-sudo
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
```
MbolotSuse, line 22: I'm a bit nervous about adding another pre-built-in binding to cluster-admin - it makes this role an escalation path for an attacker. I think there's not much we can do here since we want this have permission on any types that the webhook validates (otherwise users will be stopped by RBAC even if the WH approves), but would you mind confirming that this is the best we can do here?

MbolotSuse (changes requested):
Couple of comments on the exact place that this happens for mutators/validators. I also think it would be a good idea to have the security team give this PR a look.

pkg/admission/admission.go
```diff
@@ -223,11 +235,18 @@ func NewMutatingHandlerFunc(handler MutatingAdmissionHandler) http.HandlerFunc {
...
+		bypass := bypassValidation(review.Request)
 		response, err := handler.Admit(webReq)
 		if response == nil {
 			response = &admissionv1.AdmissionResponse{}
 		}
-		logrus.Debugf("admit result: %s %s %s user=%s allowed=%v err=%v", webReq.Operation, webReq.Kind.String(), resourceString(webReq.Namespace, webReq.Name), webReq.UserInfo.Username, response.Allowed, err)
+		logrus.Debugf("admit result: %s %s %s user=%s allowed=%t err=%v bypass=%t", webReq.Operation, webReq.Kind.String(), resourceString(webReq.Namespace, webReq.Name), webReq.UserInfo.Username, response.Allowed, err, bypass)
+
+		// even if we have an error while running the mutator we will still send an allowed response when bypassing webhook
+		if bypass {
```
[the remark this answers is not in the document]
KevinJoiner: The original goal was to allow users to maintain behavior with and without the mutator, but based on your last example and offline conversation, we will move the bypass logic before attempting any operations.

charts/rancher-webhook/templates/rbac.yaml
MbolotSuse: Discussed offline - this is minimal as far as the permissions that we can assign this.

MbolotSuse (dismissed):

JonCrowther (dismissed):

pkg/admission/admission.go
```diff
@@ -347,3 +365,16 @@ func CreateWebhookName(handler WebhookHandler, suffix string) string {
 	}
 	return fmt.Sprintf("%s.%s.%s", webhookQualifier, subPath, suffix)
 }
+
+// bypassValidation users can bypass the webhook if they are the sudo account and system:masters group
+func bypassValidation(request *admissionv1.AdmissionRequest) bool {
```
pjbgf, line 370: The description mentions we are targeting kinds `RoleTemplates` and `GlobalRoles`, however the bypass does not enforce that. Shall we add a check for that effect?

KevinJoiner: Sorry, this bypass is meant to be used for all validation and mutation performed by the webhook. The first use case is for RoleTemplates and GloabalRoles. But as we migrate validation from Norman, we will also need to use this for other Resources. For example, Team2 just added some node driver validation that they need to bypass with their controller.

pjbgf (approved):
LGTM

macedogm (approved):

MbolotSuse (approved):

PR #260, opened 2023-07-12
Reject workspace creation if namespace already exists

raulcabello:
## Issue: 
https://github.com/rancher/fleet/issues/1519
 
## Problem
Rancher allows the creation of fleet workspaces even though the namespace already exists. Deleting that workspace causes the deletion of namespace along with its contents. Deleting unwanted fleet workspace results in data loss.

## Solution
Reject the workspace creation if namespace already exists

### DIFF
--- a/pkg/resources/management.cattle.io/v3/fleetworkspace/mutator.go
+++ b/pkg/resources/management.cattle.io/v3/fleetworkspace/mutator.go
@@ -1,6 +1,8 @@
 package fleetworkspace
 
 import (
+	"fmt"
+
 	"github.com/rancher/rancher/pkg/apis/management.cattle.io"
 	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/webhook/pkg/admission"
@@ -18,6 +20,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/kubernetes/pkg/registry/rbac/validation"
 )
+
+const k8sManagedLabel = "app.kubernetes.io/managed-by"
 
 var (
 	fleetAdminRole = "fleetworkspace-admin"
@@ -84,7 +88,8 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
 
 	namespace := v1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fw.Name,
+			Name:   fw.Name,
+			Labels: map[string]string{k8sManagedLabel: "rancher"},
 		},
 	}
 	ns, err := m.namespaces.Create(&namespace)
@@ -92,17 +97,20 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
 		if !errors.IsAlreadyExists(err) {
 			return nil, err
 		}
-		// check if user has enough privilege to create fleet admin rolebinding in the namespace
+		ns, err = m.namespaces.Get(fw.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to get fleetworkspace namespace '%s': %w", fw.Name, err)
+		}
+		if ns.Labels[k8sManagedLabel] != "rancher" {
+			return admission.ResponseBadRequest(fmt.Sprintf("namespace '%s' already exists", fw.Name)), nil
+		}
 		cr, err := m.clusterroles.Cache().Get(fleetAdminRole)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to get fleetAdmin ClusterRole: %w", err)
 		}
-		response := &admissionv1.AdmissionResponse{}
-		auth.SetEscalationResponse(response, auth.ConfirmNoEscalation(request, cr.Rules, namespace.Name, m.resolver))
-
-		ns, err = m.namespaces.Get(namespace.Name, metav1.GetOptions{})
+		err = auth.ConfirmNoEscalation(request, cr.Rules, namespace.Name, m.resolver)
 		if err != nil {
-			return nil, err
+			return admission.ResponseFailedEscalation(err.Error()), nil
 		}
 	}
 
--- a/pkg/resources/management.cattle.io/v3/fleetworkspace/mutator_test.go
+++ b/pkg/resources/management.cattle.io/v3/fleetworkspace/mutator_test.go
@@ -1,0 +1,265 @@
+package fleetworkspace
+
+import (
+	"encoding/json"
+	"net/http"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/rancher/rancher/pkg/apis/management.cattle.io"
+	v3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
+	"github.com/rancher/webhook/pkg/admission"
+	"github.com/rancher/wrangler/pkg/generic/fake"
+	"github.com/stretchr/testify/require"
+	admissionv1 "k8s.io/api/admission/v1"
+	authenticationv1 "k8s.io/api/authentication/v1"
+	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/kubernetes/pkg/registry/rbac/validation"
+)
+
+const (
+	nsName = "test"
+	user   = "test-user"
+)
+
+func TestAdmit(t *testing.T) {
+	workspace := v3.FleetWorkspace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: nsName,
+		},
+	}
+	raw, _ := json.Marshal(workspace)
+	req := &admission.Request{
+		AdmissionRequest: admissionv1.AdmissionRequest{
+			Object: runtime.RawExtension{Object: &workspace, Raw: raw},
+			UserInfo: authenticationv1.UserInfo{
+				Username: user,
+			},
+		},
+	}
+
+	tests := map[string]struct {
+		m                  func(t *testing.T) Mutator
+		req                *admission.Request
+		expectAllowed      bool
+		expectResultStatus *metav1.Status
+		expectedErr        error
+	}{
+		"reject because namespace already exists": {
+			m:             nsExistMutator,
+			req:           req,
+			expectAllowed: false,
+			expectResultStatus: &metav1.Status{
+				Status:  "Failure",
+				Message: "namespace 'test' already exists",
+				Reason:  metav1.StatusReasonBadRequest,
+				Code:    http.StatusBadRequest,
+			},
+		},
+		"new namespace without the 'app.kubernetes.io/managed-by: rancher' label": {
+			m:                  newNsMutator,
+			req:                req,
+			expectAllowed:      true,
+			expectResultStatus: nil,
+		},
+		"existing namespace with the 'app.kubernetes.io/managed-by: rancher' label": {
+			m:                  newNsWithLabelAndValidPermissionsMutator,
+			req:                req,
+			expectAllowed:      true,
+			expectResultStatus: nil,
+		},
+		"reject existing namespace with the 'app.kubernetes.io/managed-by: rancher' label with invalid permissions": {
+			m:             newNsWithLabelAndInvalidPermissionsMutator,
+			req:           req,
+			expectAllowed: false,
+			expectResultStatus: &metav1.Status{
+				Message: `user "test-user" (groups=[]) is attempting to grant RBAC permissions not currently held:
+{APIGroups:["management.cattle.io"], Resources:["fleetworkspaces"], ResourceNames:["test"], Verbs:["*"]}`,
+				Status: "Failure",
+				Reason: metav1.StatusReasonForbidden,
+				Code:   http.StatusForbidden,
+			},
+		},
+		"reject because namespace can't be fetched": {
+			m:           newNsErrorMutator,
+			req:         req,
+			expectedErr: errors.NewServerTimeout(schema.GroupResource{}, "", 2),
+		},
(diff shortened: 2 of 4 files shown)


weyfonk (approved):

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
```diff
@@ -89,21 +91,18 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
-		if err != nil {
-			return nil, err
-		}
-		response := &admissionv1.AdmissionResponse{}
-		auth.SetEscalationResponse(response, auth.ConfirmNoEscalation(request, cr.Rules, namespace.Name, m.resolver))
-
-		ns, err = m.namespaces.Get(namespace.Name, metav1.GetOptions{})
-		if err != nil {
-			return nil, err
+		if errors.IsAlreadyExists(err) {
```
KevinJoiner, line 94: This will make the mutator no longer [Idempotent](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#idempotence). kube-apiserver may call mutators multiple times for one request, which would cause this to fail the second time it is ran [more info on reinvocation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy)

One way to solve the original issue is to add labels on the created namespace that indicates it was created for fleet. Then you could deny the creation of a fleet workspace for any namespaces that do not have that label. And leave the current logic for a namespace that does have that label.

KevinJoiner (changes requested):
This will also need unit tests for these new changes. You can check out some of the unit tests for clusterroletemplatebindings or secrets to get started.

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
raulcabello: This has been fix in https://github.com/rancher/webhook/pull/260/commits/803504b098c5917b1bdff52c8ecc7f91148d7df8. I also added unit tests.

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
```diff
@@ -19,6 +21,8 @@ import (
 	"k8s.io/kubernetes/pkg/registry/rbac/validation"
 )
 
+const fleetManagedLabel = "fleet-workspace-managed-resource"
```
KevinJoiner, line 24: I would suggest using a standard label like `app.kubernetes.io/managed-by: rancher`
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
```diff
@@ -84,26 +88,31 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
+		if ns.Labels[fleetManagedLabel] != "true" {
+			return &admissionv1.AdmissionResponse{
+				Result: &metav1.Status{
+					Status:  "Failure",
+					Message: fmt.Sprintf("namespace '%s' already exists", fw.Name),
+					Reason:  metav1.StatusReasonInvalid,
+					Code:    http.StatusBadRequest,
+				},
+				Allowed: false,
+			}, nil
```
KevinJoiner, line 113: use the admission helper function for consistent webhook messages.
```suggestion
			return admission.ResponseBadRequest(fmt.Sprintf("namespace '%s' already exists", fw.Name)), nil
```

pkg/resources/management.cattle.io/v3/fleetworkspace/validator_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func TestAdmit(t *testing.T) {
+	nsName := "test"
+
+	nsExistMutator := func() Mutator {
+		ctrl := gomock.NewController(t)
+		mockNamespaceController := fakes.NewMockNamespaceController(ctrl)
```
KevinJoiner, line 24: We no longer need to generate fakes for Wrangler controllers. Instead, we can use the new [generic/fake](https://github.com/rancher/wrangler/tree/master/pkg/generic/fake) package from wrangler.
Hear is an [example](https://github.com/rancher/webhook/pull/262/files#diff-7d2acc95d10395b594e7b75b2516089d511d173afa33235d3a8d0d87de41ea44) of how that change looks.

This uses gomock as well, so you shouldn't need to change any of your test logic.

KevinJoiner (changes requested):
This approach looks like it should work. Before we merge, can you squash your commits into one single commit?

pkg/resources/management.cattle.io/v3/fleetworkspace/validator_test.go
raulcabello: done in https://github.com/rancher/webhook/pull/260/commits/24334874f8922d5187275b7fdf0f1c85dfcacf5f

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
raulcabello: done in https://github.com/rancher/webhook/pull/260/commits/24334874f8922d5187275b7fdf0f1c85dfcacf5f

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
raulcabello: done in https://github.com/rancher/webhook/pull/260/commits/24334874f8922d5187275b7fdf0f1c85dfcacf5f

raulcabello:
@KevinJoiner I made the changes suggested, and I squashed the commits into a single commit.

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
```diff
@@ -84,25 +87,21 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
 
 	namespace := v1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fw.Name,
+			Name:   fw.Name,
+			Labels: map[string]string{k8sManagedLabel: "true"},
```
KevinJoiner, line 91: We should use the value of `rancher` here so users who see these namespaces know that Rancher is responsible for managing their creation and deletion.
```suggestion
			Labels: map[string]string{k8sManagedLabel: "rancher"},
```

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
```diff
@@ -1,10 +1,11 @@
 package fleetworkspace
```
KevinJoiner, line 1: Can you rename this file and the test file to `mutator.go` and `mutator_test.go`? I think this was a mistake by someone else when the file was first created.

pkg/resources/management.cattle.io/v3/fleetworkspace/validator_test.go
```diff
@@ -0,0 +1,100 @@
...
+		mockClusterRoleController := fake.NewMockNonNamespacedControllerInterface[*rbacv1.ClusterRole, *rbacv1.ClusterRoleList](ctrl)
+		mockClusterRoleController.EXPECT().Create(gomock.Any())
+
+		return Mutator{
+			namespaces:          mockNamespaceController,
+			rolebindings:        mockRoleBindingController,
+			clusterrolebindings: mockClusterRoleBindingController,
+			clusterroles:        mockClusterRoleController,
+		}
+	}
```
KevinJoiner, line 53: These functions don't need to be local variables in the test function. They can be moved out into regular functions that take in `t *testing.T` as a parameter or `ctrl *gomock.Controler`

pkg/resources/management.cattle.io/v3/fleetworkspace/validator_test.go
```diff
@@ -0,0 +1,100 @@
...
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			m := test.m()
+			admit, _ := m.Admit(test.req)
+			if admit.Allowed != test.expectAllowed {
+				t.Errorf("Response was incorrectly validated, wanted response.Allowed = '%v' got %v: result=%+v", test.expectAllowed, admit.Allowed, admit.Result)
+			}
```
KevinJoiner, line 93: We use testify for our condition checking which gives us uniformed error messages between our unit test.
```suggestion
			require.Equalf(test.expectAllowed, admit.Allowed, "Response was incorrectly validated, wanted response.Allowed = '%v' got %v: result=%+v", test.expectAllowed, admit.Allowed, admit.Result)
```

KevinJoiner (changes requested):
We recently added a new requirement for documenting webhook functionality.
Could you also add a `FleetWorkspace.md` file that explains what this mutator is doing. 
Here is [how our docs work ](https://github.com/rancher/webhook#docs) and an [example Markdown file](https://github.com/rancher/webhook/blob/release/v0.4/pkg/resources/core/v1/secret/Secret.md) for our Secret mutator.

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
raulcabello: done in https://github.com/rancher/webhook/pull/260/commits/bc3715088473c65ec63c87a6da489240e56dbaef

pkg/resources/management.cattle.io/v3/fleetworkspace/validator.go
raulcabello: done in https://github.com/rancher/webhook/pull/260/commits/bc3715088473c65ec63c87a6da489240e56dbaef

pkg/resources/management.cattle.io/v3/fleetworkspace/validator_test.go
raulcabello: done in https://github.com/rancher/webhook/pull/260/commits/bc3715088473c65ec63c87a6da489240e56dbaef

pkg/resources/management.cattle.io/v3/fleetworkspace/validator_test.go
raulcabello: done in https://github.com/rancher/webhook/pull/260/commits/bc3715088473c65ec63c87a6da489240e56dbaef

raulcabello:
@KevinJoiner I added the docs and made all the changes. Let me know what you think. Thanks!

pkg/resources/management.cattle.io/v3/fleetworkspace/mutator.go
```diff
@@ -84,25 +88,29 @@ func (m *Mutator) Admit(request *admission.Request) (*admissionv1.AdmissionRespo
...
 	ns, err := m.namespaces.Create(&namespace)
 	if err != nil {
 		if !errors.IsAlreadyExists(err) {
 			return nil, err
 		}
-		// check if user has enough privilege to create fleet admin rolebinding in the namespace
-		cr, err := m.clusterroles.Cache().Get(fleetAdminRole)
+		ns, err := m.namespaces.Get(fw.Name, metav1.GetOptions{})
 		if err != nil {
 			return nil, err
```
KevinJoiner, line 102:
```suggestion
			return nil, fmt.Errorf("failed to get fleetworkspace namespace '%s': %w, fw.Name, err)
```

pkg/resources/management.cattle.io/v3/fleetworkspace/mutator_test.go
```diff
@@ -0,0 +1,96 @@
...
+		},
+	}
+
+	tests := map[string]struct {
+		m                  func(t *testing.T) Mutator
+		req                *admission.Request
+		expectAllowed      bool
+		expectResultStatus *metav1.Status
+	}{
+		"reject because namespace already exists": {
```
KevinJoiner, line 42: Let's add the following test cases to hit all of the changes made in this PR.

- namespace already exists with the expected label
- namespace already exists with the expected label, but the user fails the escalation check
- failed to create a namespace with an error other than alreadyExist.

To add tests that hit the escalation check, you will need to create a resolver for your mutator there is an example of this [here](https://github.com/rancher/webhook/blob/release/v0.4/pkg/resources/management.cattle.io/v3/roletemplate/validator_test.go#L34C1-L50)

KevinJoiner (changes requested):
Thanks for adding the docs. After you make changes, you can re-request a review in the top right corner, and I will get the notification.

pkg/resources/management.cattle.io/v3/fleetworkspace/mutator_test.go
raulcabello: Done. Please check the latest commit.

KevinJoiner (dismissed):
Thanks for making all the updates; this looks good to go.

pkg/resources/management.cattle.io/v3/fleetworkspace/mutator_test.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			m := test.m(t)
+			admit, err := m.Admit(test.req)
+			if err != nil {
+				require.Equalf(t, test.expectedErr, err, "Error doesn't match, wanted %v got %v", test.expectedErr, err)
+				return
+			}
```
weyfonk, line 102: Making the check on the returned error, instead of the expected one, may expose us to unnoticed failures if no error is returned although one is expected.
Conversely, running equality checks only if the expected error is not `nil` means that if no error is expected but one is returned, it may fly under our radar.

TL;DR:
```suggestion
			if test.expectedErr == nil {
				require.NoError(t, err)
		        } else {
				require.Equalf(t, err, "Error doesn't match, wanted %v got %v", test.expectedErr, err)
			}
```

weyfonk (changes requested):

pkg/resources/management.cattle.io/v3/fleetworkspace/mutator_test.go
raulcabello: changed here https://github.com/rancher/webhook/pull/260/commits/fc284bd621528d5128f03b4573f4c60b05afc6dc

Issue #211, opened 2023-03-29
[Feature] Add ability to customize webhook https and capi port


fauzanelka:
So I have an EKS cluster with Calico CNI installed.
When I installed rancher into the EKS cluster via helm chart, I ran into these problems.

1. I want to use hostNetwork for rancher-webhook but I don't know how to specify helm values for rancher-webhook. It seems rancher automatically install the chart. So I just edit it directly via the deployment yaml.
3. I just did briefly check the source code but it seems didn't have the ability to specify custom webhook https and capi ports. This is needed for hostNetwork because when ports 9443 and 8777 are already used by another deployment (e.g. aws-load-balancer-controller). The rancher webhook pod won't start (stuck at pending). It's good that I can change the aws-load-balancer-controller webhook port via webhookBindPort so the problem was solved.

bryanasdev000:
I need to test, but in theory, rancher chart expose a global var, and rancher-webhook can use it.

- https://github.com/rancher/charts/blob/dev-v2.7/charts/rancher-webhook/1.0.8%2Bup0.2.9/values.yaml#L6
- https://artifacthub.io/packages/helm/rancher-stable/rancher?modal=values

Will try it.

KevinJoiner:
We are looking at this now and I will update this issue when there is some more movement.

anupama2501:
@cpinjani to validate the EKS related changes

anupama2501:
Closing the issue as the validations run as part of the issue https://github.com/rancher/rancher/issues/41142#issuecomment-1667674189 look good.

PR #266, opened 2023-07-17
Adds the ability to set custom ports for Webhook.

KevinJoiner:
## Issue: https://github.com/rancher/rancher/issues/41142 & https://github.com/rancher/webhook/issues/211

First Commit: 
Adds the ability to customize the ports used by the webhook via helm values. 
Integration tests were updated to stop the running Rancher instance from downgrading the webhook during the tests.
To fix this, the test run rancher for the initial webhook installation. Then Rancher is stopped, and K3s is manually started.

### DIFF
--- a/pkg/capi/capi.go
+++ b/pkg/capi/capi.go
@@ -4,8 +4,10 @@ package capi
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	controllerruntime "github.com/rancher/lasso/controller-runtime"
 	"github.com/rancher/webhook/pkg/clients"
@@ -43,13 +45,23 @@ func init() {
 	_ = apiextensionsv1.AddToScheme(schemes.All)
 }
 
-var (
-	tlsCert  = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs", "tls.crt")
-	capiPort = 8777
+const (
+	defaultCapiPort = 8777
+	capiPortEnvKey  = "CATTLE_CAPI_PORT"
 )
+
+var tlsCert = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs", "tls.crt")
 
 // Register registers a new CAPI webhook server and returns a start function.
 func Register(clients *clients.Clients, tlsOpts ...func(*tls.Config)) (func(ctx context.Context) error, error) {
+	capiPort := defaultCapiPort
+	if portStr := os.Getenv(capiPortEnvKey); portStr != "" {
+		var err error
+		capiPort, err = strconv.Atoi(portStr)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode CAPI port value '%s': %w", portStr, err)
+		}
+	}
 	mgr, err := ctrl.NewManager(clients.RESTConfig, ctrl.Options{
 		MetricsBindAddress: "0",
 		NewCache: controllerruntime.NewNewCacheFunc(clients.SharedControllerFactory.SharedCacheFactory(),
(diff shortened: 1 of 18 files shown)


pkg/capi/capi.go
```diff
@@ -43,13 +45,20 @@ func init() {
...
-	capiPort = 8777
-)
+const defaultCapiPort = 8777
+
+var tlsCert = filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs", "tls.crt")
 
 // Register registers a new CAPI webhook server and returns a start function.
 func Register(clients *clients.Clients, tlsOpts ...func(*tls.Config)) (func(ctx context.Context) error, error) {
+	capiPort := defaultCapiPort
+	if portStr := os.Getenv("CATTLE_CAPI_PORT"); portStr != "" {
```
cmurphy, line 55: nit: make these env var names constants

scripts/integration-test
```diff
@@ -27,10 +28,37 @@ while ! kubectl rollout status -w -n cattle-system deploy/rancher-webhook >/dev/
...
+    pods=$(kubectl get pods -n cattle-system --field-selector=status.phase=Running -o jsonpath='{.items[?(@.metadata.generateName=="helm-operation-")].metadata.name}')
+    [[ -n "$pods" ]]
+do
+    echo "Waiting for help operation to finish"
+    sleep 2
+done
+
+# Kill Rancher since we only need the CRDs and the intial webhook values.
+# We do not want Rancher to reconcile an older version of the webhook during test.
+kill ${RANCHER_PID}
```
cmurphy, line 47: Could we instead change the `rancher-webhook-version` setting to `$TAG` to prevent it from reconciling a different value?

tests/integration/port_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if webhookPod == nil {
+		m.Require().FailNow("running webhook pod not found")
+	}
+	m.Require().Equal(corev1.PodRunning, webhookPod.Status.Phase, "Rancher-webhook pod is not running Phase=%s", webhookPod.Status.Phase)
+	m.Require().Len(webhookPod.Spec.Containers, 1, "Rancher-webhook pod has the incorrect number of containers")
+	m.Require().Len(webhookPod.Spec.Containers[0].Ports, 2, "Rancher-webhook container has the incorrect number of ports")
+	havePort1 := webhookPod.Spec.Containers[0].Ports[0].ContainerPort
+	havePort2 := webhookPod.Spec.Containers[0].Ports[1].ContainerPort
+	if havePort1 != testWebhookPort && havePort2 != testWebhookPort {
```
cmurphy, line 75: Checking the values on the pod object feels closer to a unit test than an integration test to me. Could this also test that the webhook is still reachable, with a `m.createObj(namespace)` for instance?

scripts/integration-test
```diff
@@ -40,14 +68,25 @@ WEBHOOK_REPO=$(ctr image import ${IMAGE_FILE} | cut -d ' ' -f 2 | cut -d ':' -f
...
+    fi
+    RETRIES=$((RETRIES + 1))
     sleep 2
 done
 
 ./bin/rancher-webhook-integration.test -test.v -test.run IntegrationTest
 
+# Install the webhook chart with new ports.
+helm upgrade rancher-webhook ./dist/artifacts/rancher-webhook-${HELM_VERSION}.tgz -n cattle-system \
+    --wait --reuse-values --set port=443 --set capi.port=2319
```
cmurphy, line 86: With the port being so low, should this also have `--set capNetBindService=true`?

scripts/integration-test
KevinJoiner: I will also add that I was able to run 60 consecutive iterations of the drone pipeline check for flakiness.

cmurphy (commented):
linter is failing, otherwise lgtm

JonCrowther (approved):
2 minor nits

cmurphy (approved):

PR #268, opened 2023-07-24
Document PRTB validation

maxsokolovsky:

### DIFF
--- a/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
+++ b/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
@@ -2,24 +2,34 @@
 
 ### Escalation Prevention
 
-Users can only create/update ProjectRoleTemplateBindings with rights less than or equal to those they currently possess. This is to prevent privilege escalation. 
+Users can only create/update ProjectRoleTemplateBindings with rights less than or equal to those they currently possess.
+This is to prevent privilege escalation.
 
 ### Invalid Fields - Create
 
-Users cannot create ProjectRoleTemplateBindings which violate the following constraints:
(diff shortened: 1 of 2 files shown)


pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
```diff
@@ -2,24 +2,34 @@
...
+    - Not locked (`roleTemplate.Locked` must be `false`)
+    - Associated with its appropriate context (`roleTemplate.Context` must be equal to "project")
 
 ### Invalid Fields - Update
 
 Users cannot update the following fields after creation:
+
 - RoleTemplateName
 - ProjectName
+- ServiceAccount
```
MbolotSuse, line 28: It's a bit odd to me that this one is in the "can't update" section rather than in the "can update if not set but not changeable after" section with the other subject fields. I think I understand why, but would you mind clarifying why so that I'm sure I've got the idea?

MbolotSuse (commented):
A couple of overall comments:
- In the future, I would recommend not making formatting changes at the same time as meaningful changes. In particular, if you are going to make formatting changes, consider splitting them out into a separate commit.
- There's no agreed on formatting guide for these documents. What this means is that it's unwise to make line length changes and the like unless it affects readability of the markdown as displayed by Github - your editor may like it more this way but that's not universal.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
maxsokolovsky: Discussed offline, will keep existing wording.

MbolotSuse (approved):

JonCrowther (approved):

PR #273, opened 2023-08-07
Add cert-based API server authentication

cmurphy:
Add the ability to authenticate incoming requests, verifying that all requests originate from the Kubernetes API server and no where else.

Authenticating the API server requires manual steps to configure both the API server and the webhook. Follow the Kubernetes webhook documentation[1] to create an admission configuration and kubeconfig for the API server, and update the kube-apiserver flags to use them. Only cert-based authentication is supported, basic auth and token authentication will not be recognized. Then, set auth.clientCA in the webhook chart's values.yaml to the base64-encoded CA for the certs, and set auth.allowedCNs to the CN for the client cert the apiserver will present.

[1] https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#authenticate-apiservers

## Problem

The webhook accepts and processes any request from any client. This does not expose any secrets, but could lead to a DDOS attack.
 
## Solution

Offer the ability to harden the webhook by having it require authentication from the kube-apiserver.
 
## Testing

Adds unit tests for helm chart values.

See linked issue for reproduction steps and documentation that can be used for manual testing.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -4,9 +4,14 @@ package server
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"os"
+	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -38,7 +43,10 @@ const (
 	defaultWebhookHTTPSPort = 9443
 	webhookPortEnvKey       = "CATTLE_PORT"
 	webhookURLEnvKey        = "CATTLE_WEBHOOK_URL"
+	allowedCNsEnv           = "ALLOWED_CNS"
 )
+
+var caFile = filepath.Join(os.TempDir(), "k8s-webhook-server", "client-ca", "ca.crt")
 
 // tlsOpt option function applied to all webhook servers.
 var tlsOpt = func(config *tls.Config) {
@@ -51,6 +59,7 @@ var tlsOpt = func(config *tls.Config) {
 		tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
 		tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
 	}
+	config.ClientAuth = tls.RequestClientCert
 }
 
 // ListenAndServe starts the webhook server.
@@ -118,6 +127,7 @@ func listenAndServe(ctx context.Context, clients *clients.Clients, validators []
 	router := mux.NewRouter()
 	errChecker := health.NewErrorChecker("Config Applied")
 	health.RegisterHealthCheckers(router, errChecker)
+	router.Use(certAuth())
 
 	logrus.Debug("Creating Webhook routes")
 	for _, webhook := range validators {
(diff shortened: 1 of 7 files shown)


KevinJoiner (dismissed):

pkg/server/server.go
```diff
@@ -38,8 +43,11 @@ const (
 	defaultWebhookHTTPSPort = 9443
 	webhookPortEnvKey       = "CATTLE_PORT"
 	webhookURLEnvKey        = "CATTLE_WEBHOOK_URL"
+	allowedCNsEnv           = "ALLOWED_CNS"
 )
 
+var caFile = filepath.Join(os.TempDir(), "k8s-webhook-server", "client-certs", "ca.crt")
```
pjbgf, line 49: Let's change this to store the cert outside of `/tmp`. Maybe `/var/lib/rancher/webhook/client-certs`?

@macedogm thoughts?

pjbgf (changes requested):
@cmurphy great job on this. I think we need to change the location for the ca file, apart from that it LGTM.

pkg/server/server.go
macedogm: Fully agree. `/tmp` is not ideal to store sensitive data, even if temporarily.

KevinJoiner: @pjbgf / @macedogm 
I will note that this is the default dir we are already using to store certs used for our CAPI webhook
https://github.com/kubernetes-sigs/controller-runtime/blob/7cfd07f52ecec21599a686ab7ae305b98fff35de/pkg/webhook/server.go#L82C2-L82C2

https://github.com/rancher/webhook/blob/release/v0.4/pkg/capi/capi.go#L53

If we change this, then we will have users storing their certs in two separate locations.

macedogm: Thanks for the explanation @KevinJoiner and @cmurphy.

@cmurphy it's the signing CA used to sign which certificates?

macedogm: But just to confirm, wouldn't https://github.com/rancher/webhook/pull/273/files#diff-547361dafecac27b6e2da396d00d4c8c673b1fabda2bb3ad3a5860c6706aed2cR90 be a sensitive data?

cmurphy: Maybe client-certs is a bad directory name, I was trying to distinguish it from the serving-certs. The only item in that directory should be the CA which signed the kube-apiserver client certs.

cmurphy: correct

macedogm: @cmurphy thanks for the explanation. Yes, we got confused by the dir name, because it implied something sensitive, which is not the case then.

cmurphy: Will change it, thanks for the feedback!

pjbgf: Thanks for all the context @cmurphy @KevinJoiner, I wasn't aware of the capi/upstream. Let's keep it as is for the time being.

pjbgf (approved):
LGTM

macedogm (dismissed):

KevinJoiner (dismissed):

cmurphy:
Changed target branch to release/0.4, please re-approve

macedogm (approved):

PR #277, opened 2023-08-10
Fix NodeDriver validation to ensure uniqueness + prevent race condition of CRD creation

thatmidwesterncoder:
## Problem

When two Node Drivers share the same name deleting/disabling them goes through the rancher-webhook, which will admit the request if no nodes are using them and subsequently delete the CRDs associated with one or the other, which may end up removing the wrong CRDs and put the cluster in a weird state. The only way to fix this would be to wait for the reconciler to run again and recreate them.

Another side effect here is that if the CRDs are not present (like it takes a bit for the items to download) the request is not admitted which is a race condition.
 
## Solution

The first part will be switching the check to go off of the `.spec.DisplayName` instead to `.objectMeta.Name`, that way the check will be globally unique.

The second part will just be fixiable by adding a check to see if the CRD exists before listing the resources, if the CRD doesn't exist then just admit the request. 
 
## Testing

First part:
1. Create a custom node driver (or rename 2 to have the same display name)
2. Delete one of them, notice the wrong CRDs may be deleted (one of them most likely, so might take trying it in a certain order)
3. Pre-patch, weird state will be observed. Post-patch it will delete only the right CRDs since we are keying off of the name.

Second part:
1. Open 2 browser tabs
2. In the UI, enable a NodeDriver that has an external dependency (takes more time)
3. `kubectl delete nodedriver <name>` where name is the one that was just enabled
4. Pre-patch, the delete will fail since it can't list the items (no CRDs). Post-patch it will admit the request just fine since the webhook will test and see if the CRD exists and just admit the request.

### DIFF
--- a/pkg/resources/management.cattle.io/v3/nodedriver/validator.go
+++ b/pkg/resources/management.cattle.io/v3/nodedriver/validator.go
@@ -9,8 +9,12 @@ import (
 	"github.com/rancher/webhook/pkg/admission"
 	controllersv3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
 	objectsv3 "github.com/rancher/webhook/pkg/generated/objects/management.cattle.io/v3"
+	"github.com/rancher/wrangler/pkg/generic"
+	"github.com/sirupsen/logrus"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


thatmidwesterncoder:
cc @rancher/rancher-team-2-hostbusters-dev @rancher/rancher-team-1-neo-dev 

sorry just pinging again here - would like to get this in for the q4 2.8 release.

jakefhyde (changes requested):

thatmidwesterncoder:
> Has this part been done yet? If so could you link the related PR:
> 
> > The first part will be switching the check to go off of the .spec.DisplayName instead to .objectMeta.Name, that way the check will be globally unique.

@jakefhyde yes - that is in the first commit. I could pull it into a separate PR if necessary.

jakefhyde:
> > Has this part been done yet? If so could you link the related PR:
> > > The first part will be switching the check to go off of the .spec.DisplayName instead to .objectMeta.Name, that way the check will be globally unique.
> 
> @jakefhyde yes - that is in the first commit. I could pull it into a separate PR if necessary.

Ah I mean the part to change node driver generation from the display name to the actual name. It will continue to work for builtins since they are always the same, but custom out of tree node drivers (which are still broken on v2prov) need to be updated here: https://github.com/rancher/rancher/blob/HEAD/pkg/controllers/management/drivers/nodedriver/machine_driver.go#L105-L110

jakefhyde (approved):

cmurphy (approved):
lgtm from a webhook functionality perspective. Trusting that @jakefhyde understands the team 2 implications of the node driver changes.

PR #285, opened 2023-09-01
[0.3] Disallow changes to ServiceAccount field on PRTB update rather than its presence

maxsokolovsky:
## Problem
The webhook rejects a request that updates any PRTB that already has ServiceAccount field **set**.

## Solution
Instead, it should reject attempts to **change** it. Adjust the logic accordingly.

## CheckList
  
- [x] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator_test.go
+++ b/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator_test.go
@@ -409,6 +409,59 @@ func (p *ProjectRoleTemplateBindingSuite) TestValidationOnUpdate() {
 			allowed: false,
 		},
 		{
+			name: "update removing a previously set service account",
+			args: args{
+				username: adminUser,
+				oldPRTB: func() *apisv3.ProjectRoleTemplateBinding {
+					basePRTB := newBasePRTB()
+					basePRTB.UserName = ""
+					basePRTB.ServiceAccount = "p1:default"
+					return basePRTB
+				},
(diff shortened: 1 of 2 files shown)


pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator_test.go
```diff
@@ -409,78 +409,96 @@ func (p *ProjectRoleTemplateBindingSuite) TestValidationOnUpdate() {
 			allowed: false,
 		},
 		{
-			name: "update previously unset user",
+			name: "update unrelated to previously set service account",
```
MbolotSuse, line 412: Nit: This test case effectively unsets the PRTB, which is a bit different than what this name would indicate.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator_test.go
MbolotSuse, line 1: Nit: I think we are missing the "service account wasn't set but is now set" test case.

MbolotSuse (approved):
A couple of non-critical nits.

KevinJoiner (approved):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator_test.go
maxsokolovsky: Added.

MbolotSuse (approved):

KevinJoiner (approved):

PR #288, opened 2023-09-06
Migrates GlobalRoles and GlobalRoleBinding.

KevinJoiner:
Migrates Noramn validation done on GlobalRoles and GlobalRoleBinding in Rancher to the webhook.

## Problem
Norman validation and mutation for GlobalRoles and GlobalRoleBindings need to be migrated to the webhook

#### Builtin Validation

The `globalroles.builtin` field is immutable, and new builtIn GlobalRoles cannot be created.
If `globalroles.builtin` is true then all fields are immutable except  `metadata` and `newUserDefault`.
If `globalroles.builtin` is true then the GlobalRole can not be deleted.

#### Invalid Fields - Update
Users cannot update the following fields after creation:
- `userName` 
- `groupPrincipalName`
- `globalRoleName`

#### Invalid Fields - Create
GlobalRoleBindings must have either `userName` or `groupPrincipalName`, but not both.

#### On create

When a GlobalRoleBinding is created an owner reference is attached to the backing GlobalRole defined by `globalRoleName`.

## CheckList
  
- [x] Test
  
- [x] Docs


pkg/resources/management.cattle.io/v3/globalrolebinding/GlobalRoleBinding.md
```diff
@@ -8,4 +8,21 @@ Users can only create/update GlobalRoleBindings with rights less than or equal t
...
+
+### Invalid Fields - Create
+GlobalRoleBindings must have either `userName` or `groupPrincipalName`, but not both.
+All RoleTemplates which are referred to in the `inheritedClusterRoles` field must exist and not be locked. 
+
+## Mutation Checks
+
+### On create
+
+When a GlobalRoleBinding is created an owner reference is attached to the backing GlobalRole defined by `globalRoleName`.
```
MbolotSuse, line 28: Right now it sounds like the owner ref is on the GlobalRole (not referring to the global role).

```suggestion
When a GlobalRoleBinding is created an owner reference is created on the binding referring to the backing GlobalRole defined by `globalRoleName`.
```

MbolotSuse (changes requested):
Still need to look at tests, but here's what I see on the prod-side code so far.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -75,23 +79,34 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+			// This admits update operations that happen to remove finalizers.
+			// This is needed to supported the deletion of old GlobalRoles that would not pass the update check that verifies all rules have verbs.
+			return admission.ResponseAllowed(), nil
+		}
+		if isMetaOnlyChange(oldGR, newGR) {
+			// if this change only affects metadata, don't validate any further
+			// this allows users with the appropriate permissions to manage labels/annotations/finalizers
+			return admission.ResponseAllowed(), nil
+		}
+		fieldErr := a.validateUpdateFields(oldGR, newGR, fldPath)
```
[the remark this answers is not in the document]
KevinJoiner: `isMetaDataOnly` is costly due to it's calls `DeepEqual` on metadata. `grContentEqual` alone is much cheaper. That being said, I did try out the change you suggested, but the code quality wasn't great since the second call ignores the `NewUserDefault` field.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
```diff
@@ -172,8 +182,16 @@ func (a *admitter) validateGlobalRole(globalRole *v3.GlobalRole, fieldPath *fiel
 	return nil
 }
 
+// canSkipValidation checks if validation matches one of two scenarios which allow us to skip validation.
+// 1. Object is scheduled for deletion (deletion timestamp set)
+// 2. Object is a metadata only change. (this allows users with the appropriate permissions to manage labels/annotations/finalizers)
+// deletionIsScheduled returns true if the GlobalRoleBinding is being deleted.
+func canSkipValidation(op admissionv1.Operation, oldGRB, newGRB *v3.GlobalRoleBinding) bool {
+	return op == admissionv1.Update && (newGRB.DeletionTimestamp != nil || isMetaOnlyChange(oldGRB, newGRB))
```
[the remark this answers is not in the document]
KevinJoiner: I agree that it is not necessary, but I liked the increased readability of having a separate function that explains the check.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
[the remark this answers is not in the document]
KevinJoiner: This was missed I updated the validation and tests.

pkg/resources/management.cattle.io/v3/globalrolebinding/main_test.go
```diff
@@ -0,0 +1,339 @@
...
+		req.Name = oldGRB.Name
+		req.Namespace = oldGRB.Namespace
+		req.OldObject.Raw, err = json.Marshal(oldGRB)
+		assert.NoError(t, err, "Failed to marshal GRB while creating request")
+	}
+	if newGRB != nil {
+		req.Name = newGRB.Name
+		req.Namespace = newGRB.Namespace
+		req.Object.Raw, err = json.Marshal(newGRB)
+		assert.NoError(t, err, "Failed to marshal GRB while creating request")
```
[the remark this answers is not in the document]
KevinJoiner: I don't think this can be combined since one is checking the function, and the other is checking the output of the function. If you can provide an example that doesn't call `test.args.newGRB()` twice I will update.

pkg/resources/management.cattle.io/v3/globalrolebinding/mutator.go
```diff
@@ -0,0 +1,93 @@
...
+	return gvr
+}
+
+// Operations returns list of operations handled by this mutator.
+func (m *Mutator) Operations() []admissionregistrationv1.OperationType {
+	return []admissionregistrationv1.OperationType{admissionregistrationv1.Create}
+}
+
+// MutatingWebhook returns the MutatingWebhook used for this CRD.
+func (m *Mutator) MutatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) []admissionregistrationv1.MutatingWebhook {
```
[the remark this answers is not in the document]
KevinJoiner: I will test this for any errors.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
```diff
@@ -172,8 +182,16 @@ func (a *admitter) validateGlobalRole(globalRole *v3.GlobalRole, fieldPath *fiel
 	return nil
 }
 
+// canSkipValidation checks if validation matches one of two scenarios which allow us to skip validation.
+// 1. Object is scheduled for deletion (deletion timestamp set)
```
tomleb, line 186: Why do we allow updates if the object is pending deletion?

pkg/resources/management.cattle.io/v3/globalrolebinding/main_test.go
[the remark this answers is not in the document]
JonCrowther: I hadn't considered the case where the function exists but it doesn't return a GRB. I think we can leave it as is then

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
KevinJoiner: So controllers can remove their finalizers since K8s will block deletes until the finalizers are removed. I will also note that this is safe since you can not remove the deletion timestamp once it has been set.

tomleb: >  I will also note that this is safe since you can not remove the deletion timestamp once it has been set.

Yeah, I was wondering if there was a possibility that somehow (finalizers perhaps) a GRB stays alive (although still in pending delete state) and an update could still cause some security issues. eg: GRB binds to `regular-user`, then is set to be deleted but stays alive for a while, and is updated to bind to `superuser`. Does that make any sense?

MbolotSuse (changes requested):
Couple of small comments on tests.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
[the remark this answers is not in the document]
MbolotSuse: > isMetaDataOnly is costly due to it's calls DeepEqual on metadata. 

I'm not quite sure that I agree here - even though reflect isn't the quickest package, I would expect that the metadata (which on average is a smaller object) will be much quicker to compare than the rules fields (particularly at scale with large rules).

>  That being said, I did try out the change you suggested, but the code quality wasn't great since the second call ignores the NewUserDefault field.

Ok, I'll resolve this since IIRC even deep equals on everything takes quite a few rules before it starts to scale sufficiently badly (and I didn't have a super clean solution in mind).

JonCrowther (dismissed):

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
KevinJoiner: Yeah, that is a valid concern our controller, which assign permissions based on GRB, check if the deletion timestamp is set before operating on the object.

pkg/resources/management.cattle.io/v3/globalrolebinding/mutator_test.go
```diff
@@ -0,0 +1,123 @@
...
+	ctrl := gomock.NewController(t)
+	globalRoleCache := fake.NewMockNonNamespacedCacheInterface[*apisv3.GlobalRole](ctrl)
+	globalRoleCache.EXPECT().Get(adminGR.Name).Return(adminGR, nil).AnyTimes()
+	globalRoleCache.EXPECT().Get(notFoundName).Return(nil, newNotFound(notFoundName)).AnyTimes()
+	globalRoleCache.EXPECT().Get(errName).Return(nil, errServer).AnyTimes()
+
+	validator := globalrolebinding.NewMutator(globalRoleCache)
+
+	tests := []testCase{
+		{
```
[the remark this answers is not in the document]
KevinJoiner: Good suggestion I found an edge case that was missed. I am unable to test a failed patch since an invalid request body get's caught during decode.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator_test.go
```diff
@@ -1,583 +1,591 @@
...
-		rtCacheMock  *fake.MockNonNamespacedCacheInterface[*v3.RoleTemplate]
-		grCacheMock  *fake.MockNonNamespacedCacheInterface[*v3.GlobalRole]
-		grbCacheMock *fake.MockNonNamespacedCacheInterface[*v3.GlobalRoleBinding]
-	}
-
-	type testCase struct {
-		name                 string
-		globalRoleBinding    *v3.GlobalRoleBinding
-		oldGlobalRoleBinding *v3.GlobalRoleBinding
-		operation            admissionv1.Operation
```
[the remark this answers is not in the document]
KevinJoiner: Yeah, I agree these changes stemmed from the previous PR, which was written before any tests for GR and GRBs.

KevinJoiner:
1. Updated the Mutator to be idempotent. Check if the ref exist before adding it again.
2. added a reference to the struct changes in go.mod that should be removed when https://github.com/rancher/rancher/pull/42667 is merged.
3. Reverted dd7b7b1897a51dde8374554fa3ec332ea7566264

pkg/resources/management.cattle.io/v3/globalrolebinding/mutator.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	ownerReference := v1.OwnerReference{
+		APIVersion: globalRole.APIVersion,
+		Kind:       globalRole.Kind,
+		Name:       globalRole.Name,
+		UID:        globalRole.UID,
+	}
+	newRefStr := ownerReference.String()
+	for i := range newGRB.OwnerReferences {
+		if newGRB.OwnerReferences[i].String() == newRefStr {
```
MbolotSuse, line 93: I'm a bit uncomfortable on relying on `.String()` producing a value suitable for equality checks. Put differently, how confident are we that `a.String() == b.String()` iff `a == b` here? If you have some examples from other projects using this that would probably help.

MbolotSuse (changes requested):
Mostly LGTM - leaving this comment more as a block until the fork is replaced.

One minor concern on the owner ref.

pkg/resources/management.cattle.io/v3/globalrolebinding/mutator.go
KevinJoiner: Not tied to the usage of `String()`, I can also just check all the fields. I am confident  a.String() == b.String() iff a == b since the function is generated based on all of the fields [here](https://github.com/kubernetes/apimachinery/blob/v0.28.1/pkg/apis/meta/v1/generated.pb.go#L4610-L4624)

MbolotSuse: I'd go with the value comparison, that seems more solid to me than the string comparison.

tomleb: In case it can be useful, controller-runtimes implements a `SetOwnerReference()` function: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controllerutil/controllerutil.go#L97. If you follow the code you see that they compare the group, the kind and the name. Not sure if we can use that function directly, maybe you know better here.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -124,6 +139,30 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+// validateDelete checks if a global role can be deleted and returns the appropriate response.
+func validateDelete(oldRole *v3.GlobalRole, fldPath *field.Path) (*admissionv1.AdmissionResponse, error) {
+	if oldRole.Builtin {
+		return admission.ResponseBadRequest(field.Forbidden(fldPath, "cannot delete builtin GlobalRoles").Error()), nil
+	}
+	return admission.ResponseAllowed(), nil
+}
+
+// validateCreateFields blocks the creation of builtin globalRoles
+func validateCreateFields(oldRole *v3.GlobalRole, fldPath *field.Path) *field.Error {
```
tomleb, line 151: When are built-in roles created and by who? How is it bypassing this check?

MbolotSuse: Rancher creates the built-in GlobalRoles (as well as built-in role templates): https://github.com/rancher/rancher/blob/49e4618602de246cd51188d9338fb850ffac975e/pkg/data/management/rolebuilder.go#L318. To bypass the webhook, it uses the [bypass method](https://github.com/rancher/rancher-docs/blob/08ec8f2bddb89617c8f7cd0277f7649f8bbfb61a/docs/reference-guides/rancher-webhook.md#bypassing-the-webhook) (docs are being worked on), as you can see here: https://github.com/rancher/rancher/blob/49e4618602de246cd51188d9338fb850ffac975e/pkg/data/management/rolebuilder.go#L337.

tomleb: Good to know

pkg/resources/management.cattle.io/v3/globalrolebinding/mutator.go
```diff
@@ -0,0 +1,104 @@
...
+		return fmt.Errorf("failed to get referenced globalRole: %w", err)
+	}
+	ownerReference := v1.OwnerReference{
+		APIVersion: globalRole.APIVersion,
+		Kind:       globalRole.Kind,
+		Name:       globalRole.Name,
+		UID:        globalRole.UID,
+	}
+	for i := range newGRB.OwnerReferences {
+		if newGRB.OwnerReferences[i].APIVersion == ownerReference.APIVersion &&
```
MbolotSuse, line 92: Personally speaking, I prefer @tomleb 's suggestion here: https://github.com/rancher/webhook/pull/288#discussion_r1319071063 about the use of an upstream library, but since they are effectively very similar I won't block the PR based on that.

MbolotSuse (approved):
One nit, lgtm outside of that.

pkg/resources/management.cattle.io/v3/globalrolebinding/mutator.go
KevinJoiner: The use cases are a little different here. We are trying to keep our mutator idempotent, and this call only runs on create. The controller-runtime function is designed to run in a controller to reconcile the state of an object on change. I also don't like the behavior of overwriting the user's OwnerRef if they match GVK. Other nits are that the function is a lot heavier and more complex than our basic needs, and requires us to make a `runtime.Scheme`

MbolotSuse: > and requires us to make a runtime.Scheme

Yeah that part isn't great, though IIRC you should be able to import a schemeBuilder from v3 and use that to get a workable scheme somewhat easily.

> I also don't like the behavior of overwriting the user's OwnerRef if they match GVK

I'm not sure that I agree with your assessment here, from what I can see it needs to match Group, Version, Kind, and Name - https://github.com/kubernetes-sigs/controller-runtime/blob/2a553d6f910dfc4b0796142ba1e059639a21c54a/pkg/controller/controllerutil/controllerutil.go#L170, meaning that it will only match if it's for the exact GR by name. UID should be the same for resources of the same GVK + Name, and the block owner deletion seems more like a user-set option and less like an exact equality match.

That being said, I get what you mean here, and like I said, not a blocking comment.

MbolotSuse: Discussed offline, in order to not modify the user's input we are going to stay with the current implementation which never overrides the user's input.

KevinJoiner:
Rebased with with latest go.mod changes

MbolotSuse (approved):

JonCrowther (approved):

PR #289, opened 2023-09-07
Api version bump

MbolotSuse:
## Issue: 
CI in rancher which uses the new webhook is currently failing. This updates the Rancher/pkg/apis import to use new types to resolve the CI failures, and adds a temporary ci change to allow CI in the webhook to pass.

### DIFF
--- a/scripts/integration-test
+++ b/scripts/integration-test
@@ -5,6 +5,9 @@ export CATTLE_DEV_MODE=yes
 export CATTLE_SERVER_URL="https://$(ip route get 8.8.8.8 | awk '{print $7}'):443"
 export CATTLE_BOOTSTRAP_PASSWORD="admin"
 export CATTLE_FEATURES="harvester=false"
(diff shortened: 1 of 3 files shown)


scripts/integration-test
```diff
@@ -5,6 +5,9 @@ export CATTLE_DEV_MODE=yes
 export CATTLE_SERVER_URL="https://$(ip route get 8.8.8.8 | awk '{print $7}'):443"
 export CATTLE_BOOTSTRAP_PASSWORD="admin"
 export CATTLE_FEATURES="harvester=false"
+export CATTLE_CHART_DEFAULT_BRANCH="dev-v2.8"
```
tomleb, line 8: Should we document those as "temporary"?

tomleb (changes requested):

scripts/integration-test
KevinJoiner: This is should be fine since we will remove it in this week.

tomleb (approved):

KevinJoiner (approved):

PR #292, opened 2023-09-13
Don't allow PRTB namespace and the project ID of projectName to differ

maxsokolovsky:
## Problem
Users aren't forbidden to create a badly configured PRTB where the namespace and the project ID part of `projectName` don't match.

For example:
```yaml
kind: ProjectRoleTemplateBinding
metadata:
  name: test
  namespace: p-cgtq4
projectName: c-some-cluster:foobar
roleTemplateName: configmaps-manage
userName: u-tgws8
```

A valid PRTB example would have a namespace `p-cgtq4` and projectName `c-some-cluster:p-cgtq4`.

If users are allowed to specify the project ID and namespace different, there are two problems:
1. A Rancher controller will continuously fail to sync the RBAC objects to the cluster
2. If a namespace exists but doesn't equal the project ID, then the PRTB will never be cleaned up when the actual project (for which the binding was intended) gets deleted

## Solution
Extract the cluster name and project ID from `projectName` (split on colon). Ensure the project ID equals the namespace.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -182,10 +182,13 @@ This is to prevent privilege escalation.
 
 Users cannot create ProjectRoleTemplateBindings that violate the following constraints:
 
+- The `ProjectName` field must be:
+    - Provided as a non-empty value
+    - Specified using the format of `cluster-id:project-id`
+    - `project-id`, in particular, must match the namespace of the ProjectRoleTemplateBinding
 - Either a user subject (through `UserName` or `UserPrincipalName`), or a group subject (through `GroupName`
   or `GroupPrincipalName`), or a service account subject (through `ServiceAccount`) must be specified. Exactly one
   subject type of the three must be provided.
-- `ProjectName` must be specified
 - The roleTemplate indicated in `RoleTemplateName` must be:
     - Provided as a non-empty value
     - Valid (there must exist a `roleTemplate` object of given name in the `management.cattle.io/v3` API group)
--- a/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
+++ b/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
@@ -9,10 +9,13 @@ This is to prevent privilege escalation.
 
 Users cannot create ProjectRoleTemplateBindings that violate the following constraints:
 
+- The `ProjectName` field must be:
+    - Provided as a non-empty value
+    - Specified using the format of `cluster-id:project-id`
+    - `project-id`, in particular, must match the namespace of the ProjectRoleTemplateBinding
 - Either a user subject (through `UserName` or `UserPrincipalName`), or a group subject (through `GroupName`
   or `GroupPrincipalName`), or a service account subject (through `ServiceAccount`) must be specified. Exactly one
   subject type of the three must be provided.
-- `ProjectName` must be specified
 - The roleTemplate indicated in `RoleTemplateName` must be:
     - Provided as a non-empty value
     - Valid (there must exist a `roleTemplate` object of given name in the `management.cattle.io/v3` API group)
--- a/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
+++ b/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
@@ -104,8 +104,6 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		}
 	}
 
-	clusterNS, projectNS := clusterFromProject(prtb.ProjectName)
-
 	roleTemplate, err := a.roleTemplateResolver.RoleTemplateCache().Get(prtb.RoleTemplateName)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
@@ -121,6 +119,7 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		return nil, fmt.Errorf("failed to get rules from referenced roleTemplate '%s': %w", roleTemplate.Name, err)
 	}
 
+	clusterNS, projectNS := clusterAndProjectID(prtb.ProjectName)
 	err = auth.ConfirmNoEscalation(request, rules, clusterNS, a.clusterResolver)
 	if err == nil {
 		return &admissionv1.AdmissionResponse{Allowed: true}, nil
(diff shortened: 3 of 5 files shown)


pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
```diff
@@ -94,6 +94,11 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		return nil, fmt.Errorf("failed to decode PRTB object from request: %w", err)
 	}
 
+	clusterNS, projectNS := clusterFromProject(prtb.ProjectName)
+	if projectNS != prtb.Namespace {
```
MbolotSuse, line 98: I think this might be at the wrong spot in the validation process. I think in the current format, if you try to change the `projectName` on an active PRTB, chances are that you will hit this validation (saying that the `projectName` doesn't match the namespace), rather than the "right" validation (that you can't change projectName after creation). 

I'd recommend moving this to the `validateCreateFields` function rather than having it in the main flow here.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
```diff
@@ -9,6 +9,8 @@ This is to prevent privilege escalation.
 
 Users cannot create ProjectRoleTemplateBindings that violate the following constraints:
 
+- The namespace and project ID (the second part of the projectName field, as in `c-some-cluster:p-cgtq4`) must be
```
MbolotSuse, line 12:
```suggestion
- The projectName field must have the format `cluster-id:project-id`, and `project-id` must match the namespace of the ProjectRoleTemplateBinding.
```

MbolotSuse (changes requested):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go
maxsokolovsky: Good point. I mostly wanted to avoid calling `clusterFromProject` twice, but it's cheap, so that's fine. Will adjust.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
```diff
@@ -9,10 +9,12 @@ This is to prevent privilege escalation.
 
 Users cannot create ProjectRoleTemplateBindings that violate the following constraints:
 
+- The projectName field must have the format `cluster-id:project-id`, and `project-id` must match the namespace of the
```
cmurphy, line 12: This is talking about the same `ProjectName` as mentioned on line 17 right? Maybe state both requirements ("must be specified" and "have the format...") on the same line.

docs.md
```diff
@@ -182,10 +182,12 @@ This is to prevent privilege escalation.
 
 Users cannot create ProjectRoleTemplateBindings that violate the following constraints:
 
+- The projectName field must have the format `cluster-id:project-id`, and `project-id` must match the namespace of the
```
tomleb, line 185: Might be too late but why aren't we using something like:

```yaml
projectRef:
  projectId: "project-id"
  clusterId: "cluster-id"
```

which is a similar pattern in other objects like:
- `ClusterRoleBinding.roleRef`
- secretRef / configMapRef in Pods

Seems like it would be easier to forget that projectName is really two fields.

MbolotSuse: That might make more sense overall, but is subject to prior discussions on migrating resource fields, so I would not recommend attempting to address it here.

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
maxsokolovsky: Adjusted!

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
```diff
@@ -9,10 +9,12 @@ This is to prevent privilege escalation.
 
 Users cannot create ProjectRoleTemplateBindings that violate the following constraints:
 
+- The `ProjectName` field must be specified and have the format of `cluster-id:project-id`, and `project-id` must match
+  the namespace of the ProjectRoleTemplateBinding.
 - Either a user subject (through `UserName` or `UserPrincipalName`), or a group subject (through `GroupName`
   or `GroupPrincipalName`), or a service account subject (through `ServiceAccount`) must be specified. Exactly one
   subject type of the three must be provided.
-- `ProjectName` must be specified
+- `ProjectName` must be specified.
```
cmurphy, line 17: This is now duplicating the information on line 12

cmurphy (changes requested):

cmurphy (approved):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
```diff
@@ -9,10 +9,11 @@ This is to prevent privilege escalation.
 
 Users cannot create ProjectRoleTemplateBindings that violate the following constraints:
 
+- The `ProjectName` field must be specified and have the format of `cluster-id:project-id`, and `project-id` must match
+  the namespace of the ProjectRoleTemplateBinding.
```
MbolotSuse, line 13: Nit: Consider fixing the grammar so that you only use one and, like:
"The `ProjectName` field must be specified, have the format of `cluster-id:project-id`, and `project-id` must match
  the namespace of the ProjectRoleTemplateBinding."

However, since there are now 3 requirements, I would go with something like what we did for RoleTemplate:
```suggestion
- The `ProjectName` field must be :
  - Provided as a non-empty value
  - Specified using the format of `cluster-id:project-id`
    - `project-id`, in particular, must match the namespace of the ProjectRoleTemplateBinding
```

MbolotSuse (approved):
One grammar nit, but the functionality seems to work well.

cmurphy (approved):

pkg/resources/management.cattle.io/v3/projectroletemplatebinding/ProjectRoleTemplateBinding.md
maxsokolovsky: Adjusted.

MbolotSuse (approved):

PR #293, opened 2023-09-14
Disallows the updating GlobalRoles to be builtin.

KevinJoiner:
## Problem

Users could update a globalRole that was not a builtIn to be a builtin role.

## Solution

When a previously non-builtIn role is updated we check if it is being changed to a builtIn role.

## CheckList
  
- [x] Test
  
- [ ] Docs __No docs change needed__


MbolotSuse (approved):

maxsokolovsky (changes requested):
Currently, the docs say:
```If `globalroles.builtin` is true then all fields are immutable except  `metadata` and `newUserDefault`.```

I think the docs might need a small addition. When you update a non-builtIn global role, the only thing you can update is metadata.

MbolotSuse:
@maxsokolovsky 

> I think the docs might need a small addition. When you update a non-builtIn global role, the only thing you can update is metadata.

I don't think that this is true - when updating a non-builtIn global role you can change rules, newUserDefault, inheritedClusterRoles - in short, a lot of other fields. Can you clarify what is making you think this?

maxsokolovsky:
My bad, I read some code paths wrong, no issues found.

maxsokolovsky (approved):

PR #294, opened 2023-09-15
Prevent CRTBs from being created with mismatching Namespace and ClusterName

JonCrowther:
## Problem
Users can create CRTBs that have different namespaces and ClusterNames. That causes their deletion to fail when the cluster is deleted. It also doesn't appear in the UI as intended.

## Solution
- Added a check on CREATE to ensure that the namespace and custerName are the same. Deny the request if they are mismatched
- Added the same check on UPDATE, but just provide a warning to the user and allow it.

I also fixed the wording of one of the error messages to be more clear.

## CheckList
- [X] Test
- [X] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/validator.go
+++ b/pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/validator.go
@@ -168,7 +168,7 @@ func (a *admitter) validateCreateFields(newCRTB *apisv3.ClusterRoleTemplateBindi
 	hasGroupTarget := newCRTB.GroupName != "" || newCRTB.GroupPrincipalName != ""
 
 	if (hasUserTarget && hasGroupTarget) || (!hasUserTarget && !hasGroupTarget) {
-		return field.Forbidden(fieldPath, "binding must target either a user [userId]/[userPrincipalId] OR a group [groupId]/[groupPrincipalId]")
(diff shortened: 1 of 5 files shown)


pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/validator.go
```diff
@@ -91,6 +92,10 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		if err := validateUpdateFields(oldCRTB, newCRTB, fieldPath); err != nil {
 			return admission.ResponseBadRequest(err.Error()), nil
 		}
+
+		if newCRTB.ClusterName != newCRTB.Namespace {
+			warnings = append(warnings, "Using CRTBs with namespaces and clusterNames that differ is not supported")
```
cmurphy, line 97: I don't think we need this, because ClusterName and Namespace are both immutable, so this issue would be caught in `validateCreateFields`. If I'm mistaken then we should add the same validation for PRTBs to be consistent.

cmurphy (approved):

MbolotSuse (approved):
Please squash your commits when merging, but outside of that LGTM.

maxsokolovsky (dismissed):

MbolotSuse (approved):

maxsokolovsky (approved):

cmurphy (approved):

PR #295, opened 2023-09-19
Adds startup error.

KevinJoiner:
## Problem
To help users debug startup troubles with the webhook, we should add a log statement when the webhook fails to start instead of forcing users to use the /healthz endpoint only. 

## Solution
If we fail to start log the error.

## CheckList
  
- [x] Test
     __Changes could affect webhook startup which is covered by current integration tests__
  
- [ ] Docs 
   __No docs changes are needed__


MbolotSuse (approved):

pkg/server/server.go
```diff
@@ -250,6 +250,9 @@ func (s *secretHandler) sync(_ string, secret *corev1.Secret) (*corev1.Secret, e
 		Webhooks: mutatingWebhooks,
 	}
 	err := s.ensureWebhookConfiguration(validatingConfig, mutatingConfig)
+	if err != nil {
+		logrus.Errorf(err.Error())
```
tomleb, line 254: Should we use `logrus.Error` instead of `logrus.Errorf` since we don't have any format string?

tomleb (changes requested):

pkg/server/server.go
KevinJoiner: sure

tomleb (approved):

PR #296, opened 2023-09-21
Add validation for NodeDriver create operation

thatmidwesterncoder:
## Problem

We generate CRDs from the names of the NodeDriver - before this we did not have any validation on those. 
## Solution

This PR adds just some simple sanitation to the name that can be extended in the future if necessary.

## CheckList
  
- [x] Test
  
- [x] Docs


pkg/resources/management.cattle.io/v3/nodedriver/NodeDriver.md
```diff
@@ -1,7 +1,11 @@
...
-Note: checks only run if a node driver is being disabled or deleted
+Note: validation runs for node driver create, update, and delete operations
 
 ### Machine Deletion Prevention
 
 This admission webhook prevents the disabling or deletion of a NodeDriver if there are any Nodes that are under management by said driver. If there are _any_ nodes that use the driver the request will be denied.
+
+### Name Validation
+
+This admission webhook sanitizes the names of the NodeDriver since that name will be used in generating a Dynamic Schema.
```
KevinJoiner, line 11: sanitizes usually infers that we are taking the name and mutating it so that it is a valid name. Instead, we are just validating that the name matches a specific format.
```suggestion
### Display Name Validation

This admission webhook validates that the names of the NodeDriver does not contain numbers or hyphens 
```

pkg/resources/management.cattle.io/v3/nodedriver/validator.go
```diff
@@ -90,29 +102,63 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
-	// toggles from true -> false
-	if !(request.Operation == admissionv1.Delete && oldObject.Spec.Active) &&
-		!(request.Operation == admissionv1.Update && !newObject.Spec.Active && oldObject.Spec.Active) {
-		return admission.ResponseAllowed(), nil
+	switch request.Operation {
+	case admissionv1.Create:
+		// if the request is a create operation - sanitize the name a bit to make
+		// sure we can create dynamicSchemas cleanly from it
+		if !goodNameForCRD(newObject.Name) {
+			return invalidName, nil
```
KevinJoiner, line 110: use a field error to make our error messages match upstreams formats
``` suggestion
            fieldError := field.Invalid(field.NewPath("nodedriver").Child("metadata", "name"), newObject.Name, "name can not contain numbers or hyphens")
			return admission.BadRequest(fieldError.Error()), nil
```

KevinJoiner (changes requested):
Do we need this validation? In your rancher PR we are updating the dynamic Schema `v1.ObjectMeta.Name` to be generated based on the NodeDriver name which is also a `v1.ObjectMeta.Name` both have the same character requirements as defined in k8s https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

thatmidwesterncoder:
Closing this since it isn't necessary - @KevinJoiner you're right the CRDs do get created just fine - it just took some funky hacking on my end to prove it out. Basically on the rancher side I needed to make it use the name of the node driver for the rancher-machine plugin, but then use the display name for creating CRDs.

Ends up with a good config like this:
```
digitalocean-thingconfigs.rke-machine-config.cattle.io            2023-09-25T18:38:18Z
digitalocean-thingmachines.rke-machine.cattle.io                  2023-09-25T18:38:18Z
digitalocean-thingmachinetemplates.rke-machine.cattle.io          2023-09-25T18:38:18Z
```

which works even if the user puts dashes/numbers/etc. So we're good to go without validation as long as we use the name and not the display name which is the overarching goal here.

PR #298, opened 2023-09-25
Adding SA exceptions to gr escalation

MbolotSuse:
## Overview

Adds an exception for the fleet/backup-restore service accounts for the `inheritedClusterRoles` escalation check

## Issue Detail
The escalation check for the `inheritedClusterRoles` field currently only allows users to grant permissions through this field (either by modifying the globalRole or any associated binding) if one of two items is true:
- The user possess rules that they are trying to give, specifically through the `inheritedClusterRoles` field. 
  - This means that a user can only give out (for example) `clusterOwner` if they have been assigned `clusterOwner` in the `inheritedClusterRoles` field of a globalRole.
- The user is an admin/restricted admin
  - The code treats the global roles (with these names) as having `clusterOwner` in the `inheritedClusterRoles` field.

This behavior caused issues with service accounts (namely fleet and backup-restore) which were unable to pass these escalation checks, since GRBs can't be made to service accounts (meaning they never pass one of the two above conditions). This resulted in fleet/backup-restore not being able to create GlobalRoles (or bindings to those roles) which use this field, or are the `admin` or `restricted-admin` roles. 

## Fix

The current approach is to allow usage of this field for service accounts that meet all of the following requirements:
- First, they need to be service accounts (in the `system:serviceaccounts` group)
- Second, they need to be one of the two hardcoded service accounts that we allow this exception for (fleet/backup-restore)
  - This part may need to change, it hasn't been confirmed yet that these two service accounts are always in the respective namespaces defined in this PR
- Third, they need to pass a SAR check for `*` verbs on `*` resources in `*` api groups at the cluster scope.
  - This is to prevent users from bypassing this check in the case of rancher setups which aren't using fleet/backup-restore
  - Because of this check, such a user would still need to make the SA an admin through other means, which would present a security risk on it's own (even absent this check)
  - Credit to @KevinJoiner for suggesting the inclusion of this check alongside the others.  

Tests have also been added for the various cases defined in this PR.

### DIFF
--- a/pkg/resolvers/grbClusterResolver.go
+++ b/pkg/resolvers/grbClusterResolver.go
@@ -1,13 +1,19 @@
 package resolvers
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	apisv3 "github.com/rancher/rancher/pkg/apis/management.cattle.io/v3"
 	"github.com/rancher/webhook/pkg/auth"
 	v3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
+	v1 "k8s.io/api/authorization/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/serviceaccount"
 	"k8s.io/apiserver/pkg/authentication/user"
+	authorizationv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 )
 
 const (
@@ -15,20 +21,37 @@ const (
 	localCluster    = "local"
 )
 
+var (
+	exceptionServiceAccounts = []string{"rancher-backup", "fleet-agent"}
+	adminRules               = []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{rbacv1.APIGroupAll},
+			Resources: []string{rbacv1.ResourceAll},
+			Verbs:     []string{rbacv1.VerbAll},
+		},
+		{
+			NonResourceURLs: []string{rbacv1.NonResourceAll},
+			Verbs:           []string{rbacv1.VerbAll},
+		},
+	}
+)
+
 // GRBClusterRuleResolver implements the rbacv1.AuthorizationRuleResolver interface. Provides rule resolution
 // for the permissions a GRB gives that apply in a given cluster (or all clusters).
 type GRBClusterRuleResolver struct {
 	GlobalRoleBindings v3.GlobalRoleBindingCache
 	GlobalRoleResolver *auth.GlobalRoleResolver
+	sar                authorizationv1.SubjectAccessReviewInterface
 }
 
 // New NewGRBClusterRuleResolver returns a new resolver for resolving rules given through GlobalRoleBindings
 // which apply to cluster(s). This function can only be called once for each unique instance of grbCache.
-func NewGRBClusterRuleResolver(grbCache v3.GlobalRoleBindingCache, grResolver *auth.GlobalRoleResolver) *GRBClusterRuleResolver {
+func NewGRBClusterRuleResolver(grbCache v3.GlobalRoleBindingCache, grResolver *auth.GlobalRoleResolver, sar authorizationv1.SubjectAccessReviewInterface) *GRBClusterRuleResolver {
 	grbCache.AddIndexer(grbSubjectIndex, grbBySubject)
 	return &GRBClusterRuleResolver{
 		GlobalRoleBindings: grbCache,
 		GlobalRoleResolver: grResolver,
+		sar:                sar,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/resolvers/grbClusterResolver.go
```diff
@@ -86,6 +129,67 @@ func (g *GRBClusterRuleResolver) VisitRulesFor(user user.Info, namespace string,
...
+				Version:  rbacv1.APIGroupAll,
+				Resource: rbacv1.ResourceAll,
+			},
+			User:   user.GetName(),
+			Groups: user.GetGroups(),
+			UID:    user.GetUID(),
+			Extra:  extras,
+		}}, metav1.CreateOptions{})
+	if err != nil {
+		return false, err
```
KevinJoiner, line 190: wrap error

pkg/resolvers/grbClusterResolver.go
```diff
@@ -1,34 +1,77 @@
...
+	exceptionServiceAccounts = []exceptionSA{
+		{
+			name:      "rancher-backup",
+			namespace: "cattle-resources-system",
+		},
+		{
+			name:      "fleet-agent",
+			namespace: "cattle-fleet-local-system",
+		},
+	}
```
KevinJoiner, line 34: If you resolve the names here we can remove the new struct and remove the member variable from the resolver.
```suggestion
	exceptionServiceAccounts = []string{
		serviceaccount.MakeUsername("cattle-resources-system", "rancher-backup"),
		serviceaccount.MakeUsername("cattle-fleet-local-system", "fleet-agent"),
	}
```

pkg/resolvers/grbClusterResolver.go
```diff
@@ -86,6 +129,67 @@ func (g *GRBClusterRuleResolver) VisitRulesFor(user user.Info, namespace string,
 			return
 		}
 	}
+	if g.isExceptionServiceAccount(user) {
```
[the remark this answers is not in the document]
MbolotSuse: Done

pjbgf (approved):
The service account exception logic does not really improve the security posture here. However, it does add to the cognitive load required to understand the logic - we probably could do without it.

Otherwise, LGTM

KevinJoiner (dismissed):
>The service account exception logic does not really improve the security posture here. However, it does add to the cognitive load required to understand the logic - we probably could do without it.

I agree this does not increase security, but it does stop service accounts from using GlobalRoles with the `inhertedClusterRoles` which we are not currently ready to support outside of specific use cases.

pjbgf (approved):

pkg/resolvers/grbClusterResolver_test.go
```diff
@@ -34,6 +41,9 @@ func TestGRBClusterRuleResolver(t *testing.T) {
 
 func (g *GRBClusterRuleResolverSuite) SetupSuite() {
 	g.userInfo = NewUserInfo("test-user", "test-group")
+	g.broUserInfo = NewUserInfo("system:serviceaccount:cattle-resources-system:rancher-backup", "system:serviceaccounts")
```
tomleb, line 44: Should we test the same but with a different namespace? From what I see we're not looking at a namespace specifically.

pkg/resolvers/grbClusterResolver.go
```diff
@@ -86,6 +109,74 @@ func (g *GRBClusterRuleResolver) VisitRulesFor(user user.Info, namespace string,
 			return
 		}
 	}
+	// if we aren't an exception service account, no need to check add sa-specific permissions
+	if !isExceptionServiceAccount(user) {
```
tomleb, line 113: Would it be less risky to start with this check?

Let me know how wrong I got this but here's what I think the requirement is:
- No rules should be returned for any serviceaccounts except for "exceptional" SA which passe a SAR `*` check.

So then, how risky would it be if any of the code above accumulate rules for a service account? Right now we assume that `GetByIndex` for groups and user won't return anything for a serviceaccount (and the tests below also hardcode this assumption). Is this always the case? Also, what if that code above is modified later to visit some rules and by mistake this leaks for SA too.

IOW, would it be safer / more future proof to do this:

```go
func (g *GRBClusterRuleResolver) VisitRulesFor(user user.Info, namespace string, visitor func(source fmt.Stringer, rule *rbacv1.PolicyRule, err error) bool) {
    if isAServiceAccount() {
       if isExceptionServiceAccount() {
        // handle exception
        }
     // Make sure no rules for SA leaks
     return
    }
    // Regular users
```

Does that make sense?

pkg/resolvers/grbClusterResolver.go
```diff
@@ -86,6 +109,74 @@ func (g *GRBClusterRuleResolver) VisitRulesFor(user user.Info, namespace string,
...
+
+// hasAdminPermissions checks if a given user is an admin
+func (g *GRBClusterRuleResolver) hasAdminPermissions(user user.Info) (bool, error) {
+	extras := map[string]v1.ExtraValue{}
+	for extraKey, extraValue := range user.GetExtra() {
+		extras[extraKey] = v1.ExtraValue(extraValue)
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	resourceResponse, err := g.sar.Create(ctx, &v1.SubjectAccessReview{
```
tomleb, line 137: Should we add a `nil` check here since the SAR is nil in some cases (tests mostly)?

tomleb (changes requested):

pkg/resolvers/grbClusterResolver_test.go
MbolotSuse: I don't think that's necessary - the code doesn't consider the namespace, so I'm not sure what value we gain by adding that test.

tomleb: I was thinking of adding it to prevent regressions. If we allow fleet in any namespace, then we should make sure that any namespace works in the future as well with the service account check, no? Wdyt?

MbolotSuse: That logic could result in effectively infinite tests though. If I test `my-namespace-1`, what stops someone from adding a check that works for `my-namespace-1` and `cattle-local-system` but not `my-namespace-2`?

tomleb (dismissed):

tomleb (approved):

KevinJoiner (approved):

pjbgf (approved):

PR #301, opened 2023-09-29
Make project.spec.ClusterName immutable

MbolotSuse:
The clusterName field of projects was already immutable in effect since it needed to match the namespace, which is immutable. This makes it explcitly immutable, and changes the check for cluster existence to only occur on create requests

## Issue: 
rancher/rancher#42999

## Problem
On both create and update, the webhook validates that the ClusterName field refers to a valid cluster. This causes issues when a cluster is deleted - a controller then needs to remove a finalizer added to a project to allow the project to be deleted. Since the cluster no longer exists, the finalizer can't be removed, and the project is indefinitely locked in a deleting state.

## Solution
Now, the clusterName integrity check (validating that the `management.cattle.io/Cluster` object specified by `ClusterName` exists) only runs on create. Additionally, update calls enforce immutability of the `ClusterName` field (since the field needed to match the namespace previously, and the namespace was immutable, this was already enforced). 

Since the integrity check is now done only on create, the finalizers can be removed as desired.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/project/validator.go
+++ b/pkg/resources/management.cattle.io/v3/project/validator.go
@@ -81,10 +81,16 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		return nil, fmt.Errorf("failed to get old and new projects from request: %w", err)
 	}
 
-	if request.Operation == admissionv1.Delete {
+	switch request.Operation {
+	case admissionv1.Create:
+		return a.admitCreate(newProject)
+	case admissionv1.Update:
+		return a.admitUpdate(oldProject, newProject)
+	case admissionv1.Delete:
 		return a.admitDelete(oldProject)
+	default:
+		return nil, admission.ErrUnsupportedOperation
 	}
-	return a.admitCreateOrUpdate(oldProject, newProject)
 }
 
 func (a *admitter) admitDelete(project *v3.Project) (*admissionv1.AdmissionResponse, error) {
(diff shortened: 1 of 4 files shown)


pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -94,20 +100,34 @@ func (a *admitter) admitDelete(project *v3.Project) (*admissionv1.AdmissionRespo
...
 	}
 	if fieldErr != nil {
 		return admission.ResponseBadRequest(fieldErr.Error()), nil
 	}
+	return a.admitCommonCreateUpdate(nil, project)
+}
+
+func (a *admitter) admitUpdate(oldProject, newProject *v3.Project) (*admissionv1.AdmissionResponse, error) {
+	if oldProject.Spec.ClusterName != newProject.Spec.ClusterName {
+		reason := fmt.Sprintf("clusterName is immutable, was: %s, is now: %s", oldProject.Spec.ClusterName, newProject.Spec.ClusterName)
```
KevinJoiner, line 116: Nit: update reason to match Upstream and CRTB validation
https://github.com/rancher/webhook/blob/release/v0.4/pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/validator.go#L133C1-L133C1
```
The Job "update-job" is invalid: spec.template: Invalid value: 
core.PodTemplateSpec{ObjectMeta:v1.ObjectMeta{Name:"", GenerateName:"", Namespace:"", SelfLink:"", 
UID:"", ResourceVersion:"", 
.... 
: field is immutable
```
```suggestion
		reason := "field is immutable"
```

KevinJoiner (changes requested):

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -94,20 +100,33 @@ func (a *admitter) admitDelete(project *v3.Project) (*admissionv1.AdmissionRespo
...
 	}
 	if fieldErr != nil {
 		return admission.ResponseBadRequest(fieldErr.Error()), nil
 	}
+	return a.admitCommonCreateUpdate(nil, project)
+}
+
+func (a *admitter) admitUpdate(oldProject, newProject *v3.Project) (*admissionv1.AdmissionResponse, error) {
+	if oldProject.Spec.ClusterName != newProject.Spec.ClusterName {
+		fieldErr := field.Invalid(projectSpecFieldPath.Child("clusterNameField"), newProject.Spec.ClusterName, "field is immutable")
```
maxsokolovsky, line 116: Should this be the following?
```suggestion
		fieldErr := field.Invalid(projectSpecFieldPath.Child("clusterName"), newProject.Spec.ClusterName, "field is immutable")
```

maxsokolovsky: Or, rather, the constant `clusterNameField` as elsewhere?

MbolotSuse: My bad, good catch.

MbolotSuse: Fixed.

KevinJoiner:
@tomleb 
See this [comment](https://github.com/rancher/rancher/issues/42999#issuecomment-1739972142) for more details on the exact scenario.
But when an item is marked for deletion, it is seen as deleted and removed from the cache.

KevinJoiner (approved):

tomleb:
Yeah I had read that issue and the comments from it.

> But when an item is marked for deletion, it is seen as deleted and removed from the cache.

That's an unexpected behavior (to me anyway). I would've thought that the object would stay in cache until all finalizers are removed, otherwise how do we act on the finalizers that are still there?

 Is this a upstream k8s thing or norman/wrangler? Honestly I don't remember how controller-runtime/operator-sdk behave in that case.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -94,20 +100,33 @@ func (a *admitter) admitDelete(project *v3.Project) (*admissionv1.AdmissionRespo
...
+func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (*admissionv1.AdmissionResponse, error) {
 	projectQuota := newProject.Spec.ResourceQuota
 	nsQuota := newProject.Spec.NamespaceDefaultResourceQuota
 	if projectQuota == nil && nsQuota == nil {
 		return admission.ResponseAllowed(), nil
 	}
-	fieldErr, err = checkQuotaFields(projectQuota, nsQuota)
+	fieldErr, err := checkQuotaFields(projectQuota, nsQuota)
 	if err != nil {
 		return nil, fmt.Errorf("error checking project fields: %w", err)
```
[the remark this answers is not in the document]
MbolotSuse: Given how late in the release cycle this is, and the fact that this relates to an active bug, I'd prefer if you stay focused on the content of this PR.

I can make this one change, but will not make any others that are outside the scope of this PR. Fair Enough?

tomleb (approved):

PR #302, opened 2023-10-02
Adds Support for Escalate and Bind verb to GRs/GRBs

MbolotSuse:
## Issue: 

Aims to resolve several issues with creating/updating GRs/GRBs

## Problem
GlobalRoles and GlobalRoleBindings can only be created/updated by users with equivalent permissions. This created issues for things like service accounts, which were unable to gain permissions in specific fields (such as inherited cluster roles).

## Solution
This PR adds support for two new verbs (both on GlobalRoles):
- escalate
  - Users with this verb can create GlobalRoles that have permissions higher than the permissions they possess
- bind
  - Users with this verb can create GlobalRoleBindings to GlobalRoles that have permissions higher than the permissions that they possess
  - Note that the permissions here must be on the GlobalRole that the binding is made to, not the binding crd

Upstream docs on this field can be found here: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#restrictions-on-role-binding-creation-or-update.

In addition, this PR also reverts:
- fc3f4d171c685754a76c7b0c73905eac7ad37457 (this is no longer necessary since the above will allow service accounts to function properly in this context)
- Parts of the 51930f5f52ac64bde1fe7284b008ace9fdff645c which allow for changes to metadata without the required permissions

## CheckList
  
- [x] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/globalrole/GlobalRole.md
+++ b/pkg/resources/management.cattle.io/v3/globalrole/GlobalRole.md
@@ -12,8 +12,6 @@ On create or update, the following checks take place:
 
 Users can only change GlobalRoles with rights less than or equal to those they currently possess. This is to prevent privilege escalation. This includes the rules in the RoleTemplates referred to in `inheritedClusterRoles`. 
 
-This escalation checking currently prevents service accounts from modifying GlobalRoles which include permissions on downstream clusters (such as Admin, Restricted Admin, or GlobalRoles which use the `inheritedClusterRoles` field).
-
 ### Builtin Validation
 
 The `globalroles.builtin` field is immutable, and new builtIn GlobalRoles cannot be created.
--- a/pkg/resources/management.cattle.io/v3/globalrole/GlobalRole.md
+++ b/pkg/resources/management.cattle.io/v3/globalrole/GlobalRole.md
@@ -12,6 +12,8 @@ On create or update, the following checks take place:
 
 Users can only change GlobalRoles with rights less than or equal to those they currently possess. This is to prevent privilege escalation. This includes the rules in the RoleTemplates referred to in `inheritedClusterRoles`. 
 
+This escalation check is bypassed if a user has the `escalate` verb on the GlobalRole that they are attempting to update. This can also be given through a wildcard permission (i.e. the `*` verb also gives `escalate`).
+
 ### Builtin Validation
 
 The `globalroles.builtin` field is immutable, and new builtIn GlobalRoles cannot be created.
--- a/pkg/resources/management.cattle.io/v3/globalrole/validator.go
+++ b/pkg/resources/management.cattle.io/v3/globalrole/validator.go
@@ -12,13 +12,14 @@ import (
 	objectsv3 "github.com/rancher/webhook/pkg/generated/objects/management.cattle.io/v3"
 	"github.com/rancher/webhook/pkg/resolvers"
 	"github.com/rancher/webhook/pkg/resources/common"
-	"golang.org/x/exp/slices"
+	"github.com/sirupsen/logrus"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	authorizationv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/kubernetes/pkg/registry/rbac/validation"
 	"k8s.io/utils/trace"
 )
@@ -29,14 +30,19 @@ var gvr = schema.GroupVersionResource{
 	Resource: "globalroles",
 }
 
-const roleTemplateClusterContext = "cluster"
+const (
+	roleTemplateClusterContext = "cluster"
+	escalateVerb               = "escalate"
+)
 
 // NewValidator returns a new validator used for validation globalRoles.
-func NewValidator(ruleResolver validation.AuthorizationRuleResolver, grbResolver *resolvers.GRBClusterRuleResolver) *Validator {
+func NewValidator(ruleResolver validation.AuthorizationRuleResolver, grbResolver *resolvers.GRBClusterRuleResolver,
+	sar authorizationv1.SubjectAccessReviewInterface) *Validator {
 	return &Validator{
 		admitter: admitter{
 			resolver:    ruleResolver,
 			grbResolver: grbResolver,
+			sar:         sar,
 		},
 	}
 }
@@ -69,6 +75,7 @@ func (v *Validator) Admitters() []admission.Admitter {
 type admitter struct {
 	resolver    validation.AuthorizationRuleResolver
 	grbResolver *resolvers.GRBClusterRuleResolver
+	sar         authorizationv1.SubjectAccessReviewInterface
 }
 
 // Admit is the entrypoint for the validator. Admit will return an error if it's unable to process the request.
@@ -91,11 +98,6 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 			// Object is in the process of being deleted, so admit it.
 			// This admits update operations that happen to remove finalizers.
 			// This is needed to supported the deletion of old GlobalRoles that would not pass the update check that verifies all rules have verbs.
-			return admission.ResponseAllowed(), nil
-		}
-		if isMetaOnlyChange(oldGR, newGR) {
-			// if this change only affects metadata, don't validate any further
-			// this allows users with the appropriate permissions to manage labels/annotations/finalizers
 			return admission.ResponseAllowed(), nil
 		}
 		fieldErr := a.validateUpdateFields(oldGR, newGR, fldPath)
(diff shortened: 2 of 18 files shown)


pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -126,14 +127,26 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
 	}
 
 	rules := a.grbResolver.GlobalRoleResolver.GlobalRulesFromRole(newGR)
 
 	err = auth.ConfirmNoEscalation(request, rules, "", a.resolver)
 	if err != nil {
-		return admission.ResponseFailedEscalation(err.Error()), nil
+		hasEscalate, escErr := a.hasEscalateVerb(request, newGR.Name, "")
+		if escErr != nil {
+			logrus.Warnf("Failed to check for the 'escalate' verb on GlobalRoles: %v", escErr)
```
MbolotSuse, line 145: Return here - we don't want to use a potentially invalid return arg.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -232,27 +245,34 @@ func (a *admitter) validateUpdateFields(oldRole, newRole *v3.GlobalRole, fldPath
...
+	if err != nil {
+		return false, fmt.Errorf("failed to checkout create sar request: %w", err)
 	}
-	return reflect.DeepEqual(oldGR.ObjectMeta, newGR.ObjectMeta)
+
+	return resp.Status.Allowed, nil
+
 }
 
 // grContentEqual checks if two globalRoles have equivalent content (all fields excluding metadata && typeMeta).
```
MbolotSuse, line 278: remove gr content equal and replace with reflect.DeepEquals(). Ignore delta to object meta, type meta, and new user default.

KevinJoiner (changes requested):

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -126,14 +127,28 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 	}
 	err = auth.ConfirmNoEscalation(request, clusterRules, "", a.grbResolver)
 	if err != nil {
-		return admission.ResponseFailedEscalation(err.Error()), nil
+		hasEscalate, escErr := auth.RequestUserHasVerb(request, gvr, a.sar, escalateVerb, newGR.Name, "")
```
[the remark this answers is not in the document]
MbolotSuse: We can't gaurantee that we made the request (i.e. if you aren't trying to escalate clusterRules but are trying to escalate in globalRules we still need to make the request later on). I can try to do something to cache the result, but it will make the code more complex.

My thinking was the extra network call (being that it only happens in cases where you are trying to escalate in both buckets) is rare enough to justfiy keeping the code as-is for quality reasons. But I can look at it again if you think the savings are worth it.

pkg/resources/management.cattle.io/v3/globalrole/validator_test.go
```diff
@@ -72,6 +76,22 @@ func TestAdmit(t *testing.T) {
 			},
 		},
 	}
+	setSarResponse := func(allowed bool, testErr error, targetUser string, targetGrName string, sarMock *k8fake.FakeSubjectAccessReviews) {
```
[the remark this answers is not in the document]
MbolotSuse: I generally don't like making side functions for tests that are on-offs to structure the test. But I can make it a non-anonymous function if you prefer.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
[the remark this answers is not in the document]
MbolotSuse: I changed to immediately return if you are ever detected to have bind permissions. Resolving this thread, that should solve the concerns around the extra network call.

pkg/resources/management.cattle.io/v3/globalrole/validator_test.go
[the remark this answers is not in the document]
MbolotSuse: Change made.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
```diff
@@ -119,13 +132,27 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 	}
 	err = auth.ConfirmNoEscalation(request, clusterRules, "", a.grbResolver)
 	if err != nil {
-		return admission.ResponseFailedEscalation(err.Error()), nil
+		hasBind, bindErr := auth.RequestUserHasVerb(request, globalRoleGvr, a.sar, bindVerb, globalRole.Name, "")
```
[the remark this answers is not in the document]
MbolotSuse: Resolving as per: https://github.com/rancher/webhook/pull/302#discussion_r1343142094

KevinJoiner (approved):

pkg/resources/management.cattle.io/v3/globalrolebinding/validator_test.go
```diff
@@ -241,11 +254,48 @@ func TestAdmit(t *testing.T) {
...
+					return baseGRB
+				},
+				oldGRB: func() *v3.GlobalRoleBinding { return nil },
+				stateSetup: func(ts testState) {
+					setSarResponse(true, nil, testUser, adminGR.Name, ts.sarMock)
+				},
+			},
+			allowed: true,
+		},
+		{
```
tomleb, line 280: For tests in general I think it would be great to verify that it works correctly when:
- resourceNames == "*"
- resourceNames == list of known values (minimally 1 value and 2 values)

I also wonder if we should have an integration test that covers the use case we're trying to allow: granting both escalate and bind to an SA so that it can create GR/GRBs that it previously couldn't. (Mostly for regression testing)

tomleb (changes requested):
LGTM, although I think it could do with more tests. Right now, we're setting the SAR request we'd expect in test and they're done in isolation. I think we should also have a test that verifies the use case with a CR/CRB that gives escalate+bind for GRs and a test user that creates a GR/GRB with privileges it doesn't have.

pkg/resources/management.cattle.io/v3/globalrole/GlobalRole.md
```diff
@@ -12,7 +12,7 @@ On create or update, the following checks take place:
 
 Users can only change GlobalRoles with rights less than or equal to those they currently possess. This is to prevent privilege escalation. This includes the rules in the RoleTemplates referred to in `inheritedClusterRoles`. 
 
-This escalation checking currently prevents service accounts from modifying GlobalRoles which include permissions on downstream clusters (such as Admin, Restricted Admin, or GlobalRoles which use the `inheritedClusterRoles` field).
+This escalation check is bypassed if a user has the `escalate` verb on the GlobalRole that they are attempting to update. This can also be given through a wildcard permission (i.e. the `*` verb also gives `escalate`).
```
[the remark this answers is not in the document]
MbolotSuse: I'll resolve the thread, but I agree that the docs should have more info on this.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -93,11 +99,6 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 			// This is needed to supported the deletion of old GlobalRoles that would not pass the update check that verifies all rules have verbs.
 			return admission.ResponseAllowed(), nil
 		}
-		if isMetaOnlyChange(oldGR, newGR) {
```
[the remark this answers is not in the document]
MbolotSuse: This was implemented so that Rancher's controllers could add labels/annotations related to the controllers without bypassing the webhook.

Since rancher has * on * in * groups, the bind/escalate verb allows us to work without checking for metadata only changes, which is expensive and hacky. As such, this was removed.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -124,15 +125,33 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+			return admission.ResponseAllowed(), nil
+		}
 		return admission.ResponseFailedEscalation(err.Error()), nil
 	}
 
 	rules := a.grbResolver.GlobalRoleResolver.GlobalRulesFromRole(newGR)
 
 	err = auth.ConfirmNoEscalation(request, rules, "", a.resolver)
 	if err != nil {
+		hasEscalate, escErr := auth.RequestUserHasVerb(request, gvr, a.sar, escalateVerb, newGR.Name, "")
```
pjbgf, line 147: The logic in `145`-`156` is largely the same as `129`-`141`. To ensure they are always in sync, I would extract them as a new `func` that is called at both places with diff params.

MbolotSuse: Would you mind elaborating on your idea here? Given that most of this is error handling, I think that whatever I would do here would end up with similar duplicated code in the admit function (if err != nil{do stuff} is a min of 3 lines).

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
```diff
@@ -119,12 +132,29 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+		if bindErr != nil {
+			logrus.Warnf("Failed to check for the 'bind' verb on GlobalRoles: %v", bindErr)
+			return admission.ResponseFailedEscalation(err.Error()), nil
+		}
+		if hasBind {
+			return admission.ResponseAllowed(), nil
+		}
 		return admission.ResponseFailedEscalation(err.Error()), nil
 	}
```
pjbgf, line 159: nit: this could potentially be extracted as a separate `func`, and used here and at `133`-`145`.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
pjbgf: The idea had two motivations:
- decrease the cognitive load by making it clear that the logic is the same, only the rules change.
- ensure that changes to the logic always happen at both places.

This was what I was thinking (instead of `129`-`158`):

```go
if res, err, ok := determineAdmission(request, clusterRules, "", a.grbResolver); ok {
  return res, err
}

rules := a.grbResolver.GlobalRoleResolver.GlobalRulesFromRole(newGR)
res, err, _ := determineAdmission(request, rules, "", a.grbResolver)
return res, err
```

MbolotSuse: Ok, thanks for elaborating. Let me see what I can do.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator_test.go
tomleb: Discussed elsewhere, putting the conclusion here:
> I'm not sure what you mean here. ...
I was talking about the `resourceNames` in the CR.

Both comments should have been about integration tests, not unit test.

We're okay with no integration test for this, considering the above. QA could potentially automate the testing of this usecase, unclear if will happen.

pjbgf (approved):

PR #305, opened 2023-10-25
Update docs

rohitsakala:


README.md
```diff
@@ -12,7 +12,7 @@ API server and perform validation or mutation of resources prior to their persis
 the cluster database.
 The validation and mutation are performed by custom code whose logic is defined by the instance of an admission controller.
 Cluster administrators may create many such controllers that perform custom validation or mutation in series.
-The Rancher Webhook is one such controller that runs only in the local cluster.
+The Rancher Webhook is one such controller that runs both in local cluster and all downstream clusters by default.
```
KevinJoiner, line 15: We don't provide an option for people to use Rancher without the webhook.
```suggestion
The Rancher Webhook is one such controller that runs both in local cluster and all downstream clusters.
```

KevinJoiner (commented):
Nice catch thanks for the update.

KevinJoiner (approved):

KevinJoiner (changes requested):
This LGTM, but I am blocking the merge until v2.8.0 is released.

README.md
```diff
@@ -12,7 +12,7 @@ API server and perform validation or mutation of resources prior to their persis
 the cluster database.
 The validation and mutation are performed by custom code whose logic is defined by the instance of an admission controller.
 Cluster administrators may create many such controllers that perform custom validation or mutation in series.
-The Rancher Webhook is one such controller that runs only in the local cluster.
+The Rancher Webhook is one such controller that runs both in local cluster and all downstream clusters automatically.
```
maxsokolovsky, line 15: Nit: I think the definite article is needed here, since we have only one local cluster, and it's also widely known to be special.
```suggestion
The Rancher Webhook is one such controller that runs both in the local cluster and all downstream clusters automatically.
```

KevinJoiner (approved):

PR #306, opened 2023-10-27
Release/v0.4

rohitsakala:
# Add validation and mutation for ClusterRepo

## New validations:

Both GitRepo and URl cannot exist simultaneously

## Testing 

Added unit tests


pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+		{
+			name: "Both GitRepo and URL are set when updating",
+			clusterRepo: &catalogv1.ClusterRepo{
+				Spec: catalogv1.RepoSpec{
+					URL:     "https://url.com",
+					GitRepo: "https://github.com",
+				},
+			},
+			operation:   admissionv1.Create,
```
pjbgf, line 46:
```suggestion
			operation:   admissionv1.Update,
```

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/stretchr/testify/assert"
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func TestClusterRepoValidation(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
maxsokolovsky, line 19: I would add a test case for when neither GitRepo, nor URL is set.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
```diff
@@ -0,0 +1,119 @@
...
+			},
+			operation:   admissionv1.Update,
+			wantAllowed: true,
+			wantErr:     false,
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
```
maxsokolovsky, line 76: You can run subtests in parallel, too.
```suggestion
		t.Run(test.name, func(t *testing.T) {
		      t.Parallel()
```

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
```diff
@@ -0,0 +1,119 @@
...
+			req, err := createClusterRepo(test.clusterRepo, test.operation, false)
+			assert.NoError(t, err)
+			validator := NewValidator()
```
maxsokolovsky, line 79: I am not sure you need to create a new validator and admitter for every test case - could you re-use one?

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
```diff
@@ -0,0 +1,119 @@
...
+			admitters := validator.Admitters()
+			assert.Len(t, admitters, 1)
+			response, err := admitters[0].Admit(req)
+			if test.wantErr {
+				assert.Error(t, err)
```
maxsokolovsky, line 84: Perhaps it would be good to also have an assertion regarding the Allowed status...

maxsokolovsky (changes requested):

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
rohitsakala: Done

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
rohitsakala: Done

pjbgf (approved):
Small nit, otherwise LGTM.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Admitters returns the admitter objects used to validate clusterrepos.
+func (v *Validator) Admitters() []admission.Admitter {
+	return []admission.Admitter{&v.admitter}
+}
+
+type admitter struct {
+}
+
+// Admit is the entrypoint for the validator. Admit will return an error if it unable to process the request.
```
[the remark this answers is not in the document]
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/stretchr/testify/assert"
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func TestClusterRepoValidation(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
maxsokolovsky, line 19: I think you're still missing some test cases. 
You have four conditions (both missing, only GitRepo is provided, only ClusterRepo is provided, both are provided) and two operations (create and update). This makes 8 cases.

maxsokolovsky (changes requested):

pkg/resources/catalog.cattle.io/v1/clusterrepo/ClusterRepo.md
```diff
@@ -0,0 +1,13 @@
+## Validation Checks
+
+### Invalid Fields - Create
+
+Users cannnot create ClusterRepo which violate the following constraints:
```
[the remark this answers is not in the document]
rohitsakala: Done

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
rohitsakala: Done. Added 8 cases

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator_test.go
maxsokolovsky: I think it's still missing here... when you expect an error.
`assert.Equal(t, test.wantAllowed, response.Allowed)`

rohitsakala: The response is nil. So when I check the response.Allowed it will panic.

maxsokolovsky: The response shouldn't be nil. It should have Allowed set to false. For that, you'd need to change your validation logic where an error condition doesn't result in a nil response. Will post a comment with an example on the appropriate line.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,87 @@
...
+// If this function is called without NewValidator(..) calls will panic.
+func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResponse, error) {
+	if request.Operation == admissionv1.Create || request.Operation == admissionv1.Update {
+		newClusterRepo, err := v1.ClusterRepoFromRequest(&request.AdmissionRequest)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get clusterRepo from request: %w", err)
+		}
+
+		err = a.validateFields(newClusterRepo)
+		if err != nil {
```
maxsokolovsky, line 67: Here, you'd need to return the main error from Admin based on what kind of error you got from `validateFields`.
You could do something like this: https://github.com/rancher/webhook/blob/e401287e9f7f36f1a3d718c68918b0a3b3227147/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go#L105.
Or just always return `admission.ResponseBadRequest(err.Error())`, as in that example above.

maxsokolovsky: The bottom line is that if you get a normal validation error, the admission response is not nil, it just will indicate that the request is not allowed (the value Allowed on the response is false).

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+
+	return admission.ResponseAllowed(), nil
+}
+
+func (a *admitter) validateFields(newClusterrepo *catalogv1.ClusterRepo) error {
+	// Both GitRepo and URL can't be filled simultanesously
+	if newClusterrepo.Spec.URL != "" && newClusterrepo.Spec.GitRepo != "" {
+		return fmt.Errorf("fields spec.URL and spec.GitRepo cannot exist together")
```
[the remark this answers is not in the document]
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
rohitsakala: Done

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,103 @@
...
+	if request.Operation == admissionv1.Create || request.Operation == admissionv1.Update {
+		newClusterRepo, err := v1.ClusterRepoFromRequest(&request.AdmissionRequest)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get clusterRepo from request: %w", err)
+		}
+
+		var fieldErr *field.Error
+		if err := a.validateFields(newClusterRepo); err != nil {
+			if errors.As(err, &fieldErr) {
+				return admission.ResponseBadRequest(err.Error()), nil
```
pjbgf, line 72:
```suggestion
				return admission.ResponseBadRequest(fieldErr.Error()), nil
```

rohitsakala: Done

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,103 @@
...
+				return admission.ResponseBadRequest(fieldErr.Error()), nil
+			}
+			return nil, fmt.Errorf("failed to validate fields on ClusterRepo: %w", err)
+		}
+	}
+
+	return admission.ResponseAllowed(), nil
+}
+
+func (a *admitter) validateFields(newClusterrepo *catalogv1.ClusterRepo) *field.Error {
```
maxsokolovsky, line 81: This method should return a regular error in the signature. And it's in the caller, `Admit`, that you inspect the actual type of the error.

rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			return nil, fmt.Errorf("failed to validate fields on ClusterRepo: %w", err)
+		}
+	}
+
+	return admission.ResponseAllowed(), nil
+}
+
+func (a *admitter) validateFields(newClusterrepo *catalogv1.ClusterRepo) error {
+	// Both GitRepo and URL can't be filled simultanesously
```
maxsokolovsky, line 82:
```suggestion
	// Both GitRepo and URL can't be specified simultaneously.
```
Though I don't even think you even need this comment. Up to you...

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,103 @@
...
+	if newClusterrepo.Spec.URL != "" && newClusterrepo.Spec.GitRepo != "" {
+		return &field.Error{
+			Type:     field.ErrorType("Mutually exclusive fields"),
+			Field:    "Mutually exclusive field values present",
+			BadValue: "fields spec.URL and spec.GitRepo cannot exist together",
+			Detail:   "",
+		}
+	}
+
+	// Atleast one of GitRepo or URL must be filled
```
maxsokolovsky, line 92:
```suggestion
	// Either GitRepo or URL must be specified.
```

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,103 @@
...
+	if newClusterrepo.Spec.URL == "" && newClusterrepo.Spec.GitRepo == "" {
+		return &field.Error{
```
maxsokolovsky, line 94: The actual error `response.Result.Message` strings are broken right now (you can print them to see).
You can actually use special error helpers.

To set this up, in the `Admit` method, create a fieldPath:
`fieldPath := field.NewPath("clusterrepo")`

Then pass it as an argument to your `validateFields` method. 
Then you can use the helper on it.
`return field.Forbidden(fieldPath, "either spec.URL or spec.GitRepo must be specified")`

You can follow the example for [PRTBs](https://github.com/rancher/webhook/blob/e401287e9f7f36f1a3d718c68918b0a3b3227147/pkg/resources/management.cattle.io/v3/projectroletemplatebinding/validator.go#L86).

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	return admission.ResponseAllowed(), nil
+}
+
+func (a *admitter) validateFields(newClusterrepo *catalogv1.ClusterRepo) error {
+	// Both GitRepo and URL can't be filled simultanesously
+	if newClusterrepo.Spec.URL != "" && newClusterrepo.Spec.GitRepo != "" {
+		return &field.Error{
```
maxsokolovsky, line 84: Use an error helper (example in another comment).
`return field.Forbidden(fieldPath, "spec.URL and spec.GitRepo cannot be specified both")`

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
```diff
@@ -0,0 +1,103 @@
...
+				return admission.ResponseBadRequest(fieldErr.Error()), nil
+			}
+			return nil, fmt.Errorf("failed to validate fields on ClusterRepo: %w", err)
+		}
+	}
+
+	return admission.ResponseAllowed(), nil
+}
+
+func (a *admitter) validateFields(newClusterrepo *catalogv1.ClusterRepo) error {
```
maxsokolovsky, line 81: Nit
```suggestion
func (a *admitter) validateFields(repo *catalogv1.ClusterRepo) error {
```

pkg/server/handlers.go
```diff
@@ -47,6 +48,10 @@ func Validation(clients *clients.Clients) ([]admission.ValidatingAdmissionHandle
 
 		handlers = append(handlers, psact, globalRoles, globalRoleBindings, prtbs, crtbs, roleTemplates, secrets, nodeDriver, projects)
 	}
+
+	clusterrepo := clusterrepo.NewValidator()
```
maxsokolovsky, line 52: To confirm, since I don't know about the general requirements, and there is no link to an issue: do we want this to run in both local and remote webhooks?

maxsokolovsky (changes requested):

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
rohitsakala: Done.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
rohitsakala: why should I change it from newClusterRepo to repo? The resource name is ClusterRepo not just repo. Also, it is the new signifies it is the latest one.

pkg/server/handlers.go
rohitsakala: Added the Github Issue. 

What are local and remote webhooks? Can you please point me to some docs? Thanks 👍🏽

rohitsakala: Yes, I want my validations to run on both local and downstream clusters.

pkg/resources/catalog.cattle.io/v1/clusterrepo/validator.go
maxsokolovsky: It's up to you, hence I marked it as a nit. At a minimum, you might want to make the casing look better, so `newClusterRepo` instead of `newClusterrepo`.
But you sort of repeat this composite name after the type name, which reads strangely. And it doesn't matter whether the repo is the latest one, you simply validate its fields. But it's all up to you.

maxsokolovsky (approved):

PR #307, opened 2023-10-31
Add validation to Toleration and Affinitys Keys

felipe-colussi:
## Problem:

We do not validate Tolerations nor Affinity rules, If a user passes a wrong key it just break the cluster with no error. 

## Solution

For toleration I added validations to the key as I didn't find any upstream validations for the object. 
For Affinity I found an upstream function to validate LabelSelector so I'm using it. 

## Problems /  Points to consider: 
* As the upstream validations work with "arrays" I tried to keep it taking all errors, it make the UI kinda strange if we have a lot of errors (but all of them can be fixed on 1 request). 
image

* The path pointers are not being checked, it can panic if the parent path is not passed by, this only can happen if someone changes the code, and this is done in the apimachinery too.  If needed I can add validations to create a empty path. 

## CheckList
  
- [ X ] Test

  Also  did some manual tests on the fields.

  
- [X ] Docs:  I'm not sure if they are good nor I have an idea of how to make them better.

### DIFF
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"strings"
 
 	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	"github.com/rancher/webhook/pkg/admission"
@@ -22,7 +23,9 @@ import (
 	k8sv1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	authorizationv1 "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	"k8s.io/utils/trace"
 )
@@ -32,8 +35,6 @@ const failureStatus = "Failure"
 
 var mgmtNameRegex = regexp.MustCompile("^c-[a-z0-9]{5}$")
 var fleetNameRegex = regexp.MustCompile("^[^-][-a-z0-9]+$")
-var affinityAndTolerationRegexString = "([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]"
-var affinityAndTolerationRegex = regexp.MustCompile("^" + affinityAndTolerationRegexString + "$")
 
 // NewProvisioningClusterValidator returns a new validator for provisioning clusters
 func NewProvisioningClusterValidator(client *clients.Clients) *ProvisioningClusterValidator {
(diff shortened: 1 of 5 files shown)


pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -302,10 +317,157 @@ func (p *provisioningAdmitter) validatePSACT(request *admission.Request, respons
...
+	var errList field.ErrorList
+	for k, s := range selector {
+		errList = append(errList, validation.ValidateLabelName(s.Key, path.Index(k).Child("key"))...)
+	}
+	return errList
+}
+
+// validateAppendToleration validate if tolerations follows the k8s standards
+// at the moment it only validates the key by calling validation.ValidateLabelName.
+func validateAppendToleration(toleration []k8sv1.Toleration, path *field.Path) field.ErrorList {
```
jiaqiluo, line 436: I wonder if we can use the function [ValidateTolerations](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L3902) in the k8s upstream code, to avoid handling all those complex logics & checks ourselves?

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -302,10 +317,157 @@ func (p *provisioningAdmitter) validatePSACT(request *admission.Request, respons
...
+		return nil
+	}
+	var errList field.ErrorList
+
+	errList = append(errList, validateAppendToleration(customization.AppendTolerations, path.Child("appendTolerations"))...)
+	errList = append(errList, validateAffinity(customization.OverrideAffinity, path.Child("overrideAffinity"))...)
+
+	return errList
+}
+func validateAffinity(overrideAffinity *k8sv1.Affinity, path *field.Path) field.ErrorList {
```
jiaqiluo, line 331: Similiar to my another comment: can we use the function [validateAffinity](https://github.com/kubernetes/kubernetes/blob/7b1c378edd7aa6a4e48cd324ee168c54a08a7702/pkg/apis/core/validation/validation.go#L3801) from the upstream code here?

jiaqiluo (changes requested):

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
felipe-colussi: @jiaqiluo , I would love to. 

I may be wrong but for what i read on the kubernetes/kubernetes  we are not supposed to use it as a "package", instead we should use a list that they publish: https://github.com/kubernetes/kubernetes/blob/master/staging/README.md. 

I wasn't able to find those functions on the components, the ones that were available were  ValidateLabelSelector and ValidateLabelName. 

If I 'm missing something please let me know.

jiaqiluo: We would not like to copy-paste code from outside even though the license is not an issue. In this case, the best way is to keep your current solution; we do not have to match the upstream.

jiaqiluo (approved):

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -200,3 +205,373 @@ func TestValidateMachinePoolName(t *testing.T) {
...
+			if !assert.ElementsMatch(t, s, errPaths) {
+				b := strings.Builder{}
+				b.WriteString("Failed Fields and reasons: ")
+				for _, v := range err {
+					b.WriteString("\n* ")
+					b.WriteString(v.Error())
+				}
+				fmt.Println(b.String())
+			}
+		}
```
KevinJoiner, line 232: This can all be defined outside of our test function without using an anonymous function.

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -200,3 +205,373 @@ func TestValidateMachinePoolName(t *testing.T) {
...
+func Test_validateAgentDeploymentCustomization(t *testing.T) {
+	type args struct {
+		customization *v1.AgentDeploymentCustomization
+		path          *field.Path
+	}
+	type validation func(t *testing.T, err field.ErrorList)
+
+	validateFailedPaths := func(s []string) validation {
+		return func(t *testing.T, err field.ErrorList) {
+			errPaths := make([]string, len(err), len(err))
```
KevinJoiner, line 218: specify helper functions for better stack trace on error.
```suggestion
            t.Helper()
			errPaths := make([]string, len(err), len(err))
```

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -1,12 +1,17 @@
 package cluster
```
KevinJoiner, line 1: None of the changes made to the `Admit(...)` function get hit by these tests. This could be handle by updating your current test to call Admit instead of private functions.

KevinJoiner (changes requested):

snasovich:
Closing this PR for now, will circle back if/when this issue gets escalated.

PR #308, opened 2023-11-01
Adds pather interface for handlers

KevinJoiner:
Adds the Pather interface, which allows handlers to create custom paths for routing.


tomleb:
Can we get more info on why this is necessary? 

My only guess is that we currently can't have two separate validators for `namespaces`, one for say `Delete` operation and another for `Update`, because they share the same GVK, so they share the same path and thus they must have operations be the union of `Delete` and `Update`. Is this what is being fixed?

PR #309, opened 2023-11-17
Add webhook checks for NamespacedRules

JonCrowther:
NOTE: Currently uses my local rancher changes because it requires work currently in a PR

## Problem
Follow up work from: https://github.com/rancher/rancher/pull/43018
There are 2 keys changes. First we add escalation checks for the new NamespacedRules field of GlobalRoles. Second we add a check to prevent modification/removal of the gr-owner and grb-owner labels.

## Solution
Added two new validation webhooks that select based on the gr-owner and grb-owner labels.
Added a check in the the GlobalRole validation to ensure there is no escalation.

## CheckList
  
- [x] Test
  
- [x] Docs


go.mod
```diff
@@ -5,6 +5,8 @@ go 1.20
 // on release remove this wrangler replace and use the latest tag
 replace github.com/rancher/wrangler v1.1.1 => github.com/rancher/wrangler v1.1.1-0.20230831050635-df1bd5aae9df
 
+replace github.com/rancher/rancher/pkg/apis => /Users/jcrowther/git/rancher/pkg/apis
```
JonCrowther, line 8: Placeholder until https://github.com/rancher/rancher/pull/43018 gets merged

pkg/resources/management.cattle.io/v3/globalrole/setup_test.go
```diff
@@ -188,7 +188,7 @@ func createGRRequest(t *testing.T, test testCase) *admission.Request {
 		username = testUser
 	}
 	gvk := metav1.GroupVersionKind{Group: "management.cattle.io", Version: "v3", Kind: "GlobalRole"}
-	gvr := metav1.GroupVersionResource{Group: "management.cattle.io", Version: "v3", Resource: "globalrolebindings"}
+	gvr := metav1.GroupVersionResource{Group: "management.cattle.io", Version: "v3", Resource: "globalroles"}
```
JonCrowther, line 191: noticed a typo while looking in this file

pkg/resources/management.cattle.io/v3/globalrole/validator.go
MbolotSuse, line 1: I think we also need validation here that all rules have at least one verb specified - otherwise our controller could fail to create the rule due to invalid rules with little/no signal to the end user.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
MbolotSuse, line 1: Similar comments on GRBs as on GRs with regard to:
- Checking for bind/escalate
- Error messages when a user attempts to escalate

eliyamlevy (approved):
From what I know and have seen this looks pretty good to me. I am still not too familiar with the role/rolebindings/globalrole/globalrolebindings objects but if the validators and mutators are all accounted for I approve.

pkg/resources/rbac.authorization.k8s.io/v1/role/validator_test.go
```diff
@@ -0,0 +1,128 @@
...
+		args    args
+		allowed bool
+		wantErr bool
+	}
+	tests := []test{
+		{
+			name: "create allowed",
+			args: args{
+				oldRole: defaultRole.DeepCopy(),
+				op:      admissionv1.Create,
```
MbolotSuse, line 54: Technically, the admitter shouldn't handle this scenario since it does not advertise as validating create requests. I would recommend adjusting the case accordingly.

pkg/resources/rbac.authorization.k8s.io/v1/role/validator_test.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+			allowed: true,
+		},
+		{
+			name: "create with gr-owner label allowed",
+			args: args{
+				oldRole: roleWithOwnerLabel.DeepCopy(),
+				op:      admissionv1.Create,
```
MbolotSuse, line 62: Same comment here - this admitter only handles update.

pkg/resources/rbac.authorization.k8s.io/v1/role/validator_test.go
```diff
@@ -0,0 +1,128 @@
...
+		newRole *v1.Role
+		op      admissionv1.Operation
+	}
+	type test struct {
+		name    string
+		args    args
+		allowed bool
+		wantErr bool
+	}
+	tests := []test{
```
MbolotSuse, line 49: I'd also recommend the following test cases:
- Old Role labels nil, new role has something -> allowed
- Old Role labels nil, new role labels nil (I know the selector prevents this, but the `admit` function should still not panic) -> allowed
- Old Role labels empty (`map[string]string{}`), new role has something -> allowed
- Old role has something, new role has labels empty (`map[string]string{}`) -> denied
- Old role labels empty (`map[string]string{}`), new role labels empty (`map[string]string{}`) -> allowed.

pkg/resources/rbac.authorization.k8s.io/v1/rolebinding/validator.go
```diff
@@ -0,0 +1,122 @@
...
+	if fieldErr != nil {
+		return admission.ResponseBadRequest(fieldErr.Error()), nil
+	}
+
+	return admission.ResponseAllowed(), nil
+}
+
+// validateGRBOwnerLabel checks that a user isn't removing or modifying the grbOwnerLabel.
+// Users can add the label if it didn't already exist.
+func validateGRBOwnerLabel(old, new *v1.RoleBinding, fldPath *field.Path) *field.Error {
```
MbolotSuse, line 97: This logic is very similar to the logic for the grOwnerLabel (and will probably be something that we use in the future). You might want to consider moving it to `pkg/resources/common`

pkg/resources/rbac.authorization.k8s.io/v1/rolebinding/validator_test.go
MbolotSuse, line 1: Similar comments on missing test cases.

MbolotSuse (changes requested):
Overall comment:
- GRB Docs also need to be updated.

Nothing too big, just some various issues to consider.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -121,49 +120,42 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
-			logrus.Warnf("Failed to check for the 'escalate' verb on GlobalRoles: %v", escErr)
-			return false, err
-		}
-		if hasEscalate {
-			return true, nil
+	// check for escalation in each namespace of NamespacedRules
+	for namespace, rules := range newGR.NamespacedRules {
+		err = auth.ConfirmNoEscalation(request, rules, namespace, a.resolver)
+		if err != nil {
+			return admission.ResponseFailedEscalation(err.Error()), nil
```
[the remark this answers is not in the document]
tomleb: To add to this a little bit, another UX that could suck is the following scenario:

1. User tries to create GR with namespaced rules: `nsA: rulesA` & `nsB: rulesB`. Both fails but user only sees nsA that fails first.
2. User adjust rules on `nsA`.
3. User tries again to create GR. Fails with `nsB`.
4. User adjust rules on `nsB`.
5. User tries again to create GR. Fails again on `nsA`.

Again, the fact that the map is not deterministic combined with returning on first error means the user can't know _for sure_ that the adjustments made were correct.

For this reason I'd be in favor of trying all the rules and returning all errors. I think the performance hit would be negligible since that's only done when a `GlobalRole` is either created or updated, which is most probably not done often.

pkg/resources/rbac.authorization.k8s.io/v1/role/validator.go
```diff
@@ -0,0 +1,117 @@
...
+	fieldErr := validateGROwnerLabel(oldRole, newRole, fldPath)
+	if fieldErr != nil {
+		return admission.ResponseBadRequest(fieldErr.Error()), nil
+	}
+
+	return admission.ResponseAllowed(), nil
+}
+
+// validateGROwnerLabel checks that a user isn't adding, removing or modifying the grOwnerLabel.
+// Users can add the label if it didn't already exist.
```
tomleb, line 91: I'm confused by the following:

> validateGROwnerLabel checks that a user isn't adding [...] the grOwnerLabel

and 

> Users can add the label if it didn't already exist.

Can user add the label or not?

tomleb (changes requested):
Seems good but I would like to see at least one unit tests handling multiple namespaces without the need for escalation (say `nsA` with rules `A` and `nsB` with rules `B`).

pkg/resources/management.cattle.io/v3/globalrole/validator.go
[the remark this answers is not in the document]
JonCrowther: You both make some great points here. I think returning a `multierror` with all the escalation failures (in Rules, ClusterRules and NamespacedRules) would definitely be an improvement on the UX. While you could infer that a single escalation failure may be relevant to the other Rules, having a full picture makes a lot of sense.

I think the performance hit is preferable to having partial information.

pkg/resources/rbac.authorization.k8s.io/v1/role/validator.go
JonCrowther: Good catch. I originally was going to prevent adding the label but opted to allow it because my GR changes in rancher/rancher will update the Role to add the label.

See here: https://github.com/rancher/rancher/blob/3284ace2d964abbf79a370a64eaabf8ecccda664/pkg/controllers/management/auth/globalroles/globalrole_handler.go#L300C1-L300C30

pkg/resources/management.cattle.io/v3/globalrole/validator.go
JonCrowther: I added a basic check to ensure that the list of verbs isn't empty. I couldn't find an existing validation method for rules. Would we like to check that the verbs/APIGroup are valid values too?

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -121,49 +120,42 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+	// check if the user has the escalate verb
+	hasEscalate, escErr := auth.RequestUserHasVerb(request, gvr, a.sar, escalateVerb, newGR.Name, "")
+	if escErr != nil {
+		logrus.Warnf("Failed to check for the 'escalate' verb on GlobalRoles: %v", escErr)
+	}
+
+	// if we have the escalate verb, no need to check global/cluster/namespaced permissions
+	if hasEscalate {
+		return admission.ResponseAllowed(), nil
+	}
```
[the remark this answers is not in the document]
JonCrowther: I am aware of this and am doing some investigation to see the performance impact. My hunch is you are correct since the SAR check is an API call vs just checking the cache for the rules and is therefore much slower and we should avoid it.

My only thought is that given there could be many rules to look at, there may be some threshold where it would be better to check the verb first. I am spending some time to investigate this, but in the mean time will address the rest of the comments.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
```diff
@@ -220,3 +223,21 @@ func (a *admitter) validateGlobalRole(globalRole *v3.GlobalRole, fieldPath *fiel
 	}
 	return nil
 }
+
+func (a *admitter) validateNamespacedRules(request *admission.Request, rules map[string][]v1.PolicyRule) []error {
```
tomleb, line 227: nit: Wondering if that should be shared with the other `validateNamespacedRules` functions. I understand that they're not exactly identical since for GRB you don't need to verify that there's at least one verb.

tomleb (approved):

pkg/resources/common/common.go
MbolotSuse, line 1: Since `IsModifyingLabel` is an exported function in the package, I would recommend writing tests for it.

pkg/resources/management.cattle.io/v3/globalrole/GlobalRole.md
```diff
@@ -10,9 +10,12 @@ On create or update, the following checks take place:
 
 ### Escalation Prevention
 
-Users can only change GlobalRoles with rights less than or equal to those they currently possess. This is to prevent privilege escalation. This includes the rules in the RoleTemplates referred to in `inheritedClusterRoles`. 
+ Escalation checks are bypassed if a user has the `escalate` verb on the GlobalRole that they are attempting to update. This can also be given through a wildcard permission (i.e. the `*` verb also gives `escalate`).
```
MbolotSuse, line 13: Nit: This applies to more than just update, Escalation checks are also present on create. I know that you didn't write this, if you could fix it I would appreciate it.

pkg/resources/management.cattle.io/v3/globalrole/GlobalRole.md
```diff
@@ -10,9 +10,12 @@ On create or update, the following checks take place:
...
 
-This escalation check is bypassed if a user has the `escalate` verb on the GlobalRole that they are attempting to update. This can also be given through a wildcard permission (i.e. the `*` verb also gives `escalate`).
+Users can only change GlobalRoles with rights less than or equal to those they currently possess. This is to prevent privilege escalation. This includes the rules in the RoleTemplates referred to in `inheritedClusterRoles`.
+
+Users can only have rules in the `NamespacedRules` field with rights less than or equal to those they currently possess. This works on a per namespace basis, meaning that the user must have the permission
```
MbolotSuse, line 17:
```suggestion
Users can only grant rules in the `NamespacedRules` field with rights less than or equal to those they currently possess. This works on a per namespace basis, meaning that the user must have the permission
```

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -121,49 +125,50 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+		allErrors = append(allErrors, err)
 	}
 
-	rules := a.grbResolver.GlobalRoleResolver.GlobalRulesFromRole(newGR)
-	// don't need to check if this request was allowed due to escalation since this is the last permission check
-	// if others are added in the future a short-circuit here will be needed like for the clusterRules
-	_, err = a.isRulesAllowed(request, rules, newGR.Name, a.resolver)
+	globalRules := a.grbResolver.GlobalRoleResolver.GlobalRulesFromRole(newGR)
+	err = validateRules(globalRules)
+	if err != nil {
```
MbolotSuse, line 155: I would recommend returning here rather than continuing on. I'm not sure how the SAR check will react to invalid rules being used in an escalation check - better to force the user to supply valid rules first IMO.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -264,3 +269,35 @@ func (a *admitter) validateUpdateFields(oldRole, newRole *v3.GlobalRole, fldPath
...
+func (a *admitter) validateNamespacedRules(request *admission.Request, rules map[string][]v1.PolicyRule) []error {
+	var returnErrs []error
+	// check for escalation in each namespace of the GlobalRole's NamespacedRules
+	for namespace, rules := range rules {
+		// check that namespace exists
+		_, err := a.namespaces.Get(namespace)
+		if err != nil {
+			returnErrs = append(returnErrs, fmt.Errorf("error getting namespace %v: %w", namespace, err))
+			continue
+		}
```
MbolotSuse, line 282: I thought about this a bit more and realized that this is poor UX. I think that we can use the status field/the controllers to indicate when a user has attempted to grant permissions in a non-existent namespace. That combined with the enqueuer on the namespace should do fine. Would you mind removing the check on Namespace existence for now?

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
```diff
@@ -220,3 +223,21 @@ func (a *admitter) validateGlobalRole(globalRole *v3.GlobalRole, fieldPath *fiel
...
+	var returnErrs []error
+	// check for escalation in each namespace of the GlobalRole's NamespacedRules
+	for namespace, rules := range rules {
+		// check that namespace exists
+		_, err := a.namespaces.Get(namespace)
+		if err != nil {
+			returnErrs = append(returnErrs, fmt.Errorf("error getting namespace %v: %w", namespace, err))
+			continue
+		}
```
MbolotSuse, line 236: I think the validation for this part should probably be on the GlobalRole rather than the binding (if it is anywhere).

pkg/resources/management.cattle.io/v3/globalrole/validator.go
MbolotSuse: I would recommend sending both `Rules` and `NamespaceRules` through the [validation used by upstream](https://github.com/kubernetes/kubernetes/blob/022d50fe3a1bdf2386395da7c266fede0c110040/pkg/apis/rbac/validation/validation.go#L103).

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
MbolotSuse, line 227: Nit: I would recommend changing the name of this to indicate the escalation checks. You don't really need to validate the rules, more that your binding isn't giving excessive rules.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator_test.go
```diff
@@ -672,6 +837,11 @@ func TestAdmit(t *testing.T) {
 			}
 			require.NoError(t, err)
 			require.Equalf(t, test.allowed, response.Allowed, "Response was incorrectly validated wanted response.Allowed = '%v' got '%v' message=%+v", test.allowed, response.Allowed, response.Result)
+			if test.respMessageContains != nil {
```
MbolotSuse, line 840: I would not recommend validating on the message. The message isn't part of the interface, it can contain an arbitrary string which is meaningful to the user, but from a code-perspective there's no guarantee on what is in that string.

pkg/resources/rbac.authorization.k8s.io/v1/role/validator.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Admitters returns the admitter objects used to validate roles.
+func (v *Validator) Admitters() []admission.Admitter {
+	return []admission.Admitter{&v.admitter}
+}
+
+type admitter struct{}
+
+// Admit is the entrypoint for the validator. Admit will return an error if it's unable to process the request.
+// If this function is called without NewValidator(..) calls will panic.
```
MbolotSuse, line 70: Is this true? This admitter doesn't seem to use any internal resources like a cache, so we shouldn't have a panic here.

pkg/resources/rbac.authorization.k8s.io/v1/rolebinding/validator.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Admitters returns the admitter objects used to validate roles.
+func (v *Validator) Admitters() []admission.Admitter {
+	return []admission.Admitter{&v.admitter}
+}
+
+type admitter struct{}
+
+// Admit is the entrypoint for the validator. Admit will return an error if it's unable to process the request.
+// If this function is called without NewValidator(..) calls will panic.
```
MbolotSuse, line 71: Similar comment here on if this is necessary.

MbolotSuse (changes requested):
A couple of minor nits as well as some small issues:
- There are other criteria for invalid rules in addition to verbs. I'd recommend including them all.
- Sorry to go back on you, but I think that the check to see if the namespace exists might do more harm then good.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -264,3 +269,35 @@ func (a *admitter) validateUpdateFields(oldRole, newRole *v3.GlobalRole, fldPath
 	}
 	return nil
 }
+
+func (a *admitter) validateNamespacedRules(request *admission.Request, rules map[string][]v1.PolicyRule) []error {
+	var returnErrs []error
+	// check for escalation in each namespace of the GlobalRole's NamespacedRules
+	for namespace, rules := range rules {
```
MbolotSuse, line 276: Nit: Given that rules is already the name for a variable in this function, I would recommend changing this to something like `namespaceRules` - it makes it easier to follow what is being used where.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -121,51 +122,55 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+	if err != nil {
+		if hasEscalate("") {
+			return admission.ResponseAllowed(), nil
+		}
+		returnError = errors.Join(returnError, err)
+	}
+	for namespace, rules := range newGR.NamespacedRules {
+		err = auth.ConfirmNoEscalation(request, rules, namespace, a.resolver)
+		if err != nil {
+			if hasEscalate(namespace) {
```
MbolotSuse, line 160: Since GlobalRoles are cluster-scoped resources, you want to check if the user has "Escalate" at the cluster scope. Otherwise, users could take advantage of odd rbac configurations to give escalate in one namespace but not others, which we want to avoid.

pkg/resources/management.cattle.io/v3/globalrole/validator.go
```diff
@@ -121,51 +122,55 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		return nil, err
 	}
 
-	// check for escalation separately between cluster permissions and global permissions to prevent crossover
+	var returnError error
+
+	// Validate the global and namespaced rules of the new GR
+	globalRules := a.grbResolver.GlobalRoleResolver.GlobalRulesFromRole(newGR)
+	returnError = errors.Join(returnError, validateRules(globalRules, false, fldPath))
```
MbolotSuse, line 129: Nit: I don't think we need the "errors.Join" call here since returnErr will be nil, so we could just do:

```suggestion
	returnError = validateRules(globalRules, false, fldPath))
```

But I might be missing something on this library.

pkg/resources/management.cattle.io/v3/globalrolebinding/validator.go
```diff
@@ -131,42 +131,38 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+			return admission.ResponseAllowed(), nil
 		}
-		if hasBind {
-			return true, nil
+		returnError = errors.Join(returnError, err)
+	}
+	for namespace, rules := range globalRole.NamespacedRules {
+		err = auth.ConfirmNoEscalation(request, rules, namespace, a.resolver)
+		if err != nil {
+			if hasBind(namespace) {
```
MbolotSuse, line 155: Similar comment here about checking for bind permissions in a specific namespace - GRs and GRBs are non-namespaced, so we want to check for this at the cluster scope.

pkg/resources/management.cattle.io/v3/globalrole/validator_test.go
```diff
@@ -545,6 +545,146 @@ func TestAdmit(t *testing.T) {
 			},
 			allowed: false,
 		},
+		{
```
MbolotSuse, line 548: Nit:
If you can add two tests with invalid rules (rules without verbs should do it), one for namespacedRules and one for Rules, I would appreciate it.

MbolotSuse (changes requested):
Couple of changes on the escalation checking logic, and some minor nits.

pkg/resources/common/common.go
```diff
@@ -14,3 +19,50 @@ func ConvertAuthnExtras(extra map[string]authnv1.ExtraValue) map[string]authzv1.
...
+		return true
+	}
+	if oldValue != newValue {
+		return true
+	}
+
+	return false
+}
+
+// NewVerbChecker returns a function that checks if a user has the specified verb
```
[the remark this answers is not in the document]
tomleb: The code becomes more complex if we want to 
1. Return all possible validation error instead of just the first one encountered
2. Want to do an escalation check whenever we find an error
3. Only do an escalation once

Caching whether we've done the escalation check and its result makes it simpler.

Though I think we'd want it to be clear that it's a "cached result" type of thing. `NewVerbChecker` doesn't really describes that.

I think either renaming the function (and changing its logic a bit) would make it clearer.

Something like:

```go
type cachedEscalationChecker struct {
  checked bool
  result bool
}

func NewCachedEscalationChecker() *cachedEscalationChecker {
    return cacheEscalationChecker{}
}

func (c *cachedEscalationChecker) CanEscalate(...) (bool, error) {
    if c.checked {
        return c.result // return result, not false
    }
    // actually do the escalation check
    c.checked = true
    c.result = <check>
    return c.result
}
```

Then using it is pretty much unchanged from how the PR uses it.

(Also, naming things is hard, huh...)

>  I think this additional abstraction is adding complexity in an are where it can be dangerous

I'd say no abstraction here was more complex imo and more difficult to reason about so more dangerous.

What's your thought on the proposed changes above, would that still be too abstracted in your opinion?

pkg/resources/management.cattle.io/v3/globalrole/validator.go
tomleb: Could we add a test here to make the current implementation fail so that we don't introduce this mistake in the future?

tomleb (dismissed):
@MbolotSuse 's got a couple of good comments. Nothing to add on top.

pkg/resources/common/common.go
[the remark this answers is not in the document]
MbolotSuse: > The code becomes more complex if we want to

The first requirement is more fungible than the others, but sure, there is definitely some extra complexity introduced.

> Caching whether we've done the escalation check and its result makes it simpler

The issue here is that this code doesn't effectively handle that. 

> Do you think the proposed changes above would make it less complex? Or are you opposed in a caching behavior for this area of the code?

I don't think that the proposed changes reduce complexity, particularly with the issues that it introduces. I'm not opposed to making sure that the escalate verb is checked for at most one time, but I don't think this abstraction solves that problem in a secure fashion.

I think that maintaining the state in the Admit function with two bools (and possibly using a side function to do the debouncing/caching) is a better option. However, I'm open to discussion.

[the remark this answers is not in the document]
tomleb: > using a side function to do the debouncing/caching

Can you expand on this? Sounds like something similar to above but more "local" to the code that needs it.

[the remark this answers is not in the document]
MbolotSuse: Sure, I'm thinking something like:

```golang

func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResponse, error) {
// some other code before
var hasEscalate, escalateChecked bool
err = auth.ConfirmNoEscalation(request, clusterRules, "", a.grbResolver)
if err != nil {
	hasEscalate, escErr := a.hasEscalate(hasEscalate, escalateChecked, request, newGR.Name)
	if err != nil {
		return nil, escErr
	}
	escalateChecked = true
	if hasEscalate {
	}
	returnError = errors.Join(returnError, err)
}

// some other code after
}

func (a *admitter) hasEscalate(alreadyChecked bool, prevAnswer bool, request *admission.Request, grName string) (bool, error) {
	if alreadyChecked {
		return prevAnswer, nil
	}
	hasVerb, err := auth.RequestUserHasVerb(request, gvr, a.sar, escalateVerb, grName, "")
	if err != nil {
		return false, fmt.Errorf("unable to check if user has escalate: %w", err)
	}
	return hasVerb, nil
}

```

This has the following benefits over the current approach:
- The specific requirements of the GR/GRB validator (i.e. escalate is only checked once, but we only care about that one resource for any given request) are kept internal to that validator
- The state of what was called and when is kept inside the admit function, which is directing the flow and maintaining the state for this call (rather than in another call elsewhere

This is just my initial answer, if you see other ways to structure the side functions I'm also open to that. My key concern here is making sure that what we do here

pkg/resources/common/common.go
```diff
@@ -14,3 +22,79 @@ func ConvertAuthnExtras(extra map[string]authnv1.ExtraValue) map[string]authzv1.
...
+	if oldValue != newValue {
+		return true
+	}
+
+	return false
+}
+
+// cachedVerbChecker is used for caching if a request for a non-namespaced gvr with specified name has the given overrideVerb.
+// This is meant to eliminate the need to perform multiple calls to the provided SubjectAccessReview for the overrideVerb.
+type cachedVerbChecker struct {
```
MbolotSuse, line 56: A couple of overall comments on this struct:
- From what I've heard, typically in go you want to avoid constructors and allow objects to be constructed directly. @KevinJoiner might feel more strongly than I do on that point, and may have an opinion on if this is permissible in this case.
- It's a bit odd to have an unexported type, which doesn't implement an interface, returned from an exported function. I think your CI is failing because of this behavior (and an associated revive check) so you will probably need to adjust this
- I'd recommend emphasizing in your comment the limitations of this struct. Namely:
  - Each admitter must request a new verbChecker for each request.
  - Verb checkers must not be shared across different admitters.
  - If the caller needs to change what it is checking (regardless of if it's a different verb, resource name, or resourceType), then they need to make a separate verb checker.
 The reason I'm emphasizing these limitations is that this struct is used for access control, so a bug here is likely to turn into a CVE. We need to be very clear here to avoid issues.

pkg/resources/common/common.go
```diff
@@ -14,3 +22,79 @@ func ConvertAuthnExtras(extra map[string]authnv1.ExtraValue) map[string]authzv1.
...
+		if escErr != nil {
+			err = errors.Join(err, escErr)
+		}
+		c.hasVerbBeenChecked = true
+		c.hasVerb = hasVerb
+	}
+	return err
+}
+
+// HasVerb returns if the request have the overrideVerb. Defaults to false if it hasn't been checked.
```
MbolotSuse, line 97: Nit:
```suggestion
// HasVerb returns if the request has the overrideVerb. Defaults to false if it hasn't been checked.
```

pkg/resources/common/common_test.go
MbolotSuse, line 1: I think you are aware of this, but we should get some tests for the VerbChecker here as well. Leaving this comment more as a placeholder.

MbolotSuse (commented):
Couple of comments on the new verb checker.

tomleb (approved):

pkg/resources/common/common.go
```diff
@@ -51,40 +51,30 @@ func IsModifyingLabel(oldLabels, newLabels map[string]string, label string) bool
...
-		hasVerbBeenChecked: false,
-		hasVerb:            false,
-	}
-}
-
 // IsRulesAllowed checks if the request has permissions to create the rules provided.
 // If not it checks if the user has the needed overrideVerb to bypass the check.
-func (c *cachedVerbChecker) IsRulesAllowed(rules []v1.PolicyRule, resolver validation.AuthorizationRuleResolver, namespace string) error {
-	err := auth.ConfirmNoEscalation(c.request, rules, namespace, resolver)
+func (c *CachedVerbChecker) IsRulesAllowed(rules []v1.PolicyRule, resolver validation.AuthorizationRuleResolver, namespace string) error {
```
MbolotSuse, line 71: Right now this returns an error if the user is trying to escalate but has the verb for override. I think that's a bit of a confusing UX since the presence of the override verb should allow the action - but the function requires a separate check for the consuming area to see that happened. 

In short, right now in override scenarios this returns an error - I think it should not.

pkg/resources/common/common.go
```diff
@@ -94,7 +84,7 @@ func (c *cachedVerbChecker) IsRulesAllowed(rules []v1.PolicyRule, resolver valid
 	return err
 }
 
-// HasVerb returns if the request have the overrideVerb. Defaults to false if it hasn't been checked.
-func (c *cachedVerbChecker) HasVerb() bool {
+// HasVerb returns if the request has the overrideVerb. Defaults to false if it hasn't been checked.
```
MbolotSuse, line 87: Right now this won't check if the user has an override verb if IsRulesAllowed hasn't been called. You do note this  in the function doc string, so you can keep it this way if you want, but I think the better UX would be to check if the user has the verb if we haven't already, given that the object should have all of the tools necessary to do so.

pkg/resources/common/common_test.go
```diff
@@ -1,7 +1,18 @@
...
+	authorizationv1 "k8s.io/api/authorization/v1"
+	v1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/authentication/user"
+	k8fake "k8s.io/client-go/kubernetes/typed/authorization/v1/fake"
+	k8testing "k8s.io/client-go/testing"
 )
 
 func TestIsModifyingLabel(t *testing.T) {
```
MbolotSuse, line 18: I think there's one test missing from this - the case where a the labels exist and are populated but there's no change at all. This causes a line to be missing from the coverage stats of the test, would you mind adding that?

MbolotSuse (changes requested):
Couple of small changes on the new CachedVerbChecker.

pkg/resources/common/common.go
```diff
@@ -14,3 +20,82 @@ func ConvertAuthnExtras(extra map[string]authnv1.ExtraValue) map[string]authzv1.
...
+	}
+	return err
+}
+
+// HasVerb returns if the request has the overrideVerb. Only checks the request the first time called, after that it returns the cached value.
+func (c *CachedVerbChecker) HasVerb() bool {
+	if c.hasVerbBeenChecked {
+		return c.hasVerb
+	}
+	c.hasVerb, _ = auth.RequestUserHasVerb(c.request, c.gvr, c.sar, c.overrideVerb, c.name, "")
```
MbolotSuse, line 98: Nit: Generally speaking you never want to ignore an error. In this case, I think we can allow it since the calling function is one that we wrote and the functionality is pretty clear on the cases where it returns an error, but it's definitely something that we should avoid. I'd go with something like:

```suggestion
	c.hasVerb, err = auth.RequestUserHasVerb(c.request, c.gvr, c.sar, c.overrideVerb, c.name, "")
        if err != nil{
                // Ideally this wouldn't log and would return up to the calling function/admitters, but this would be acceptable
	        logrus.Errorf("got an error when checking if a user has a verb %s", err)
	        return false
	}
```

This also has the disadvantage of "eating the error" so if there's something wrong with the SAR logic we will have no way of telling.

MbolotSuse (changes requested):
One comment on an unhandled error. Outside of that looks good.

pkg/resources/common/common.go
JonCrowther: I added a log and made a test for it

MbolotSuse (approved):

tomleb (approved):

PR #316, opened 2024-01-10
Add validation to Toleration and Affinitys Keys

thatmidwesterncoder:
## Problem:

We do not validate Tolerations nor Affinity rules, If a user passes a wrong key it just break the cluster with no error. 

## Solution

For toleration I added validations to the key as I didn't find any upstream validations for the object. 
For Affinity I found an upstream function to validate LabelSelector so I'm using it. 

## Problems /  Points to consider: 
* As the upstream validations work with "arrays" I tried to keep it taking all errors, it make the UI kinda strange if we have a lot of errors (but all of them can be fixed on 1 request). 
image

* The path pointers are not being checked, it can panic if the parent path is not passed by, this only can happen if someone changes the code, and this is done in the apimachinery too.  If needed I can add validations to create a empty path. 

## CheckList
  
- [ X ] Test

  Also  did some manual tests on the fields.

  
- [X ] Docs:  I'm not sure if they are good nor I have an idea of how to make them better.


docs.md
```diff
@@ -276,3 +276,26 @@ If `roletemplates.builtin` is true then all fields are immutable except:
...
+
+# provisioning.cattle.io/v1 
+
+## Cluster 
+
+### Validation Checks
+
+#### cluster.spec.clusterAgentDeploymentCustomization and cluster.spec.fleetAgentDeploymentCustomization
+
+`Key rule`: The keys are being validated to match the kubernetes requirement. 63 Characteres maximum, Regex:
```
HarrisonWAffel, line 288: Odd wording, maybe `The keys being validated against the kubernetes requirements`, or something along those lines?

docs.md
```diff
@@ -276,3 +276,26 @@ If `roletemplates.builtin` is true then all fields are immutable except:
...
+`([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]` it also accepts a "/"
+
+`appendTolerations`:  is being validated with the  `{Key Rule}` which live in [this apimachinery function](https://github.com/kubernetes/apimachinery/blob/02a41040d88da08de6765573ae2b1a51f424e1ca/pkg/apis/meta/v1/validation/validation.go#L96)
```
HarrisonWAffel, line 291: Odd wording, maybe, '`appendTolerations`: is validated against the '{key Rule}' which lives in [this api machinery function](fakeurl)' ?

docs.md
```diff
@@ -276,3 +276,26 @@ If `roletemplates.builtin` is true then all fields are immutable except:
...
+
+`affinity.nodeAffinity`:  The nodeSelectorTerms  (`requiredDuringSchedulingIgnoredDuringExecution` and
+`preferredDuringSchedulingIgnoredDuringExecution.preference`) are being validated with the `{Key Rule}`.
+
+`affinity.podAffinity and affinity.podAntiAffinity`:  The labelSelectors
```
HarrisonWAffel, line 296: Looks like an extra back-tick is included here, which breaks the highlighting on 'preferredDuringSchedulingIgnoredDuringExecution.podAffinityTerm'

HarrisonWAffel (commented):
I have a few comments regarding how the docs taken from the original are worded / formatted, and I'm not sure that https://github.com/rancher/webhook/pull/307/files#r1431633048 was addressed. If there was a reason why the above comment wasn't addressed lmk and I'll approve, other than some nits this lgtm

thatmidwesterncoder:
@HarrisonWAffel I ended up redoing the docs piece - lmk what you think!

I'm working on using the `Admit()` function currently - it might just take a bit more than expected since none of the other tests use the admit function either.

HarrisonWAffel (commented):
Thanks for updating the docs! This looks good, I'll give a final approve once `Admit` is used in the unit test, or we agree that it isn't needed

nflynt:
Hrm... since the Admit additions are essentially just wrapping `validateAgentDeploymentCustomization` and that **is** tested, my gut is that additional tests on the Admit function would be redundant. But I won't stop you if you want to add them.

nflynt (approved):
I'm able to follow the validation logic reasonably well and I can see that all of Kevin's original feedback has been addressed. There's an argument to be made that we should be testing through the exported `Admit()` function since that's the behavior we are trying to guarantee, but that's ultimately a nitpick and I won't block on it. The new business logic is covered decently in the private test suite. :+1:

ericpromislow (approved):
Verified that it works when deployed, and the tests are good.

PR #318, opened 2024-01-12
Modify `fleetNameRegex` to better conform to RFC 1123

aditditto:
## Issue: 

rancher/rancher#43535
## Problem

As the issue described, the validator for cluster creation throws an error when a cluster with single character names. From what I can tell I don't think this is intentional, so instead of changing the error message, I opted to fix the validating Regex to accept names with length of 1 character.

In addition, I also changed the  
## CheckList
  
- [x] Test
  
- [x] Docs


crobby (approved):
lgtm

PR #323, opened 2024-01-18
[DNM] Validate container default resource limit on projects

maxsokolovsky:
## Problem

The Webhook doesn't validate container resource default limits that are associated with a new or updated Project. This results in controllers responsible for making the corresponding limit ranges in the project's namespaces to fail asynchronously. The K8s API server rejects limit ranges whose requests are greater than limits for any resource.
 
## Solution

Add validation to Webhook on Project create and update.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -189,6 +189,11 @@ The system project cannot be deleted.
 
 Project quotas and default limits must be consistent with one another and must be sufficient for the requirements of active namespaces.
 
+#### Container default resource limit validation
+
+Container default resource limit configuration must have no negative values for any requests and limits.
+Requests cannot exceed limits for any resource.
+
 ### Mutations
 
 #### On create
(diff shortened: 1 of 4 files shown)


crobby (commented):
Is the target branch (release/v0.4) correct for this?

maxsokolovsky:
@crobby, yes, this is for 2.8's webhook. I created this PR too soon, thinking it's OK to first make it for 2.8 (since the original issue is for 2.8's milestone), then forward-port to 2.9. But we're actually supposed to make all PRs for 2.9, then back port them to 2.8. Once the 2.9 PR is merged, I'll cherry-pick the commit and thus replace the contents of this PR.

maxsokolovsky:
This won't be backported to 2.8.

PR #326, opened 2024-01-22
Validate container default resource limit on projects

maxsokolovsky:
## Problem

The Webhook doesn't validate container resource default limits that are associated with a new or updated Project. This results in controllers responsible for making the corresponding limit ranges in the project's namespaces to fail asynchronously. The K8s API server rejects limit ranges whose requests are greater than limits for any resource.
 
## Solution

Add validation to Webhook on Project create and update.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -192,6 +192,13 @@ The system project cannot be deleted.
 
 Project quotas and default limits must be consistent with one another and must be sufficient for the requirements of active namespaces.
 
+#### Container default resource limit validation
+
+Validation mimics the upstream behavior of the Kubernetes API server when it validates LimitRanges.
+The container default resource configuration must have properly formatted quantities for all requests and limits.
+
+Limits for any resource must not be less than requests.
+
 ### Mutations
 
 #### On create
--- a/pkg/resources/management.cattle.io/v3/project/validator.go
+++ b/pkg/resources/management.cattle.io/v3/project/validator.go
@@ -1,6 +1,7 @@
 package project
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -14,6 +15,7 @@ import (
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/utils/trace"
@@ -24,6 +26,7 @@ const (
 	projectQuotaField   = "resourceQuota"
 	clusterNameField    = "clusterName"
 	namespaceQuotaField = "namespaceDefaultResourceQuota"
+	containerLimitField = "containerDefaultResourceLimit"
 )
 
 var projectSpecFieldPath = field.NewPath("project").Child("spec")
@@ -123,6 +126,10 @@ func (a *admitter) admitUpdate(oldProject, newProject *v3.Project) (*admissionv1
 func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (*admissionv1.AdmissionResponse, error) {
 	projectQuota := newProject.Spec.ResourceQuota
 	nsQuota := newProject.Spec.NamespaceDefaultResourceQuota
+	containerLimit := newProject.Spec.ContainerDefaultResourceLimit
+	if fieldErr := a.validateContainerDefaultResourceLimit(containerLimit); fieldErr != nil {
+		return admission.ResponseBadRequest(fieldErr.Error()), nil
+	}
 	if projectQuota == nil && nsQuota == nil {
 		return admission.ResponseAllowed(), nil
 	}
@@ -141,6 +148,41 @@ func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (
 		return admission.ResponseBadRequest(fieldErr.Error()), nil
 	}
 	return admission.ResponseAllowed(), nil
+}
+
+// validateContainerDefaultResourceLimit checks all resource requests and limits.
+// It returns a fieldError. If the method is ever changed to also return a regular error, the caller's logic
+// needs to be updated to act appropriately based on the kind of error.
+func (a *admitter) validateContainerDefaultResourceLimit(limit *v3.ContainerResourceLimit) error {
+	if limit == nil {
+		return nil
+	}
+	fieldPath := projectSpecFieldPath.Child(containerLimitField)
+	requestsCPU, err := parseResource(limit.RequestsCPU)
+	if err != nil {
+		return field.Invalid(fieldPath, limit.RequestsCPU, fmt.Sprintf("failed to parse container default requested CPU: %s", err))
+	}
+	limitsCPU, err := parseResource(limit.LimitsCPU)
+	if err != nil {
+		return field.Invalid(fieldPath, limit.LimitsCPU, fmt.Sprintf("failed to parse container default CPU limit: %s", err))
+	}
+	requestsMemory, err := parseResource(limit.RequestsMemory)
+	if err != nil {
+		return field.Invalid(fieldPath, limit.RequestsMemory, fmt.Sprintf("failed to parse container default requested memory: %s", err))
+	}
+	limitsMemory, err := parseResource(limit.LimitsMemory)
+	if err != nil {
+		return field.Invalid(fieldPath, limit.LimitsMemory, fmt.Sprintf("failed to parse container default memory limit: %s", err))
+	}
+	if requestsCPU != nil && limitsCPU != nil && requestsCPU.Cmp(*limitsCPU) > 0 {
+		fieldErr := field.Invalid(fieldPath, limit, fmt.Sprintf("requested CPU %s is greater than limit %s", limit.RequestsCPU, limit.LimitsCPU))
+		err = errors.Join(err, fieldErr)
+	}
+	if requestsMemory != nil && limitsMemory != nil && requestsMemory.Cmp(*limitsMemory) > 0 {
+		fieldErr := field.Invalid(fieldPath, limit, fmt.Sprintf("requested memory %s is greater than limit %s", limit.RequestsMemory, limit.LimitsMemory))
+		err = errors.Join(err, fieldErr)
+	}
+	return err
 }
 
 func (a *admitter) checkClusterExists(project *v3.Project) (*field.Error, error) {
@@ -248,4 +290,13 @@ func formatResourceList(resources v1.ResourceList) string {
 	// sort the results for consistent log output
 	sort.Strings(resourceStrings)
 	return strings.Join(resourceStrings, ",")
+}
+
+func parseResource(s string) (*resource.Quantity, error) {
+	if s == "" {
+		// Upstream `resource.ParseQuantity` will return an error when given an empty string.
+		return nil, nil
+	}
+	q, err := resource.ParseQuantity(s)
+	return &q, err
 }
--- a/pkg/resources/management.cattle.io/v3/project/validator_test.go
+++ b/pkg/resources/management.cattle.io/v3/project/validator_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/golang/mock/gomock"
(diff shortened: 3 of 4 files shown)


pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -24,6 +27,7 @@ const (
 	projectQuotaField   = "resourceQuota"
 	clusterNameField    = "clusterName"
 	namespaceQuotaField = "namespaceDefaultResourceQuota"
+	containerLimitField = "namespaceDefaultResourceQuota"
```
MbolotSuse, line 30: Is this accurate? I think this should be changed to:
```suggestion
	containerLimitField = "containerDefaultResourceLimit"
```

Let me know if I am misunderstanding or have something wrong there.

pkg/resources/management.cattle.io/v3/project/validator_test.go
MbolotSuse, line 1: There's a couple of other test cases that I think would be good to add:
- CPU limit negative
- Memory Requests negative
- CPU limit < req on update
- Memory limit < req (only, not both CPU + Memory) on create
- Unparseable values for CPU Limit, CPU Request, Memory Limit, and Memory Request (this is a string format field, would be good to have a test case for random user input like "Example").

As far as testing this on both create and update goes, optimally we would have full coverage of both create and update (i.e. over on create, over on update, equal on create, equal on update, less than on create, less than on update for both CPU and Mem, negative on create, negative on update, etc.). There are ways to add these test cases so that you don't have to fully enumerate out both the create and update cases, but we can discuss further on if that would be helpful if we try it and it gets to be too much.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -143,6 +151,48 @@ func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (
...
 }
 
+func (a *admitter) validateContainerDefaultResourceLimit(limit *v3.ContainerResourceLimit) error {
+	if limit == nil {
+		return nil
+	}
+	fieldPath := projectSpecFieldPath.Child(containerLimitField)
+	requestsCPU, err := resource.ParseQuantity(limit.RequestsCPU)
+	if err != nil {
+		return field.Invalid(fieldPath, limit.RequestsCPU, "failed to parse container default requested CPU")
```
MbolotSuse, line 161: I think it would be good in this case (and the others) to preserve the parsing error message as well:
```suggestion
		return field.Invalid(fieldPath, limit.RequestsCPU, fmt.Sprintf("failed to parse container default requested CPU: %s", err.Error()))
```

MbolotSuse (changes requested):
Couple of small concerns and missed test cases.

pkg/resources/management.cattle.io/v3/project/validator.go
maxsokolovsky: Nope, it's just my oversight.

pkg/resources/management.cattle.io/v3/project/validator.go
maxsokolovsky: Will adjust.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -143,6 +151,48 @@ func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (
 	return admission.ResponseAllowed(), nil
 }
 
+func (a *admitter) validateContainerDefaultResourceLimit(limit *v3.ContainerResourceLimit) error {
+	if limit == nil {
+		return nil
+	}
+	fieldPath := projectSpecFieldPath.Child(containerLimitField)
+	requestsCPU, err := resource.ParseQuantity(limit.RequestsCPU)
```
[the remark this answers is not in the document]
maxsokolovsky: Good point, will restructure the logic to allow for such scenarios.

pkg/resources/management.cattle.io/v3/project/validator_test.go
```diff
@@ -901,7 +908,410 @@ func TestProjectValidation(t *testing.T) {
 			wantAllowed: false,
 			wantErr:     true,
 		},
+		{
```
tomleb, line 911: I think it would be worth splitting the test into two. The old test was testing mostly quota related stuff, we could introduce a new `TestProjectLimits` or something that just tests the limits. This can also shorten the amount of code written by having a test table that only takes in the `v3.ContainerResourceLimit` struct instead of the whole `v3.Project`.

Something like:

```go
func TestProjectLimit(t *testing.T) {
	t.Parallel()
	tests := []struct {
		name        string
		newLimit    *v3.ContainerResourceLimit
		oldLimit    *v3.ContainerResourceLimit
		wantAllowed bool
		wantErr     bool
	}{
		{
			name: "container default resource limit specified cpu and memory request and limits",
			newLimit: &v3.ContainerResourceLimit{
				RequestsCPU:    "1m",
				LimitsCPU:      "20m",
				RequestsMemory: "1Mi",
				LimitsMemory:   "20Mi",
			},
			wantAllowed: true,
		},
		{
			name:        "container default resource limit with no requests or limits",
			newLimit:    &v3.ContainerResourceLimit{},
			wantAllowed: true,
		},

```

Imo this would make the tests more maintainable

MbolotSuse: Personally I'm not a fan of separating a test for a single function into two separate functions. If the function under test is updated, you now have two separate tests to debug/fix, and its harder to see what the effective coverage for a single test is since you now have two test functions, each with their own methods of mocking/setting up the tests. 

For those reasons I would vote against this, though this is mostly a style concern so ultimately it's up to max.

maxsokolovsky: I agree with Tom and, in general, prefer to split tests into separate cases based on the aspects of behavior being tested. The quota setup may get in the way of setting up tests for limits. Create/update would have multiple separate test cases, but the cover tool would still show how many statements are covered in the package as pertains to create/update.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -143,6 +151,48 @@ func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (
...
+	}
+	if negativeResources := quota.IsNegative(resourceList); len(negativeResources) > 0 {
+		return field.Invalid(fieldPath, limit,
+			fmt.Sprintf("resource requests and limits can't have negative values, adjust the request/limit configurations for the following resources: %v", negativeResources))
+	}
+	if !requestsCPU.IsZero() && !limitCPU.IsZero() && requestsCPU.Cmp(limitCPU) > 0 {
+		fieldErr := field.Invalid(fieldPath, limit, fmt.Sprintf("requested CPU %s is greater than limit %s", limit.RequestsCPU, limit.LimitsCPU))
+		err = errors.Join(err, fieldErr)
+	}
+	if !requestsCPU.IsZero() && !limitCPU.IsZero() && requestsMemory.Cmp(limitMemory) > 0 {
```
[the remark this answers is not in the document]
maxsokolovsky: Good catch! The reason for all this is the difference between a literal "0" in limits or requests and absence of a value. Kubernetes treats a "0" as something restrictive. So I decided against setting a limit or request to "0" if the value in the request is an empty string. Using "0" here wouldn't matter because this is webhook, it only inspects a request, and nothing would literally set anything to "0", but I didn't want to operate with these misleading terms here.

Yet I am using this `IsZero` method, which could still be slightly misleading...

But the main check is needed here because it's OK to provide only request or only a limit. So we don't want to reject such configurations.

pkg/resources/management.cattle.io/v3/project/validator_test.go
MbolotSuse: > I agree with Tom and, in general, prefer to split tests into separate cases based on the aspects of behavior being tested. 

We've argued about this at length before. I think we will just need to agree to disagree here.

> but the cover tool would still show how many statements are covered in the package as it pertains to create/update.

The cover tool only indicates if a line of code was run, not if it is effectively tested. Here I'm speaking about coverage of the logic, not necessarily the code, which is not detectable by the coverage tool.

maxsokolovsky: The cover tool will show coverage of the logic (expressed in statements), just not of all possible values. Keeping the tests in a single case won't help with this either.

pkg/resources/management.cattle.io/v3/project/validator.go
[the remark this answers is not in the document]
maxsokolovsky: The tests didn't catch this because I mostly had both CPU and memory specified, I adjusted the tests to make them a bit more varied.

pkg/resources/management.cattle.io/v3/project/validator_test.go
tomleb: I can give my reasoning as to why I added the suggestion if it helps anybody. 

The current test already separates the tests into two categories: quotas and limits, but it's not made explicit. There aren't any test cases that covers both quotas & limits checks.

We might as well have something like:

```go
func TestProjectValidation(t *testing.T) {
  // Setup table for quota validation
  ...
  // Test quota
  ...
  // Setup table for limit validation
  ...
  // Test limit
  ...
}
```

and the test would be identical in the results, but each table could have specialized struct that makes it easier to review,add,modify,troubleshoot tests (IMO). **In my opinion, it's harder to find the tests that test limits to see if we have enough coverage when these tests can be randomly placed in a table full of other tests.**

We could separate the tests as mentioned above, or we could also separate into different functions. We could even test the method `validateContainerDefaultResourceLimit` at length, and instead test a few things only for the `Admit()` call.

I'm okay with whatever you choose to do though, just wanted to put my 2 cents here.

pkg/resources/management.cattle.io/v3/project/validator_test.go
maxsokolovsky: Made the test separate and set it up in a way that it runs all assertions over both create and update operations in a loop.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -143,6 +151,48 @@ func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (
...
+		v1.ResourceRequestsCPU:    requestsCPU,
+		v1.ResourceLimitsCPU:      limitCPU,
+		v1.ResourceRequestsMemory: requestsMemory,
+		v1.ResourceLimitsMemory:   limitMemory,
+	}
+	if negativeResources := quota.IsNegative(resourceList); len(negativeResources) > 0 {
+		return field.Invalid(fieldPath, limit,
+			fmt.Sprintf("resource requests and limits can't have negative values, adjust the request/limit configurations for the following resources: %v", negativeResources))
+	}
+	if !requestsCPU.IsZero() && !limitCPU.IsZero() && requestsCPU.Cmp(limitCPU) > 0 {
```
MbolotSuse, line 185: Unfortunately I don't think that this works. When I try to parse a 0 value on my local it returns a quantity that evaluates to true for this statement:

```golang
zeroStringM, err := resource.ParseQuantity("0m")
if err != nil {
	panic("error when parsing 0m string")
}
// prints 0m string is zero: true 
fmt.Printf("0m string is zero: %t \n", zeroStringM.IsZero())
```

This means that the following limit range would pass your webhook check (since the CPU limit is zero):
```yaml
apiVersion: v1
kind: LimitRange
metadata:
  name: test
spec:
  limits:
  - default:
      cpu: 0m
    defaultRequest:
      cpu: 1000m
    type: Container
```

However, when I try to create this in k8s, I get an error:
```
The LimitRange "test" is invalid: spec.limits[0].defaultRequest[cpu]: Invalid value: resource.Quantity{i:resource.int64Amount{value:1000, scale:-3}, d:resource.infDecAmount{Dec:(*inf.Dec)(nil)}, s:"1", Format:"DecimalSI"}: default request value 1 is greater than default limit value 0
```

I think that this will cause the original defect to present under certain circumstances (namely the use of a 0 quantity) and would recommend fixing this.

docs.md
```diff
@@ -192,6 +192,11 @@ The system project cannot be deleted.
 
 Project quotas and default limits must be consistent with one another and must be sufficient for the requirements of active namespaces.
 
+#### Container default resource limit validation
+
+Container default resource limit configuration must have no negative values for any requests and limits.
```
MbolotSuse, line 197: Nit: I would recommend trying to indicate that these limitations are sourced from upstream in some way. While you don't have to link to the docs, you could try and say something like "The container default limit configuration must have no negative values for any requests and limits, and the values must be able to be placed into a valid LimitRange". That way we are clear that we are mirroring limitations from upstream, rather than inventing our own.

pkg/resources/management.cattle.io/v3/project/validator_test.go
```diff
@@ -929,6 +930,417 @@ func TestProjectValidation(t *testing.T) {
...
+				Spec: v3.ProjectSpec{
+					ClusterName: "testcluster",
+					ContainerDefaultResourceLimit: &v3.ContainerResourceLimit{
+						RequestsCPU:    "1m",
+						LimitsCPU:      "20m",
+						RequestsMemory: "1Mi",
+						LimitsMemory:   "20Mi",
+					},
+				},
+			},
```
MbolotSuse, line 962: Nit:

One benefit of splitting out the tests was that you didn't need to define the whole project each time, you could instead just define the ContainerDefault resource limit, like the below:

```suggestion
			newLimit: &v3.ContainerResourceLimit{
						RequestsCPU:    "1m",
						LimitsCPU:      "20m",
						RequestsMemory: "1Mi",
						LimitsMemory:   "20Mi",
					},
```

pkg/resources/management.cattle.io/v3/project/validator_test.go
```diff
@@ -929,6 +930,417 @@ func TestProjectValidation(t *testing.T) {
...
+				}
+				if test.operation == admissionv1.Create {
+					state.clusterCache.EXPECT().Get("testcluster").Return(&v3.Cluster{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "testcluster",
+						},
+					}, nil)
+				}
+				if test.operation == admissionv1.Update {
+					test.oldProject = &v3.Project{
```
MbolotSuse, line 1311: Nit:
Since you don't actually specify the old project in your tests, you don't need to add the project as a member variable on the test. This has the additional benefit of just allowing you to have a var on the test called `test.Project` instead of only "new" project.

MbolotSuse (changes requested):

docs.md
maxsokolovsky: Agreed!

pkg/resources/management.cattle.io/v3/project/validator.go
maxsokolovsky: Good catch! I added those test cases and changed the logic, so that it's correct and also doesn't use the "zero" terms, which are ambiguous.

pkg/resources/management.cattle.io/v3/project/validator_test.go
maxsokolovsky: Yup, simplified.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -249,3 +302,15 @@ func formatResourceList(resources v1.ResourceList) string {
...
+
+func parseResource(s string) (resource.Quantity, error) {
+	if s == "" {
+		// Not a literal "0".
+		return resource.Quantity{}, nil
+	}
+	return resource.ParseQuantity(s)
+}
+
+func resourceSpecified(q resource.Quantity) bool {
```
MbolotSuse, line 314: I dislike this function since it assumes that `resource.ParseQuantity()` won't return a `resource.Quantity{}`. Given that we don't do any direct comparison of the values here (i.e. we are using `resource.Quantity.Cmp()`) I don't think that this is a safe call to make. Can you elaborate on why you don't return a `*resource.Quantity` and use a nil check instead?

MbolotSuse (commented):
One issue on the use of a default value and an unresolved comment around the docs, but outside of that mostly LGTM.

pkg/resources/management.cattle.io/v3/project/validator.go
maxsokolovsky: It does assume that `resource.ParseQuantity` will return a `resource.Quantity`.
My `parseResource` and `resourceSpecified` functions are meant to be used together.

`parseResource` is a wrapper around `resource.ParseQuantity()`. On its own, `resource.ParseQuantity()` returns an error when it gets an empty string. Thus we need that wrapper. And `resourceSpecified` simply compares a given quantity with its zero value, which is what I chose to use as a sign of a zero-value, unspecified quantity. What is unsafe here?

The reason for all this is the need to know if a value is invalid or just hasn't been specified.

MbolotSuse: > It does assume that `resource.ParseQuantity` will return a `resource.Quantity`

To be clear, I'm not talking about the type here. I'm referring to the fact that you are relying on the parse function not to return the default value, since you assume in your function that the default value means that the quantity was not specified. It's that specific behavior that I think is dangerous.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -143,6 +151,42 @@ func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (
...
+	}
+	requestsMemory, err := parseResource(limit.RequestsMemory)
+	if err != nil {
+		return field.Invalid(fieldPath, limit.RequestsMemory, fmt.Sprintf("failed to parse container default requested memory: %s", err))
+	}
+	limitsMemory, err := parseResource(limit.LimitsMemory)
+	if err != nil {
+		return field.Invalid(fieldPath, limit.LimitsMemory, fmt.Sprintf("failed to parse container default memory limit: %s", err))
+	}
+	if resources := negativeResources(requestsCPU, limitsCPU, requestsMemory, limitsMemory); len(resources) > 0 {
```
MbolotSuse, line 175: One alternative way of doing this would be to add a side function that adds a value to a list if not nil, and do the check here instead:
```suggestion
         resourceList := make(v1.ResourceList)
        addIfNotNil(resourceList, v1.ResourceRequestsCPU, requestsCPU)
        addIfNotNil(resourceList, v1.ResourceLimitsCPU, limitsCPU)
        addIfNotNil(resourceList, v1.ResourceRequestsMemory, requestsMemory)
        addIfNotNil(resourceList, v1.ResourceLimitsMemory, limitsMemory)
	if resources := quota.IsNegative(resourceList); len(resources) > 0 {
```

MbolotSuse (commented):
Some nits/quality suggestions.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -249,3 +293,40 @@ func formatResourceList(resources v1.ResourceList) string {
...
+	if requestsMemory == nil {
+		requestsMemory = &resource.Quantity{}
+	}
+	list[v1.ResourceRequestsMemory] = *requestsMemory
+
+	if limitsMemory == nil {
+		limitsMemory = &resource.Quantity{}
+	}
+	list[v1.ResourceLimitsMemory] = *limitsMemory
+	return quota.IsNegative(list)
```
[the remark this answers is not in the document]
maxsokolovsky: Good point, adjusted!

MbolotSuse (approved):

docs.md
```diff
@@ -192,6 +192,12 @@ The system project cannot be deleted.
 
 Project quotas and default limits must be consistent with one another and must be sufficient for the requirements of active namespaces.
 
+#### Container default resource limit validation
+
+Validation mimics the upstream behavior of the Kubernetes API server when it validates LimitRanges.
+The container default resource configuration must have properly formatted quantities, with no negative values
+for any requests and limits. Limits for any resource must not be less than requests.
```
tomleb, line 199: I'm able to create limitranges with negative CPU just fine.

```yaml
apiVersion: v1
kind: LimitRange
metadata:
  name: cpu-resource-constraint
spec:
  limits:
  - default: # this section defines default limits
      cpu: "-500m"
    type: Container
```

Applying works:

```
$ k apply -f limitrange.yaml 
limitrange/cpu-resource-constraint created
```

Then I can view it:

```sh
$ k get limitrange -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: LimitRange
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"v1","kind":"LimitRange","metadata":{"annotations":{},"name":"cpu-resource-constraint","namespace":"default"},"spec":{"limits":[{"default":{"cpu":"-500m"},"type":"Container"}]}}
    creationTimestamp: "2024-02-05T14:30:48Z"
    name: cpu-resource-constraint
    namespace: default
    resourceVersion: "1994455"
    uid: 79e4e545-4e3c-4bd1-8106-38c712ef8a2b
  spec:
    limits:
    - default:
        cpu: -500m
      defaultRequest:
        cpu: -500m
      type: Container
kind: List
metadata:
  resourceVersion: ""
```

I don't think there is a use for negative values though for CPU/memory so I think we should keep that check for better UI/UX, but the description seems wrong here that we're  mimicking the Kubernetes API validation.

tomleb: And now I'm wondering if someone is using project limits with negative values.. This fix would not break existing objects but it would break new objects with similar use case.

Difficult to know how impactful that would be.

CC: @MbolotSuse

tomleb (commented):
In general LGTM, I like the way the test looks now, great job :+1: 

Some comments about negative values, need to verify if we're okay with this.

docs.md
MbolotSuse: Good catch @tomleb. I think that we should not provide additional validation beyond what k8s provides. We don't want to make assumptions on what use cases are/are not relevant for end users. The description here was on purpose, we want to just make sure that we can create a valid limit range, the rest is up to the end user.

@maxsokolovsky can you review this and explain why validation was added for negative values?

maxsokolovsky: You're both correct.

```
apiVersion: v1
kind: LimitRange
metadata:
  name: cpu-resource-constraint
spec:
  limits:
  - defaultRequest:
      cpu: "-150"
    type: Container
```

The API server will accept this request. But when you create a regular pod without specifying resources and limits, you get an error.

```
$ k run nginx --image=nginx
The Pod "nginx" is invalid: spec.containers[0].resources.requests[cpu]: Invalid value: "-150": must be greater than or equal to 0
```

I originally thought that it'd be helpful for the user to alert about negative values early, as they aren't otherwise useful or meaningful in vanilla K8s LimitRange objects. But perhaps we should just stick to what the API server accepts now. And we could just rely on whatever internal errors from K8s might bubble up (if any).

maxsokolovsky: Interestingly, in my limited testing, I noticed that the API server rejects negative values on resource quotas.
For example:

```
apiVersion: v1
kind: ResourceQuota
metadata:
  name: pods-high
spec:
  hard:
    cpu: "-1000"
    memory: -200Gi
    pods: "-10"
```

Trying to submit the above manifest results in errors:
```
The ResourceQuota "pods-high" is invalid: 
* spec.hard[pods]: Invalid value: "-10": must be greater than or equal to 0
* spec.hard[cpu]: Invalid value: "-1k": must be greater than or equal to 0
* spec.hard[memory]: Invalid value: "-200Gi": must be greater than or equal to 0
```

tomleb (approved):

MbolotSuse (approved):

PR #332, opened 2024-02-15
Validate RoleTemplate as non-namespaced

tomleb:
## Issue

This should fix the integration tests in rancher/rancher issue that was introduced in https://github.com/rancher/webhook/pull/328 (see also https://github.com/rancher/rancher/issues/40584) which results in the following tests to fail:

```
test_default_roles.py::test_cluster_create_role_locked
test_default_roles.py::test_project_create_default_role
test_default_roles.py::test_project_create_role_locked
test_default_roles.py::test_user_create_default_role
```

The error looks like the following:

```
"rancher.cattle.io.roletemplates.management.cattle.io" denied the request: roletemplate.rules[8].nonResourceURLs: Invalid value: []string{"*"}: namespaced rules cannot apply to non-resource URLs\n\t{\'baseType\': \'error\', \'code\': \'BadRequest\', \'message\': \'admission webhook "rancher.cattle.io.roletemplates.management.cattle.io" denied the request: roletemplate.rules[8].nonResourceURLs: Invalid value: []string{"*"}: namespaced rules cannot apply to non-resource URLs\', \'status\': 400, \'type\': \'error\'}')
```

The issue is that namespaced resources cannot have `nonResourceURLs` in them. The `true` here is passed down to the upstream k8s validation function, which sees the nonResourceURLs in the `cluster-owner` RoleTemplate.

The fix simply uses the rules from non-namespaced resources, which skips the check for nonResourceURLs. Note this is not perfectly 1:1 with k8s but it's still more checks than we had previously.

@andreas-kupries As mentioned in Slack, feel free to take this as inspiration for a fix, or review it as is. Leaving as draft for now.

### DIFF
--- a/pkg/resources/management.cattle.io/v3/roletemplate/main_test.go
+++ b/pkg/resources/management.cattle.io/v3/roletemplate/main_test.go
@@ -65,6 +65,10 @@ func (c *RoleTemplateSuite) SetupSuite() {
 		APIGroups: []string{"*"},
 		Resources: []string{"*"},
 	}
+	ruleAdminNonResource := rbacv1.PolicyRule{
+		Verbs:           []string{"*"},
+		NonResourceURLs: []string{"*"},
+	}
 	c.ruleEmptyVerbs = rbacv1.PolicyRule{
 		Verbs:     nil,
 		APIGroups: []string{"v1"},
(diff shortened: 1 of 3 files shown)


andreas-kupries (approved):
Looking over it, and reading the explanations, this looks to me like a good fix.
And apologies for missing this namespace vs not) when writing the original PR.

Checked and confirmed for myself that `RoleTemplates` are not namespaced.
Cross-checking `GlobalRoles` are also not namespaced  and their calls to ValidateRules are very much :ok: , using `false` outright.

This confirms to me that the change to use `false` is the correct fix.
:heavy_check_mark: `make test` passes.

maxsokolovsky (approved):

PR #334, opened 2024-02-19
Ensure that Rules in RoleTemplates are encoded

raulcabello:
## Problem
If a `RoleTemplate` is created with an empty list in any `Rule`, then rancher creates a corresponding `ClusterRole` with a nil list. When Rancher goes to sync `RoleTemplates` and `ClusterRoles` it will detect that emptyList != nil list and attempt to update the `ClusterRole`.

The fields of `PolicyRule` have the `omitempty` JSON tag, indicating that they will be excluded if they are empty when encoded to JSON. It appears that Kubernetes encodes to JSON before storing `Roles` or `ClusterRoles` in etcd, resulting in the removal of `resourceNames` and `nonResourceURLs` if they are empty. Rancher doesn't do that, and that's why we see the difference.

## Solution
New mutating webhook that applies a patch to ensure that the `RoleTemplate` is encoded when the encoded rules differ from the non-encoded rules.  

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -291,6 +291,13 @@ If `roletemplates.builtin` is true then all fields are immutable except:
 
 RoleTemplate can not be deleted if they are referenced by other RoleTemplates via `roletemplates.roleTemplateNames` or by GlobalRoles via `globalRoles.inheritedClusterRoles`
 
+### Mutation Checks
+
+#### Ensure Rules are encoded
+
+The fields of `PolicyRule` have the `omitempty` JSON tag, indicating that they will be excluded if they are empty when encoded to JSON.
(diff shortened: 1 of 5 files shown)


tomleb:
Have you thought of using `k8s.io/apimachinery/pkg/api/equality.Semantic.DeepEqual` as a replacement of `reflect.DeepEqual` where the RBAC controllers are comparing the RT and the ClusterRole? That function will consider a `nil` list and an empty list to be equal, which I think would solve the problem without needing a mutating webhook.

raulcabello:
Thanks for pointing that out @tomleb ! I wasn't aware of this function. That is a much better approach. I'll close this PR

raulcabello:
superseded by https://github.com/rancher/rancher/pull/44519

PR #335, opened 2024-02-28
[v0.4] Add support for k8s 1.28

tomleb:
## Issue: https://github.com/rancher/rancher/issues/44018

Depends on https://github.com/rancher/rancher/issues/44619

## Problem

We want to support 1.28 k8s.

## Solution

Bump relevant dependencies. cluster-api v1.5.5 conflicts with k8s.io/* 0.28. One option is to remove it entirely from webhook since we aren't using it anymore (since v2.7.7).

The changes are:
- Upgraded cluster-api to v1.5.5 to match rancher/rancher v2.8
- Upgraded controller-runtime to v0.15.3
- Upgraded github.com/rancher/rke to v1.5.7-rc2
- Upgraded lasso which supports 1.28
- Upgraded to wrangler v2 so that's why there is a LOT of files modified.
- Upgraded dynamiclistener to v0.4.0-rc2 because it contains 1.28 support + wrangler v2. Otherwise conflicts, can't compile. Here's the comparison of the previous version vs this one: https://github.com/rancher/dynamiclistener/compare/v0.3.5...v0.4.0-rc2. Some non-trivial changes here, but it seemed okay to do in rancher/rancher so should be safe here.
- Upgrade the k8s dependencies to v0.28.6

## CheckList

- [ x] Tested a subset of integration tests from rancher/rancher


go.mod
```diff
@@ -2,148 +2,126 @@ module github.com/rancher/webhook
...
+	k8s.io/kube-proxy => k8s.io/kube-proxy v0.28.6
+	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.28.6
+	k8s.io/kubectl => k8s.io/kubectl v0.28.6
+	k8s.io/kubelet => k8s.io/kubelet v0.28.6
+	k8s.io/kubernetes => k8s.io/kubernetes v1.28.6
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.28.6
+	k8s.io/metrics => k8s.io/metrics v0.28.6
+	k8s.io/mount-utils => k8s.io/mount-utils v0.28.6
+	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.28.6
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.28.6
```
tomleb, line 31: Anybody knows why we're duplicating most of these entries in `require` and `replace`?

Some of them aren't even imported directly nor indirectly (eg: `k8s.io/cli-runtime`).

If there's no good reason then I can nuke this and only keep `replace` for things that require replace like the client-go fork.

JonCrowther (approved):

ericpromislow (approved):
Tests pass, code changes make sense.

PR #337, opened 2024-03-05
Add validation to not admit RoleTemplates with both context=cluster & ProjectCreatorDefault=true

vatsalparekh:
## Issue: 

Add validation to not admit RoleTemplates with both context=cluster & ProjectCreatorDefault=true

## Problem

SURE-7635

## Solution
Add a check for this field in validation

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/tests/integration/project_test.go
+++ b/tests/integration/project_test.go
@@ -101,6 +101,7 @@ func (m *IntegrationSuite) TestMutateProject() {
 			GenerateName: "test-project-creator-",
 		},
 		ProjectCreatorDefault: true,
+		Context:               "project",
 	}
 	err = rtClient.Create(ctx, "", rt, rt, metav1.CreateOptions{})
 	m.Require().NoError(err)


samjustus:
https://github.com/rancher/rancher/issues/44860

tests/integration/roleTemplate_test.go
```diff
@@ -118,3 +118,35 @@ func (m *IntegrationSuite) TestRoleTemplateNoAPIGroups() {
 	}
 	validateEndpoints(m.T(), endPoints, m.clientFactory)
 }
+
+func (m *IntegrationSuite) TestRoleTemplateInvalidContext() {
+	newObj := func() *v3.RoleTemplate { return &v3.RoleTemplate{} }
```
bigkevmcd, line 123: This test can maybe be named a bit better?

Someone reading it would need to figure out that the only significant thing in this test is the `invalidCreate.ProjectCreatorDefault` line?

Also, maybe we could improve the validation a bit?

bigkevmcd (commented):
While I think the actual fix is fine, I think we can improve the tests to clarify that they're testing.

tests/integration/roleTemplate_test.go
vatsalparekh: > maybe we could improve the validation a bit?

can you elaborate how

pkg/resources/management.cattle.io/v3/roletemplate/validator.go
```diff
@@ -192,6 +192,9 @@ func validateCreateFields(newRole *v3.RoleTemplate, fldPath *field.Path) *field.
 }
 
 func validateContextValue(newRole *v3.RoleTemplate, fldPath *field.Path) *field.Error {
+	if newRole.Context == clusterContext && newRole.ProjectCreatorDefault == true {
```
JonCrowther, line 195: In the docs you stated

> If the `roletemplate.ProjectCreatorDefault` is true, context must equal `"project"`

But this checks only that the context is `"cluster"`. Context can be `"cluster"` `"project"` or `""` ([source](https://github.com/rancher/rancher/blob/c5b4413739fb7a9f935eadeb7723b28b90ee22be/pkg/apis/management.cattle.io/v3/authz_types.go#L280)), so this check should probably be:
```suggestion
	if newRole.Context != projectContext && newRole.ProjectCreatorDefault == true {
```

Alternatively, the doc should say 

> If the `roletemplate.ProjectCreatorDefault` is true, context can not be `"cluster"`

tests/integration/roleTemplate_test.go
```diff
@@ -118,3 +121,32 @@ func (m *IntegrationSuite) TestRoleTemplateNoAPIGroups() {
...
+	m.Require().NoError(err, "Failed to get gvk")
+	client, err := m.clientFactory.ForKind(gvk)
+	m.Require().NoError(err, "Failed to create client")
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	err = client.Create(ctx, invalidRoleTemplate.GetNamespace(), invalidRoleTemplate, nil, v1.CreateOptions{})
+	m.Assert().NotNil(err)
+	m.Assert().Contains(err.Error(), "RoleTemplate context must be Project when projectCreatorDefault=true")
```
bigkevmcd, line 151: I don't think this error is right?
```suggestion
	m.Assert().Contains(err.Error(), "RoleTemplate context must be project when projectCreatorDefault=true")
```

tests/integration/roleTemplate_test.go
```diff
@@ -118,3 +121,32 @@ func (m *IntegrationSuite) TestRoleTemplateNoAPIGroups() {
...
+	gvk, err := m.clientFactory.GVKForObject(invalidRoleTemplate)
+	m.Require().NoError(err, "Failed to get gvk")
+	client, err := m.clientFactory.ForKind(gvk)
+	m.Require().NoError(err, "Failed to create client")
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	err = client.Create(ctx, invalidRoleTemplate.GetNamespace(), invalidRoleTemplate, nil, v1.CreateOptions{})
+	m.Assert().NotNil(err)
```
bigkevmcd, line 150: You don't need this check, because the test will fail if it's not `nil`.

tests/integration/roleTemplate_test.go
vatsalparekh: Fixed

JonCrowther (approved):

bigkevmcd (approved):
Looks good, thank you!

PR #340, opened 2024-03-11
Backport of PRs #328 and #332, plus go module refresh

andreas-kupries:
## Issue:

Ticket https://github.com/rancher/rancher/issues/44356 to backport the solution for https://github.com/rancher/rancher/issues/40584

Original solution and a fix are PRs
  - #328, and
  - #332 

## Problem

See https://github.com/rancher/rancher/issues/40584

## Solution

See #328 and #332.

:warning: Checking the `NamespacedRules` was removed from the backport. This field is not present in the definition of `GlobalRole` used by the 0.4.x release series of webhook.

## CheckList

- [x] Test
- [x] Docs


pkg/resources/common/validation_test.go
```diff
@@ -0,0 +1,85 @@
...
+		})
+		t.Run("namespaced/"+testcase.name, func(t *testing.T) {
+			err := ValidateRules([]v1.PolicyRule{
+				testcase.data,
+			}, true, nsField)
+			if testcase.haserr {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
```
pmatseykanets, line 82: Q: Should this stay given that the `NamespacedRules` is not yet available in v2.8 release line?

andreas-kupries: This has nothing to do with the `NamespacedRules` field not present in 2.8.

These are the unit tests for `ValidateRules`. This function which can check global and namespaced structures, as per its `isNamespaced` argument. It is my belief that the unit tests should cover both cases, despite the function only used with `isNamespaced=false` in this release line.

pmatseykanets (approved):

JonCrowther (approved):

tomleb:
I think we should wait before merging this since we're in code freeze for 2.8

samjustus:
https://github.com/rancher/rancher/issues/44356

PR #342, opened 2024-03-19
Change codeowner to new team

JonCrowther:
## Problem
The codeowner file references a group that was recently deleted and is no longer valid

## Solution
Updated the codeowner to the new team name


tomleb (approved):
For those that want to verify the team name you can type @ followed by rancher/ and keep writing until you see the full team name.

nflynt (approved):

PR #345, opened 2024-03-25
[v0.4] Upgrade golangci-lint and fix warnings and errors associated with the new version

JonCrowther:
## Problem
Renovate bot is attempting to update `golangci-lint`, but in the newer version there are checks that fail. Specifically, it no longer allows variables in functions that go unused, instead requiring they be removed or set to `_`. Additionally, `run.skip-files` and `run.skip-dirs` are deprecated for `golangci-lint`.

## Solution
Ran `golangci-lint` manually and fixed up all the errors. Used `issues.exclude-dirs` and `issues.exclude-files`, which are the newer, supported method of ignoring files/dirs.

### DIFF
--- a/.golangci.json
+++ b/.golangci.json
@@ -22,9 +22,6 @@
 		],
 		"exclude-dirs": [
 			"pkg/generated"
-		],
-		"exclude-files": [
-			"//zz_generated_"
 		]
 	}
 }
--- a/.golangci.json
+++ b/.golangci.json
@@ -11,12 +11,6 @@
 		]
 	},
 	"run": {
-		"skip-dirs": [
-			"pkg/generated"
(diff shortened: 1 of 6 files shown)


.golangci.json
```diff
@@ -25,6 +19,12 @@
 				"linters": "revive",
 				"text": "should have comment or be unexported"
 			}
+		],
+		"exclude-dirs": [
+			"pkg/generated"
+		],
+		"exclude-files": [
+			"//zz_generated_"
```
tomleb, line 27: Do we need this? I can't find such files in webhook

tomleb (approved):

MbolotSuse (approved):

PR #366, opened 2024-05-03
Bump to k8s 1.29

ericpromislow:
Related to: https://github.com/rancher/rancher/issues/45351


pkg/resources/management.cattle.io/v3/cluster/validator.go
```diff
@@ -266,23 +262,3 @@ func (a *admitter) checkPSAConfigOnCluster(cluster *apisv3.Cluster) (*admissionv
 	}
 	return admission.ResponseAllowed(), nil
 }
-
-// validatePSP validates if the PSP feature is enabled in a cluster which version is 1.25 or above.
-func (a *admitter) validatePSP(request *admission.Request) (*admissionv1.AdmissionResponse, error) {
```
nflynt, line 271: It looks like the webhook is still set to permit itself to run on k8s 1.23.0-0 as a minimum version. Do we forsee problems with the removal of this logic without bumping the webhook's required minimum to 1.25, where we can be more certain that the PSP objects won't still be functional on the cluster where the webhook is run?

ericpromislow: I don't know enough about this. If this version of webhook runs against an API that supports PSPs, but it never mentions them, is that a problem?

Also looking at various charts, I see the min. version varies all over the place. This is a trivial change to make; I'm just not sure it's correct.

nflynt: Per discussion with the frameworks team, it is a bit odd for webhook to specify a minimum k8s version while Rancher does not. The thinking is that, should a user _actually_ try to upgrade Rancher with a particularly old downstream or something, the webhook would fail to install entirely, which is a much bigger issue than it merely skipping PSP checks.

For now, let's remove the minimum version, and have the k8s version requirements for webhook match rancher/rancher, which also specifies no minimum version. We should be mentioning the PSP check removal prominently in the release notes, and from there, it will be up to end users to administer their clusters according to that guidance. (We can separately discuss whether rancher/rancher _should_ specify a minimum version, but let's not block the webhook PR on that.)

ericpromislow: I removed the minimum version.

nflynt (approved):

ericpromislow:
`.drone.yml` was updated to `curl` down helm instead of relying on presence of `zypper`, so no need to base drone on an older rancher image.

nflynt (approved):

maxsokolovsky (approved):

go.mod
```diff
@@ -1,171 +1,165 @@
 module github.com/rancher/webhook
 
-go 1.22
+go 1.22.0
+
+toolchain go1.22.1
```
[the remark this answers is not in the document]
tomleb: As discussed on Slack we'll merge as is and then revisit in a future PR that should follow pretty quickly.

PR #368, opened 2024-05-06
Dynamic schema rfc

jakefhyde:
## Problem

An issue was identified (https://github.com/rancher/rancher/issues/40609), where updating the node driver and changing the flags would result in CAPI detecting the supplied fields do not match the existing fields on the machine infrastructure custom resource. The fix for this issue was to persist the dynamicSchema onto the cluster object under the machinePools section, and stripping fields not present at the time of initial provisioning from the related CR, as the alternative of properly versioning CRDs for each node driver was not and continues to not be a viable solution. While this change initially fixed the issue of unintended cluster reprovisioning on node driver update, other cluster update solutions such as terraform and kubectl are still susceptible to this issue as an update unintentionally removing the dynamicSchemaSpec fields will still result in reprovisioning if new fields (whether used or not) are added to a node driver. Since the dynamicSchemaSpec is added by a controller within Rancher, a cluster create request followed by updating the node driver and a subsequent update request (identical to the cluster create request) will cause reprovisioning, due to the dynamicSchemaSpec being unpopulated within the request.

## Solution

Add an annotation which the webhook will use to determine whether the removal of the dynamicSchemaSpec is intentional. If not present or ”false”, the dynamicSchemaSpec will be reinserted into the cluster object as-is, preventing an unintentional cluster-wide machine rollout. If true, the fields will stay removed from the request, and allow provisioning with the previous behavior, allowing users to opt-in for a potential rollout if the node driver was indeed changed.

## CheckList
  
- [x] Test
  
- [ ] Docs

There isn't currently an existing markdown file for provisioning clusters. I'm happy to create one in a follow up PR if necessary.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,8 @@ replace (
 	k8s.io/controller-manager => k8s.io/controller-manager v0.30.1
 	k8s.io/cri-api => k8s.io/cri-api v0.30.1
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.30.1
+	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.30.1
+	k8s.io/endpointslice => k8s.io/endpointslice v0.30.1
 	k8s.io/kube-aggregator => k8s.io/kube-aggregator v0.30.1
 	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.30.1
 	k8s.io/kube-proxy => k8s.io/kube-proxy v0.30.1
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
@@ -103,7 +103,7 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 	listTrace := trace.New("provisioningCluster Admit", trace.Field{Key: "user", Value: request.UserInfo.Username})
 	defer listTrace.LogIfLong(admission.SlowTraceDuration)
 
-	cluster, err := objectsv1.ClusterFromRequest(&request.AdmissionRequest)
+	cluster, oldCluster, err := objectsv1.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
 	if err != nil {
 		return nil, err
 	}
@@ -134,12 +134,14 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 		return response, nil
 	}
 
-	response, err = m.handleDynamicSchemaDrop(request, cluster)
-	if err != nil {
-		return nil, err
-	}
-	if response.Result != nil {
-		return response, nil
+	if request.Operation == admissionv1.Update {
+		response, err = m.handleDynamicSchemaDrop(request, oldCluster, cluster)
+		if err != nil {
+			return nil, fmt.Errorf("unable to evaluate dynamic schema drop, %w", err)
+		}
+		if response.Result != nil {
+			return response, nil
+		}
 	}
 
 	response.Allowed = true
@@ -152,21 +154,12 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 // handleDynamicSchemaDrop watches for provisioning cluster updates, and reinserts the previous value of the
 // dynamicSchemaSpec field for a machine pool if the "provisioning.cattle.io/allow-dynamic-schema-drop" annotation is
 // not present and true on the cluster. If the value of the annotation is true, no mutation is performed.
-func (m *ProvisioningClusterMutator) handleDynamicSchemaDrop(request *admission.Request, cluster *v1.Cluster) (*admissionv1.AdmissionResponse, error) {
+func (m *ProvisioningClusterMutator) handleDynamicSchemaDrop(request *admission.Request, oldCluster, cluster *v1.Cluster) (*admissionv1.AdmissionResponse, error) {
 	if cluster.Name == "local" || cluster.Spec.RKEConfig == nil {
 		return admission.ResponseAllowed(), nil
 	}
 
-	if request.Operation != admissionv1.Update {
-		return admission.ResponseAllowed(), nil
-	}
-
-	oldCluster, newCluster, err := objectsv1.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
-	if err != nil {
-		return nil, fmt.Errorf("failed to extract old and new cluster objects: %w", err)
-	}
-
-	if newCluster.Annotations[allowDynamicSchemaDropAnnotation] == "true" {
+	if cluster.Annotations[allowDynamicSchemaDropAnnotation] == "true" {
 		return admission.ResponseAllowed(), nil
 	}
 
@@ -175,20 +168,15 @@ func (m *ProvisioningClusterMutator) handleDynamicSchemaDrop(request *admission.
 		oldClusterPools[mp.Name] = &mp
 	}
 
-	newClusterPools := map[string]*v1.RKEMachinePool{}
-	for _, mp := range newCluster.Spec.RKEConfig.MachinePools {
-		newClusterPools[mp.Name] = &mp
-	}
-
-	for name, newPool := range newClusterPools {
-		oldPool, ok := oldClusterPools[name]
+	for i, newPool := range cluster.Spec.RKEConfig.MachinePools {
+		oldPool, ok := oldClusterPools[newPool.Name]
 		if !ok {
-			logrus.Debugf("new machine pool, skipping validation of dynamic schema spec")
+			logrus.Debugf("[%s] new machine pool: %s, skipping validation of dynamic schema spec", request.UID, newPool.Name)
 			continue
 		}
 		if oldPool.DynamicSchemaSpec != "" && newPool.DynamicSchemaSpec == "" {
-			logrus.Infof("provisioning cluster %s/%s machine pool %s dynamic schema spec mutated without supplying annotation %s, reverting", cluster.Namespace, cluster.Name, name, allowDynamicSchemaDropAnnotation)
-			newPool.DynamicSchemaSpec = oldPool.DynamicSchemaSpec
+			logrus.Infof("provisioning cluster %s/%s machine pool %s dynamic schema spec mutated without supplying annotation %s, reverting", cluster.Namespace, cluster.Name, newPool.Name, allowDynamicSchemaDropAnnotation)
+			cluster.Spec.RKEConfig.MachinePools[i].DynamicSchemaSpec = oldPool.DynamicSchemaSpec
 		}
 	}
 	return admission.ResponseAllowed(), nil
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
@@ -103,7 +103,7 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 	listTrace := trace.New("provisioningCluster Admit", trace.Field{Key: "user", Value: request.UserInfo.Username})
 	defer listTrace.LogIfLong(admission.SlowTraceDuration)
 
-	cluster, oldCluster, err := objectsv1.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
+	oldCluster, cluster, err := objectsv1.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
 	if err != nil {
 		return nil, err
 	}
@@ -135,10 +135,7 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 	}
 
 	if request.Operation == admissionv1.Update {
-		response, err = m.handleDynamicSchemaDrop(request, oldCluster, cluster)
-		if err != nil {
-			return nil, fmt.Errorf("unable to evaluate dynamic schema drop, %w", err)
-		}
+		response = m.handleDynamicSchemaDrop(request, oldCluster, cluster)
 		if response.Result != nil {
 			return response, nil
 		}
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -18,6 +18,8 @@ replace (
 	k8s.io/controller-manager => k8s.io/controller-manager v0.29.3
 	k8s.io/cri-api => k8s.io/cri-api v0.29.3
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.29.3
+	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.29.3
+	k8s.io/endpointslice => k8s.io/endpointslice v0.29.3
```
jakefhyde, line 22: Intellisense fails without doing this.

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -133,13 +134,66 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
...
 		return nil, fmt.Errorf("failed to create patch: %w", err)
 	}
 	return response, nil
 }
 
+// handleDynamicSchemaDrop watches for provisioning cluster updates, and reinserts the previous value of the
+// dynamicSchemaSpec field for a machine pool if the "provisioning.cattle.io/allow-dynamic-schema-drop" annotation is
+// not present and true on the cluster. If the value of the annotation is true, no mutation is performed.
+func (m *ProvisioningClusterMutator) handleDynamicSchemaDrop(request *admission.Request, cluster *v1.Cluster) (*admissionv1.AdmissionResponse, error) {
+	if cluster.Name == "local" || cluster.Spec.RKEConfig == nil {
```
Oats87, line 156: we should make sure the harvester team isn't somehow relying on dynamic schema for the local cluster in rancherd

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -133,13 +134,66 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
...
+	}
+
+	for name, newPool := range newClusterPools {
+		oldPool, ok := oldClusterPools[name]
+		if !ok {
+			logrus.Debugf("new machine pool, skipping validation of dynamic schema spec")
+			continue
+		}
+		if oldPool.DynamicSchemaSpec != "" && newPool.DynamicSchemaSpec == "" {
+			logrus.Infof("cluster %s machine pool %s dynamic schema spec mutated without supplying annotation %s, reverting", cluster.Name, name, "provisioning.cattle.io/allow-dynamic-schema-drop")
```
Oats87, line 190: nit: for posterity we should probably include the namespace of the cluster in question i.e. `provisioning cluster %s/%s`

Oats87 (approved):

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
jakefhyde: Updated

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
ibrokethecloud: i don't think we are leveraging dynamic schema on local cluster for anything during the harvester bootstrap / lifecycle management. @bk201 @Vicente-Cheng

jakefhyde: Sounds good then, as long as harvester isn't worried about the dynamic schema changing with new flags and causing reprovisioning for the local cluster.

bk201: @FrankYang0529 Please help take a look too.

Oats87 (approved):

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
FrankYang0529: We don't use `MachinePools` field in `RKEConfig`, but we use `RotateCertificates`. Not sure whether it's related? Thanks.

Oats87 (dismissed):

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -133,13 +134,66 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
...
+
+	newClusterPools := map[string]*v1.RKEMachinePool{}
+	for _, mp := range newCluster.Spec.RKEConfig.MachinePools {
+		newClusterPools[mp.Name] = &mp
+	}
+
+	for name, newPool := range newClusterPools {
+		oldPool, ok := oldClusterPools[name]
+		if !ok {
+			logrus.Debugf("new machine pool, skipping validation of dynamic schema spec")
```
MbolotSuse, line 186: Nit: This should probably include more detail that allows us to correlate which request/object this log statement is for. Something like:

```suggestion
			logrus.Debugf("[%s] new machine pool: %s, skipping validation of dynamic schema spec", request.UID, name)
```

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -133,13 +134,66 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
...
+			continue
+		}
+		if oldPool.DynamicSchemaSpec != "" && newPool.DynamicSchemaSpec == "" {
+			logrus.Infof("provisioning cluster %s/%s machine pool %s dynamic schema spec mutated without supplying annotation %s, reverting", cluster.Namespace, cluster.Name, name, allowDynamicSchemaDropAnnotation)
+			newPool.DynamicSchemaSpec = oldPool.DynamicSchemaSpec
```
MbolotSuse, line 191: I'm not quite sure that this works. From what I can see, this is updating the value in-memory on the newCluster object, which was re-parsed from the request. However, we are re-marshalling/adding the original cluster object which was passed in to the function. 

In short, I'm not sure that this is editing the original cluster object like this function needs. Can you double check that and let me know what you find?

pkg/resources/provisioning.cattle.io/v1/cluster/mutator_test.go
```diff
@@ -617,3 +618,96 @@ func TestAdmitPreserveUnknownFields(t *testing.T) {
...
+	raw, err = json.Marshal(cluster)
+	assert.Nil(t, err)
+	request.AdmissionRequest.Object.Raw = raw
+	output = cluster.DeepCopy()
+	response, err = m.handleDynamicSchemaDrop(request, output)
+	assert.Nil(t, err)
+	assert.True(t, response.Allowed)
+	assert.True(t, reflect.DeepEqual(cluster, output))
+
+	// Rejected on update if schema is removed
```
MbolotSuse, line 689: Nit: This comment is a bit inaccurate - the request is always allowed, but we mutate it under select circumstances.
```suggestion
	// Allowed on updated if schema is removed, but old SchemaSpec is re-introduced.
```

pkg/resources/provisioning.cattle.io/v1/cluster/mutator_test.go
```diff
@@ -617,3 +618,96 @@ func TestAdmitPreserveUnknownFields(t *testing.T) {
...
+	raw, err = json.Marshal(cluster)
+	assert.Nil(t, err)
+	request.AdmissionRequest.Object.Raw = raw
+	output = cluster.DeepCopy()
+	response, err = m.handleDynamicSchemaDrop(request, output)
+	assert.Nil(t, err)
+	assert.True(t, response.Allowed)
+	expected := cluster.DeepCopy()
+	expected.Spec.RKEConfig.MachinePools[0].DynamicSchemaSpec = "a"
+	assert.True(t, reflect.DeepEqual(cluster, output))
```
MbolotSuse, line 700: Should this be checking expected instead?

pkg/resources/provisioning.cattle.io/v1/cluster/mutator_test.go
```diff
@@ -617,3 +618,96 @@ func TestAdmitPreserveUnknownFields(t *testing.T) {
 	assert.Nil(t, response.Patch)
 
 }
+
+func TestDynamicSchemaDrop(t *testing.T) {
```
MbolotSuse, line 622: A few test cases that would be good to add:
- Error getting old/new objects from the request (can be hard to mock, don't necessarily need this)
- Annotation is set, but not to true (i.e. annotation is set to false).

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -133,13 +134,66 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 		return response, nil
 	}
 
+	response, err = m.handleDynamicSchemaDrop(request, cluster)
+	if err != nil {
+		return nil, err
```
MbolotSuse, line 139: Nit: would be good to wrap this so that we know which sub-function caused the error:

```suggestion
		return nil, fmt.Errorf("unable to evaluate dynamic schema drop, %w", err)
```

MbolotSuse (changes requested):
One significant bug, and a few nits.

pkg/resources/provisioning.cattle.io/v1/cluster/mutator_test.go
jakefhyde: Added a test case for the latter, AFAIK for the former there should never be an empty `OldObject` on update. I refactored it a bit to do get the old object one level higher.

jakefhyde: Also refactored the test function to be easier to follow and be more structured per test case.

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -102,7 +103,7 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 	listTrace := trace.New("provisioningCluster Admit", trace.Field{Key: "user", Value: request.UserInfo.Username})
 	defer listTrace.LogIfLong(admission.SlowTraceDuration)
 
-	cluster, err := objectsv1.ClusterFromRequest(&request.AdmissionRequest)
+	cluster, oldCluster, err := objectsv1.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
```
MbolotSuse, line 106: Based on the docs of this function and the [return values](https://github.com/rancher/webhook/blob/97679b88dd800c1cbad320721d721e4250fa7a91/pkg/generated/objects/provisioning.cattle.io/v1/objects.go#L38), I think the order is old first, then new. So this should be switched around:

```suggestion
	oldCluster, cluster, err := objectsv1.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
```

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -133,13 +134,54 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
...
+	}
+
+	for i, newPool := range cluster.Spec.RKEConfig.MachinePools {
+		oldPool, ok := oldClusterPools[newPool.Name]
+		if !ok {
+			logrus.Debugf("[%s] new machine pool: %s, skipping validation of dynamic schema spec", request.UID, newPool.Name)
+			continue
+		}
+		if oldPool.DynamicSchemaSpec != "" && newPool.DynamicSchemaSpec == "" {
+			logrus.Infof("provisioning cluster %s/%s machine pool %s dynamic schema spec mutated without supplying annotation %s, reverting", cluster.Namespace, cluster.Name, newPool.Name, allowDynamicSchemaDropAnnotation)
```
MbolotSuse, line 178: Would you mind shifting this down to debug? Running the test on my local, info seems to be the default level, and it occurs to me that this could flood the logs in certain environments in the current format.

MbolotSuse (changes requested):
2 small comments - one on an incorrect order of old/new object, one on log level.

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -133,13 +134,54 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
...
 	if err = patch.CreatePatch(clusterJSON, cluster, response); err != nil {
 		return nil, fmt.Errorf("failed to create patch: %w", err)
 	}
 	return response, nil
 }
 
+// handleDynamicSchemaDrop watches for provisioning cluster updates, and reinserts the previous value of the
+// dynamicSchemaSpec field for a machine pool if the "provisioning.cattle.io/allow-dynamic-schema-drop" annotation is
+// not present and true on the cluster. If the value of the annotation is true, no mutation is performed.
+func (m *ProvisioningClusterMutator) handleDynamicSchemaDrop(request *admission.Request, oldCluster, cluster *v1.Cluster) (*admissionv1.AdmissionResponse, error) {
```
MbolotSuse, line 157: Now that we removed the old and new extraction, I think that this doesn't need to return an error any  more. WDYT?

pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
jakefhyde: @FrankYang0529 Yeah, not realted to any of the day 2 ops, only when new fields are added to a driver.

pkg/resources/provisioning.cattle.io/v1/cluster/mutator_test.go
```diff
@@ -617,3 +621,275 @@ func TestAdmitPreserveUnknownFields(t *testing.T) {
...
+				{
+					Name: "a",
+				},
+			},
+		},
+		{
+			name:    "new machine pool without schema",
+			request: &admission.Request{AdmissionRequest: admissionv1.AdmissionRequest{Operation: admissionv1.Update}},
+			cluster: &v1.Cluster{
+				ObjectMeta: metav1.ObjectMeta{Annotations: map[string]string{"provisioning.cattle.io/allow-dynamic-schema-drop": "true"}},
```
MbolotSuse, line 844: Since this has the annotation allowing drop, it never gets to the statement that allows a bypass on the new machine pool. I would recommend removing the annotation so that this case is also covered.

MbolotSuse (approved):
LGTM, please squash when merging.

Oats87 (approved):

PR #383, opened 2024-05-23
Migrate CI from drone to GHA

ericpromislow:
Related to https://github.com/rancher/rancher/issues/45231

Actual PR at https://github.com/rancher/webhook/pull/394 (in order to access secrets for the publish and release actions)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,134 @@
...
+    strategy:
+      matrix:
+          arch:
+          - amd64
+          - arm64
+    steps:
+      - name : Checkout repository
+        uses : actions/checkout@v4
+        with :
+          fetch-depth : 0
```
tomleb, line 38: Do we need to fetch the whole history from rancher/webhook? If not then I suggest sticking with the default.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,134 @@
...
+    strategy:
+      matrix:
+        arch:
+        - amd64
+        - arm64
+    steps:
+      - name : Checkout repository
+        uses : actions/checkout@v4
+        with :
+          fetch-depth : 0
```
tomleb, line 67: Do we need to fetch the whole history from rancher/webhook? If not then I suggest sticking with the default.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+      # TODO: Pull this next one out once there's a release of rancher 2.9 that runs on k8s 1.29
+      - name : Checkout rancher/rancher and build the chart
+        run: |
+          pushd "${{ runner.temp}}"
+          git clone https://github.com/rancher/rancher.git rancherDir
+          cd rancherDir
+          git checkout release/v2.9
```
tomleb, line 75: This could be replaced by 

```sh
git clone -b release/v2.9 https://github.com/rancher/rancher.git rancherDir
cd rancherDir
```

and perhaps even the following to avoid downloading the whole history for that repo.

```sh
git clone --depth 1 -b release/v2.9 https://github.com/rancher/rancher.git rancherDir
cd rancherDir
```

We could also just re-use the `actions/checkout@v4` action for this. Something like:

```sh
- uses: actions/checkout@v4
  with:
    repository: 'rancher/rancher'
    ref: 'release/v2.9'
    path: 'rancherDir'
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+      - name : Fetch build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: webhook-build-artifacts-${{ matrix.arch }}
+          path: ./dist/artifacts
+
+      - name : Debug
+        run : ls -R ./dist/artifacts
+
```
tomleb, line 89: Is this meant to be kept here?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - name : Retrieve builds
+        run : |
+          mkdir -p bin
+          for x in rancher-webhook-integration.test  webhook ; do
+            if [ -e "dist/artifacts/$x" ] ; then
+              cp "dist/artifacts/$x" bin/
+            fi
+          done
```
tomleb, line 97: Since we already know which files to expect, I think it would be best to just copy them without a for loop.

```sh
cp dist/artifacts/rancher-webhook-integration.test
cp dist/artifacts/webhook
```

That way if one of those isn't there for whatever reason, the pipeline will fail here instead of in a later step (since it would silently be ignored here).

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+      - name : setup cluster
+        run : ./.github/workflows/scripts/setup-cluster.sh
+        env:
+          CLUSTER_NAME: webhook
+          K3S_VERSION: v1.28.9-k3s1
+          ARCH: "${{ matrix.arch }}"
+
+# Build the image this way because using the docker/build-push-action
+# results in an
```
tomleb, line 110: Results in an what?

.github/workflows/scripts/start-rancher.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+kubectl get ns | grep -s cattle && flip 'rancher already installed?'
+
+set -exu
+
+TMP="${TMP:-/tmp}"
+
+# `helm repo list` emits error code 1 when there are no repos to list
+set +e
```
tomleb, line 15: I don't see `helm repo list` being used here so do we still need to remove the `-e` flag?

.github/workflows/scripts/start-rancher.sh
```diff
@@ -0,0 +1,31 @@
...
+flip() {
+    echo $1
+    exit 1
+}
+
+kubectl get ns | grep -s cattle && flip 'rancher already installed?'
+
+set -exu
+
+TMP="${TMP:-/tmp}"
```
tomleb, line 12: It isn't clear to me why this variable is being set. I don't see it used elsewhere in the script.

.github/workflows/scripts/start-rancher.sh
```diff
@@ -0,0 +1,31 @@
...
+
+helm upgrade --install cert-manager --namespace cert-manager cert-manager/cert-manager --set installCRDs=true --set "extraArgs[0]=--enable-certificate-owner-ref=true" --create-namespace --wait --timeout=10m
+
+# kubectl get pods --namespace cert-manager
+kubectl rollout status --namespace cert-manager deploy/cert-manager --timeout 1m
+
+# Chart based
+
+CHART_PATH="${CHART_PATH:-/tmp/rancher.tgz}"
+helm upgrade --install rancher $CHART_PATH --namespace cattle-system --set hostname=localhost --wait --timeout=10m --create-namespace --version "${VERSION:-2.9}" --set rancherImage=rancher/rancher --set rancherImageTag="${RANCHER_IMAGE_TAG:-v2.9.0-alpha2}"
```
tomleb, line 31: Personally, I'd rather not have default values for these variables: `VERSION`, `RANCHER_IMAGE_TAG` because:
1. it's yet another place we probably want to update whenever we bump the rancher version 
2. If we don't update / forget to update the defaults, we could be running the integration test on an older version of rancher by mistake, not realizing it, and getting the CI to be green.

charts/rancher-webhook/Chart.yaml
```diff
@@ -11,4 +11,4 @@ annotations:
   catalog.cattle.io/os: linux
   catalog.cattle.io/permits-os: linux,windows
   catalog.cattle.io/rancher-version: ">= 2.9.0-0 < 2.10.0-0"
-  catalog.cattle.io/kube-version: ">= 1.23.0-0 < 1.29.0-0"
```
tomleb, line 14: Imo this should be done as part of https://github.com/rancher/webhook/pull/366. The current version of webhook doesn't support 1.29 so I wouldn't expect the chart to say it does.

scripts/version
```diff
@@ -13,8 +14,12 @@ else
     VERSION="0.0.0-${COMMIT}${DIRTY}"
 fi
 
-if [ -z "$ARCH" ]; then
+if [ -z "${ARCH:-}" ] ; then
     ARCH=$(go env GOHOSTARCH)
+else
+    case "$ARCH" in
+     linux/*) ARCH="${ARCH#*/}" ;;
+    esac
```
tomleb, line 22: That's odd, where have you seen this being needed?

On my machine, `go env GOHOSTARCH` only reports `amd64`, not `linux/amd64`.

scripts/package-for-ci
```diff
@@ -0,0 +1,29 @@
...
+if [[ ${USE_DOCKER_BUILDX:-0} -eq 1 ]]; then
+    docker buildx build --platform linux/amd64 -f ${DOCKERFILE} . -t ${IMAGE} 
+else
+    docker build -f ${DOCKERFILE} -t ${IMAGE} .
+fi
+echo Built ${IMAGE}
+
+docker image save rancher/webhook:$TAG -o ./dist/artifacts/webhook.img
+
+./scripts/package-helm
```
tomleb, line 29: Isn't `./scripts/package-helm` already being run in `./script/package`?

So now `make ci` is running `./scripts/package-helm` twice, correct?

scripts/ci
```diff
@@ -8,4 +8,5 @@ cd $(dirname $0)
 ./validate
 ./validate-ci
 ./package
+./package-for-ci
```
tomleb, line 11: These two scripts are doing duplicate work it seems and I'm not sure why that's necessary.

scripts/integration-test
```diff
@@ -1,83 +1,55 @@
...
+
+# Shut down the core rancher part, but leave the rest of the rancher environment running
 
-# Wait for cluster to start.
-while ! kubectl version >/dev/null 2>&1; do
-    echo "Waiting for cluster to start"
-    sleep 5
-done
+kubectl scale deploy rancher -n cattle-system --replicas=0
+kubectl wait pods -l app=rancher-webhook --for=delete --namespace cattle-system
```
tomleb, line 34: In the unlikely event that the rancher pod dies before the start of the `kubectl wait` call, would the script fail, wait forever or just ignore and continue?

scripts/integration-test
```diff
@@ -1,83 +1,55 @@
...
 
 echo "Uploading new webhook image"
 
-###### Upload the newly created webhook image to containerd, then install the webhook chart using the new image
-IMAGE_FILE=./dist/rancher-webhook-image.tar
-# import image to containerd and get the image name
-WEBHOOK_REPO=$(ctr image import ${IMAGE_FILE} | cut -d ' ' -f 2 | cut -d ':' -f 1)
+#### NO!!! Use ghcr.io ###### Upload the newly created webhook image to containerd, then install the webhook chart using the new image
+#### IMAGE_FILE=./dist/rancher-webhook-image.tar
+#### k3d image import $IMAGE_FILE -c ${CLUSTER_NAME}
```
tomleb, line 42: Are these comments still up to date?

tomleb (commented):
I've only been able to review the CI workflow and the change in the scripts.

Overall I think the approach of using k3d to start a cluster and Helm to start rancher is great.

Couple of things:
- There's a lot of variables spread out in multiple scripts. I think it would be worth it to centralize them in a place to make it easy to change them. For example, the k3d version is in the k3d scripts, some versions are in the ci.yaml file, etc.
- I'd get rid of the default values for most environment variables and just fail if they are unset. This way we avoid having things work with default (potentially stale) values if we remove the override by mistake.
- The .hide-drone.yml file should be deleted instead of hidden.
- Some scripts duplicate work (eg: scripts/package and scripts/package-ci). Fwiw, Dapper makes it difficult (it doesn't solve that in any way), and I'm slowly working on an alternative which hopefully would prevent this..

charts/rancher-webhook/Chart.yaml
```diff
@@ -11,4 +11,4 @@ annotations:
...
+  catalog.cattle.io/kube-version: ">= 1.23.0-0 < 1.30.0-0"
```
tomleb, line 14: Imo this should be done as part of https://github.com/rancher/webhook/pull/366. The current version of webhook doesn't support 1.29 so I wouldn't expect the chart to say it does.

.github/workflows/ci.yaml
ericpromislow: In the repos there are 9 other instances of’fetch-depth: 0’ and 30 of ‘… 1’ which is the default. They can probably all be deleted

.github/workflows/ci.yaml
ericpromislow: Yup, removed

.github/workflows/ci.yaml
ericpromislow: And if I pull  out the part of this that supports k8s 1.29, none of this is needed.

.github/workflows/ci.yaml
ericpromislow: Nope, that's how I discoverd `cp`ing from `dist/artifacts` was losing the `x` bits

.github/workflows/ci.yaml
ericpromislow: true

scripts/version
ericpromislow: This was temporary, not needed anymore.

scripts/package-for-ci
ericpromislow: Right, the `scripts/package` should create the image and install the bits, and `package-helm` builds a chart

scripts/integration-test
ericpromislow: I added timeouts. And because `-e` is on, if a timeout is hit the script should end.

scripts/integration-test
ericpromislow: Cruft for while I was trying to get the image into `k3d`

charts/rancher-webhook/Chart.yaml
ericpromislow: Same as above: this should be done as part of PR 366

scripts/integration-test
```diff
@@ -1,83 +1,55 @@
...
+try --delay 2 --max 30 --waitmsg "Waiting for rancher to start" --failmsg "No rancher here" kubectl rollout status --watch=true --timeout=10s -n cattle-system deploy/rancher
+echo "Rancher deployed"
 
-# Tail the rancher logs if rancher fails to deploy the webhook after 5 minutes.
-bash -c "sleep 300 && echo 'Rancher has not deployed webhook after 5m tailing logs' && tail -f ./rancher.log" &
-# Get PID of the tail command so we can kill it if needed
-TAIL_PID=$!
+# Wait for rancher to start up
+try --delay 2 --max 30 --waitmsg "Waiting for rancher/webhook to be deployed" --failmsg "No rancher here" kubectl rollout status --watch=true --timeout=10s -n cattle-system deploy/rancher-webhook
+echo "Rancher deployed"
```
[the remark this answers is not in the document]
ericpromislow: Done

scripts/integration-test
```diff
@@ -1,83 +1,55 @@
...
-	[[ "$status" != "deployed" ]]
-do 
-	echo "Waiting for helm operation to finish, current status $status"
-	sleep 2
-done
-
-# Kill Rancher since we only need the CRDs and the initial webhook values.
-# We do not want Rancher to reconcile an older version of the webhook during test.
-kill ${RANCHER_PID}
+app_catalog_deployed() {
```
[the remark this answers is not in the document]
ericpromislow: Yes, not sure why I did that....

scripts/integration-test
```diff
@@ -1,83 +1,55 @@
...
 
-# Wait for cluster to start.
-while ! kubectl version >/dev/null 2>&1; do
-    echo "Waiting for cluster to start"
-    sleep 5
-done
+kubectl scale deploy rancher -n cattle-system --replicas=0
+kubectl wait pods -l app=rancher-webhook --for=delete --namespace cattle-system
+
+echo "Rancher has been stopped starting K3s."
```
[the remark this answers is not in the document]
ericpromislow: yup

ericpromislow:
The key to getting an image into the `k3d` cluster's registry is to *not* add a repository prefix to it. Just call it `rancher/webhook` and then when a pod controller tries to load it, it will look in the local registry first before looking at `docker.io` or any other configured repository.

ericpromislow:
Also a note that I added `set -u` to most of the scripts so that they should fail when trying to access an unset variable.

scripts/package-for-ci
MbolotSuse, line 1: I might be missing something, but we don't seem to be using this. Can we remove it?

scripts/version
```diff
@@ -1,35 +1,36 @@
...
 
+DIRTY=
 if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
     DIRTY="-dirty"
 fi
 
-COMMIT=$(git rev-parse --short HEAD)
-GIT_TAG=${DRONE_TAG:-$(git tag -l --contains HEAD | head -n 1)}
+COMMIT=${COMMIT:-$(git rev-parse --short HEAD)}
+GIT_TAG=${GIT_TAG:-${DRONE_TAG:-$(git tag -l --contains HEAD | head -n 1)}}
```
MbolotSuse, line 9: Might also be good to remove the `DRONE_TAG` part since we are off drone now.

MbolotSuse (commented):
Some minor issues, but overall a lot of good work so far.

package/Dockerfile
```diff
@@ -11,4 +11,4 @@ COPY bin/webhook /usr/bin/
 
 USER $user
 
-CMD ["webhook"]
+CMD ["/usr/bin/webhook"]
```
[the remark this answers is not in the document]
ericpromislow: That's an artifact from when I discovered that the `docker-build` action wasn't setting the execute bit. Should be no need for it.

scripts/build
```diff
@@ -14,5 +14,5 @@ LINKFLAGS="-X main.GitCommit=$COMMIT $LINKFLAGS"
 CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/webhook
 if [ "$CROSS" = "true" ] && [ "$ARCH" = "amd64" ]; then
     GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/webhook-darwin
-    GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/webhook-windows
+    GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/webhook-windows-amd64.exe
```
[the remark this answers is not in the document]
ericpromislow: I'm not sure the original code ever worked on Windows (and got agreement from a team member):
```
$ git grep -n webhook-windows
scripts/build:17:    GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/webhook-windows
scripts/package:30:        if [ "$i" = webhook-windows-amd64 ]; then
scripts/package:31:            cp $i dist/artifacts/webhook-windows-amd64.exe
```

So on Windows we're creating an `exe` called `bin/webhook-windows` but explicitly copying
`bin\webhook-windows-amd64` to `bin/webhook-windows-amd64.exe` ?

Believed it was simpler to just create the file with the desired name and then we don't need a special-case `cp` command.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,84 @@
...
+        uses : actions/checkout@v4
+
+      # TODO: Pull this next one out once there's a helm-release for rancher 2.9
+      - name : Checkout rancher/rancher and build the chart
+        run: |
+          pushd "${{ runner.temp}}"
+          git clone --depth 1 -b release/v2.9 https://github.com/rancher/rancher.git rancherDir
+          cd rancherDir
+          ./scripts/chart/build chart
+          sed -i -e 's/kubeVersion: < 1.29.0-0/kubeVersion: < 1.30.0-0/' build/chart/rancher/Chart.yaml
```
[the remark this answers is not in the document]
ericpromislow: That's a less-than, to get 1.29 support. If I didn't include the `sed` command the helm install fails.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
+name : Publish Webhook Images
+
+on:
+  push:
+    tags:
+      - "*"
```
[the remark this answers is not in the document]
ericpromislow: Sure

.github/workflows/ci.yaml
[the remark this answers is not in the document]
MbolotSuse: Thanks for clarifying - below on the setup-cluster stage I see the k3s version as 1.28.9. Even without this we shouldn't need 1.29 support, right?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,37 @@
...
+      ci
+    ]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+      - uses: actions/setup-go@v5
+        with:
+          go-version: 1.22
```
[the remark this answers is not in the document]
ericpromislow: good idea

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+name: Webhook CI
+
+on:
+  workflow_call:
+  push:
+    branches:
+      - release/v*
+    paths-ignore:
+      - '*.md'
```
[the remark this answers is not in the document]
ericpromislow: sure

.github/workflows/scripts/start-rancher.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set +e
+helm repo add cert-manager https://charts.jetstack.io
+helm repo add rancher-latest https://releases.rancher.com/server-charts/latest
+helm repo add jetstack https://charts.jetstack.io
+set -e
+
+helm repo update
+
+helm upgrade --install cert-manager --namespace cert-manager cert-manager/cert-manager --set installCRDs=true --set "extraArgs[0]=--enable-certificate-owner-ref=true" --create-namespace --wait --timeout=10m
```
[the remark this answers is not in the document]
ericpromislow: That's been part of the command-line I've used to install `cert-manager` since the first time I installed Rancher when testing Rancher Desktop.

.github/workflows/scripts/install-k3d.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+set -x
+
+K3D_URL=https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh
+DEFAULT_K3D_VERSION=v5.4.6
```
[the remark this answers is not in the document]
ericpromislow: Sure. Hoping we can automate detecting when versions of things used in scripts change.

.github/workflows/scripts/install-k3d.sh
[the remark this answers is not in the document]
ericpromislow: There are a couple of existing third-party actions. I used one to see how to find the latest release and `install-k3d.sh` script now takes `K3D_VERSION: latest` (not specifying it does the same).

.github/workflows/scripts/setup-cluster.sh
```diff
@@ -0,0 +1,74 @@
...
+    if [[ $readyNodes == 1 ]]
+    then
+      break
+    fi
+    sleep 1
+  done
+}
+
+k3d registry list --no-headers -o json | jq -r '.[].name' | xargs --no-run-if-empty k3d registry delete
+k3d cluster delete $CLUSTER_NAME || true
```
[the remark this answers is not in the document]
ericpromislow: Because I was running this script locally during development.  I moved this code to a local script.

.github/workflows/scripts/install-k3d.sh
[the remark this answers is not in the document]
ericpromislow: I changed the script to pull the latest release if no version is specified (in `K3D_VERSION`)

.github/workflows/scripts/setup-cluster.sh
```diff
@@ -0,0 +1,74 @@
...
+k3d registry create gha -p 42765
+k3d cluster create $CLUSTER_NAME --servers 1 --agents 1 \
+    --registry-use gha:42765 \
+    --image "docker.io/rancher/k3s:${K3S_VERSION}" --api-port 6550
```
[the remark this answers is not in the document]
ericpromislow: I did that when I was getting the k3d registry to work. No longer useful.

.github/workflows/scripts/install-k3d.sh
```diff
@@ -0,0 +1,17 @@
...
+
+install_k3d(){
+  local k3dVersion=${K3D_VERSION:-${DEFAULT_K3D_VERSION}}
+  echo -e "Downloading k3d@${k3dVersion} see: ${K3D_URL}"
+  curl --silent --fail ${K3D_URL} | TAG=${k3dVersion} bash
```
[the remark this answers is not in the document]
ericpromislow: I like the current script -- you can hardwire a version (in the calling YAML code), or have it download the latest release. And it's using the `curl ... | TAG= bash` form from the docs in the same README

.github/workflows/release.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Package release helm charts
+        run: make package-helm
+      - run: mkdir -p ./build/artifacts/ && mv -v ./dist/artifacts/ ./build/
```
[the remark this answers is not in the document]
ericpromislow: yup

.goreleaser.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - CGO_ENABLED=0
+archives:
+    - id: webhook
+      builds:
+        - webhook
+      name_template: '{{ .Binary }}-linux-{{ .Arch }}'
+release:
+  prerelease: auto
+  extra_files:
+    - glob : ./build/artifacts/*.tgz
```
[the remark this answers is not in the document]
ericpromislow: I'm not sure.  From a debug run:

```
Run find dist/artifacts
dist/artifacts
dist/artifacts/webhook-linux-amd64
dist/artifacts/rancher-webhook-0.0.0-2bfcb08.tgz
```

I assume the goreleaser knows about `webhook-linux-amd64` and needs to be told to save the `.tgz` file. And the `checksum` entry is based on the checksum file names in the current release.

I don't see a straightforward way to test this until we push the next release :(

scripts/integration-test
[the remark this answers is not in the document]
ericpromislow: Done. I've never successfully run this before locally, and was hoping I'd be able to now. But if it needs fixing that should happen at a later point.

scripts/package-for-ci
ericpromislow: Once I figured out how to communicate the webhook image tag back to `ci.yaml` there was no need for a separate script.

scripts/version
ericpromislow: Sure.

.github/workflows/publish.yaml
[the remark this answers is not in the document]
MbolotSuse: Resolving in favor of a thread on the new solution.

.github/workflows/scripts/start-rancher.sh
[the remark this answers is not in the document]
MbolotSuse: I don't see that argument as part of our [docs](https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster#4-install-cert-manager). Personally I don't think I've ever used it. I think it would be good to avoid including it if we don't need it

.github/workflows/scripts/install-k3d.sh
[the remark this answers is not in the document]
MbolotSuse: Please double check the licenses of the project(s) that you used for this code and make sure that there's no issues with what was adapted. I think we can make our own action eventually, but that's not necessary for the initial implementation.

.goreleaser.yaml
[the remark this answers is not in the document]
MbolotSuse: The sha256 files seem to be missing from there. So I think even now we can say that something is missing. That being said, I agree that it's hard to test these things outside of a live run. Running an RC once this has merged should give you a way to test. Can you look at the sha256 issue and then we can run the live tests later after things have merged?

.goreleaser.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - linux
+      goarch:
+        - amd64
+        - arm64
+      binary: webhook
+      ldflags:
+        - -extldflags
+        - -static
+        - -s
+        - -X main.Version={{.Version}} -X main.GitCommit={{.Commit}}
```
tomleb, line 18: I guess my main concern is now we have 2 different binaries being built:
- The one with gorelease on releases
- The one with `script/build`

The binaries built in the latter are copied in the docker image and pushed to the container registry.

The binaries built in the former are pushed to the release/pre-release. I'm not sure what they're used for tbh, since afaik what's most important is the Helm chart (.tgz) for the charts repo and the docker image.

We're not trying to redo all our CI, so I'm not suggesting we drop the binaries from the release. However I wanna understand the reason to have two different ways to build the binaries, used in two different artifacts. Maintenance-wise, these two build mechanism will need to be kept in sync. I haven't used goreleaser before so I'd like to know how it benefits us for our use case.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,36 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - release/v*
```
tomleb, line 6: Looking at the tags in webhook, there are no `release/` prefix: https://github.com/rancher/webhook/tags.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
+name : Publish Webhook Images
+
+on:
+  push:
+    tags:
+      - release/v*
```
tomleb, line 6: Looking at the tags in webhook, there are no `release/` prefix: https://github.com/rancher/webhook/tags.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  ci :
+    uses: rancher/webhook/.github/workflows/ci.yaml@release/v0.5
+    permissions:
+      contents: read
+  push:
+    needs : [
+      ci
+    ]
```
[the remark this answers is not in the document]
tomleb: Same applies for the `.github/workflows/release.yaml` file

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
...
+    permissions:
+      contents : read
```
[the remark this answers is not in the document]
ericpromislow: +1 that no-space is better. The space-before-colon form appears in multiple place and my copy-pastos retained it. I'm all for getting rid of it.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
...
+      id-token: write
+    name : Build and push Webhook images
+    runs-on : ubuntu-latest
+    steps:
+      - name : "Read vault secrets"
+        uses : rancher-eio/read-vault-secrets@main
```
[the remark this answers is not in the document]
ericpromislow: same as previous

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          secrets: |
+            secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials username | DOCKER_USERNAME ;
+            secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials password | DOCKER_PASSWORD
+      - name : Checkout repository
```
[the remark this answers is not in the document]
ericpromislow: same as previous

.github/workflows/publish.yaml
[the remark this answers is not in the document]
ericpromislow: yup

.github/workflows/publish.yaml
ericpromislow: Fixed

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
...
+    tags:
+      - release/v*
+
+env:
+  REGISTRY: docker.io
+  REPO : rancher
+
+jobs:
+  ci :
+    uses: rancher/webhook/.github/workflows/ci.yaml@release/v0.5
```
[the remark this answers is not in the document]
ericpromislow: `github.base_ref` returns the branch in `ci.yaml`, I'll need to verify this works in the other workloads (see Michael's comment on further testing of the other workloads still needed).

.github/workflows/release.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+permissions:
+  contents: write
+
+jobs:
+  ci:
+    uses: rancher/webhook/.github/workflows/ci.yaml@release/v0.5
```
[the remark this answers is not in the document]
ericpromislow: see above

MbolotSuse (commented):
Some small changes. Also, would be good to switch the actions to use a SHA instead of the tag to give better control to upgrades.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,54 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ env.DOCKER_USERNAME }}
+          password: ${{ env.DOCKER_PASSWORD }}
+        # setup tag name
+      - if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        run: |
+          echo TAG_NAME=$(echo $GITHUB_REF | sed -e "s|refs/tags/||") >> $GITHUB_ENV
+      - name: Build and push the webhook image
```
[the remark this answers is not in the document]
ericpromislow: I think we're ok. From the logs:
```
Retrieving: docker-24.0.7_ce-150000.198.2.x86_64 (SLE_BCI) (52/52),  25.5 MiB

File: .goreleaser.yaml (hunk 34 above)
Comment (ericpromislow) [reply .goreleaser.yaml:18]: There's a directive around somewhere that we should avoid storing binaries in `artifacts` directories for purposes of SLSA level-3 compliance, so I didn't see a way that we could test the same binaries we push to github without loading everything into one job.

File: .github/workflows/publish.yaml (hunk 36 above)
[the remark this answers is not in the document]
Comment (ericpromislow) [reply .github/workflows/publish.yaml]: I don't think it needs to be quoted because the string value doesn't start with a `*`:
```
$ pbpaste | yq .
{
  "on": {
    "push": {
      "tags": [
        "release/v*"
      ]
    }
  }
}
```

File: .github/workflows/publish.yaml (hunk 41 above)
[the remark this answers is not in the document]
Comment (tomleb) [reply .github/workflows/publish.yaml]: Here's what I've found about re-using a workflow that's part of the current repo. Have a look at https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow:
> If you use the second syntax option (without {owner}/{repo} and @{ref}) the called workflow is from the same commit as the caller workflow.

This would look like:

```yaml

jobs:
  ci :
    uses: ./.github/workflows/ci.yml
```

This way no need to fiddle with any branch names/commits,etc.

.goreleaser.yaml
tomleb: Is there anything that would prevent replacing the goreleaser step with a `make build` step, or something like that?

ericpromislow: I thought we want to use goreleaser because it automates more of the release process.

ericpromislow:
Actions now use SHAs

charts/rancher-webhook/Chart.yaml
tomleb: @ericpromislow This is still part of the PR fyi, I still think it should be removed

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,83 @@
...
+jobs:
+  build:
+    name: CI
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+          arch:
+          - amd64
+          # TODO: See wrangler for an example of how to run tests on arm64, not ready here
+          # - arm64
```
tomleb, line 32: With wrangler supporting both, we should be able to make it work here as well.

ericpromislow:
My mistake to close this.  The plan is to merge this PR and then create a test-release and test the publish and release parts.

ericpromislow:
We're going to move ahead with this PR and skip the failing test on ARM64 CI runs ... this is logged at
https://github.com/rancher/rancher/issues/45837

tomleb (approved):

.goreleaser.yaml
```diff
@@ -0,0 +1,41 @@
+# Make sure to check the documentation at https://goreleaser.com
```
tomleb, line 1: Is this still needed? I don't see anything using it.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: get vars
+        run: cat dist/image_tag >> $GITHUB_ENV
+
+      - name: Run integration tests
+        id: integrationTests
+        run: ./.github/workflows/scripts/integration-test-ci
+        # See https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context
+        # for explanation of steps.continue-on-error and steps.ID.outcome
+        continue-on-error: true
```
tomleb, line 87: Can you clarify why we are continuing on errors? Does the CI still fail if there is an error or does it consider the run a success? Reading the docs, I understand that we're have `steps.ID.outcome` and `steps.ID.conclusion` with different values, but I don't see them used anywhere in the GHA scripts here, so why is this necessary?

tests/integration/rkeMachineConfig_test.go
```diff
@@ -1,12 +1,18 @@
...
+
 	"github.com/rancher/webhook/pkg/auth"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 func (m *IntegrationSuite) TestRKEMachineConfig() {
+	if runtime.GOARCH == "arm64" && os.Getenv("CI") != "" {
+		m.T().Skip("Skipping the RKE Machine-Config test on arm64 in CI -- machine info not available")
+	}
```
tomleb, line 15: nit: Could add a comment to describe why we're skipping a test

.goreleaser.yaml
ericpromislow: It came from the boilerplate. I'll modify it because I like to have pointers to docs for third-party components, even when the link is obvious

.github/workflows/ci.yaml
ericpromislow: cruft

tests/integration/rkeMachineConfig_test.go
ericpromislow: Sure

tomleb (approved):

tomleb (approved):

MbolotSuse (approved):

MbolotSuse:
LGTM in interest of functionality, some things we will want to follow up on.

PR #384, opened 2024-05-29
[v0.4] [GHA] Migrate v0.4 CI to GHA

ericpromislow:
Backport PR 383 (migrate v0.5 to GHA)

Related to https://github.com/rancher/rancher/issues/45231


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,73 @@
...
+          K3S_VERSION: v1.28.11-k3s1
+          ARCH: "${{ matrix.archBox.arch }}"
+
+      - name: import image
+        run: k3d image import dist/rancher-webhook-image.tar -c webhook
+
+      - name: start rancher
+        run: ./.github/workflows/scripts/start-rancher.sh
+        env:
+          VERSION: "2.8"
```
tomleb, line 62: Imo this should use the same mechanism that we use for release/v0.5 because right now this only allows running tests on a previously released version of Rancher.

eg: If I add a `Foo` object in rancher 2.8, the CI will not pick it up because it will install a released version of rancher. We should use the v2.8-head image instead and its chart so we can test unreleased changes to CRDs.

.github/workflows/scripts/start-rancher.sh
```diff
@@ -0,0 +1,25 @@
...
+set -e
+
+helm repo update
+
+helm upgrade --install cert-manager --namespace cert-manager cert-manager/cert-manager --set installCRDs=true --create-namespace --wait --timeout=10m
+
+# kubectl get pods --namespace cert-manager
+kubectl rollout status --namespace cert-manager deploy/cert-manager --timeout 1m
+
+helm upgrade --install rancher rancher-latest/rancher --namespace cattle-system --set hostname=localhost --wait --timeout=10m --create-namespace --version "$VERSION"
```
tomleb, line 25: Same as mentioned above, this should use v2.8-head and a chart from a freshly cloned repo.

tomleb (commented):
LGTM except for the change on how we start Rancher for the CI

.github/workflows/ci.yaml
ericpromislow: done

.github/workflows/scripts/start-rancher.sh
ericpromislow: done

tomleb (dismissed):

MbolotSuse (commented):
One comment on upgraded go version but LGTM otherwise.

MbolotSuse (approved):

PR #398, opened 2024-06-13
Don't error if external-rules feature flag is not found when evaluating RoleTemplates

raulcabello:
## Problem
webhook is automatically upgraded in Rancher 2.8.3. This is causing webhook to fail `RoleTemplate` evaluation because the feature flag doesn't exist.

## Solution
act as if feature flag is disabled if it is not present.

## CheckList
  
- [ ] Test
  
- [ ] Docs


pmatseykanets (approved):

crobby (approved):

tomleb (approved):
LGTM

PR #402, opened 2024-06-17
[v0.4] s4: Fixes 227 (#390)

pmatseykanets:
## Issue: 

Port of #390

* Validate user retention userattribute fields and settings
* [v0.4.s4] Don't allow empty lastLogin user attribute

## CheckList
  
- [ ] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/setting/Setting.md
+++ b/pkg/resources/management.cattle.io/v3/setting/Setting.md
@@ -1,0 +1,19 @@
+## Validation Checks
+
+### Invalid Fields - Create
+
+When a Setting is created, the following checks take place:
+
+- If set, `disable-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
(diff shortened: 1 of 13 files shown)


maxsokolovsky (approved):

pkg/resources/management.cattle.io/v3/setting/Setting.md
```diff
@@ -0,0 +1,19 @@
+## Validation Checks
+
+### Invalid Fields - Create
+
+When a Setting is created, the following checks take place:
+
+- If set, `disable-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
+- If set, `delete-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
```
andreas-kupries, line 8: The `MinDelete...` checks seems to be generally not documented. Is that intentional ?

pmatseykanets: Good catch, thanks! But needs to be added outside of the port PR.

PR #404, opened 2024-06-17
[v0.5] s4: Fixes 227

pmatseykanets:
## Issue: 

Port of #390

* Validate user retention userattribute fields and settings
* [v0.4.s4] Don't allow empty lastLogin user attribute

## CheckList
  
- [ ] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/setting/Setting.md
+++ b/pkg/resources/management.cattle.io/v3/setting/Setting.md
@@ -1,0 +1,19 @@
+## Validation Checks
+
+### Invalid Fields - Create
+
+When a Setting is created, the following checks take place:
+
+- If set, `disable-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
+- If set, `delete-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
+- If set, `user-last-login-default` must be a date time according to RFC3339 (e.g. `2023-11-29T00:00:00Z`).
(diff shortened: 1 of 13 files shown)


ericpromislow (changes requested):
The PR fails because the drone config uses `rancher/rancher:v2.9-head` and that image quietly changed its base image from `registry.suse.com/bci-base` to `bci-micro`, and the latter doesn't include `zypper` .

I fixed this on other PRs that still use drone by moving it back to `v2.9.0-alpha1`

pmatseykanets:
@ericpromislow Thanks for the insights! It seems like the required change is not related to this port in any way and should be applied is a separate PR. Then this one can be rebased.

pmatseykanets:
Depends on and needs to be rebased off of #406

JonCrowther (approved):

raulcabello (approved):

pkg/resources/management.cattle.io/v3/setting/Setting.md
```diff
@@ -0,0 +1,19 @@
+## Validation Checks
+
+### Invalid Fields - Create
+
+When a Setting is created, the following checks take place:
+
+- If set, `disable-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
+- If set, `delete-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
```
andreas-kupries, line 8: The `MinDeleteInactiveUserAfter` check does not seem to be documented. Is that intentional ?

PR #406, opened 2024-06-17
Fix integration test

tomleb:
Rancher v2.9-head image no longer contains zypper so we need another way to install Helm.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -21,7 +21,7 @@ steps:
     pull: always
     privileged: true
     commands:
-      - zypper -n install helm
+      - curl --cacert /etc/ssl/ca-bundle.pem -sL https://get.helm.sh/helm-v3.13.3-linux-amd64.tar.gz | tar xvzf - -C /usr/local/bin --strip-components=1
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -21,7 +21,7 @@ steps:
     pull: always
     privileged: true
     commands:
-      - zypper -n install helm
+      - curl --cacert /etc/ssl/ca-bundle.pem -sL https://get.helm.sh/helm-v3.13.3-linux-amd64.tar.gz | tar xvzf - -C /usr/local/bin --strip-components=1
```
tomleb, line 24: rancher-v2.9 uses its own ssl bundles somewhere else so curl doesn't work without this.

pmatseykanets (approved):

.drone.yml
ericpromislow: This is fine. I don't care whether we leave the `v` flag in the `tar` command or not -- seeing the file get extracted vs. the extra output.

ericpromislow (approved):

PR #410, opened 2024-06-19
Data dir rfc

jakefhyde:
## CheckList
  
- [ ] Test
  
- [ ] Docs


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/rancher/webhook
 
 go 1.22
 
+replace github.com/rancher/rancher/pkg/apis => /home/jfh/Documents/work/rancher/rancher/pkg/apis
```
ericpromislow, line 5: I don't think you intended to include this

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +125,75 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+func (p *provisioningAdmitter) validateAgentEnvVars(cluster *v1.Cluster) *metav1.Status {
+	if cluster.Spec.RKEConfig == nil {
+		return nil
+	}
+	if slices.ContainsFunc(cluster.Spec.AgentEnvVars, func(envVar rkev1.EnvVar) bool {
+		return envVar.Name == "CATTLE_AGENT_VAR_DIR"
+	}) {
+		return &metav1.Status{
+			Status:  "Failure",
+			Message: `"CATTLE_AGENT_VAR_DIR" cannot bet set within "cluster.Spec.RKEConfig.AgentEnvVars": use "cluster.Spec.RKEConfig.DataDirectories.SystemAgent"`,
```
ericpromislow, line 139: typo: `s/bet/be/`

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +125,75 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+			Code:    http.StatusBadRequest,
+		}
+	}
+
+	return nil
+}
+
+// validateDataDirectories will validate updates to the cluster object to ensure the data directories are not changed.
+// The only exception when a data directory is allowed to be changed is if cluster.Spec.AgentEnvVars has an env var with
+// for "CATTLE_AGENT_VAR_DIR", which Rancher will perform a once time migration to set the
```
ericpromislow, line 150: comment typo: `s/once time/one-time/`

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +125,75 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+// cluster.Spec.RKEConfig.DataDirectories.SystemAgent field for. validateAgentEnvVars will ensure "CATTLE_AGENT_VAR_DIR"
```
ericpromislow, line 151: `s/ for.//` ?

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +125,75 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+		return nil
+	}
+	oldEnvVars := slices.DeleteFunc(slices.Clone(oldCluster.Spec.AgentEnvVars), func(envVar rkev1.EnvVar) bool {
+		return envVar.Name != "CATTLE_AGENT_VAR_DIR"
+	})
+	// will be true when rancher performs one time migration to set SystemAgent data directory
+	if len(oldEnvVars) > 0 && newCluster.Spec.RKEConfig.DataDirectories.SystemAgent != oldEnvVars[len(oldEnvVars)-1].Value {
+		return &metav1.Status{
+			Status:  "Failure",
+			Message: `"CATTLE_AGENT_VAR_DIR" cannot bet set within "cluster.Spec.RKEConfig.AgentEnvVars": use "cluster.Spec.RKEConfig.DataDirectories.SystemAgent"`,
```
ericpromislow, line 167: same typo: `s/bet/be/`

ericpromislow (dismissed):
I don't have context, but found typos and don't expect a `go.mod replace` pointing to a personal path to ever work

jakefhyde:
@ericpromislow Updated. CI won't pass until https://github.com/rancher/rancher/pull/45326 is merged anyways.

go.mod
```diff
@@ -18,6 +18,8 @@ replace (
 	k8s.io/controller-manager => k8s.io/controller-manager v0.29.3
 	k8s.io/cri-api => k8s.io/cri-api v0.29.3
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.29.3
+	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.29.3
+	k8s.io/endpointslice => k8s.io/endpointslice v0.29.3
```
jakefhyde, line 22: Intellisense fails without doing this.

Oats87 (dismissed):
high level LGTM

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +126,108 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+				newCount++
+			}
+		}
+		if newCount > oldCount {
+			return &metav1.Status{
+				Status:  "Failure",
+				Message: `"CATTLE_AGENT_VAR_DIR" cannot be set within "cluster.Spec.RKEConfig.AgentEnvVars": use "cluster.Spec.RKEConfig.DataDirectories.SystemAgent"`,
+				Reason:  metav1.StatusReasonBadRequest,
+				Code:    http.StatusBadRequest,
+			}
```
MbolotSuse, line 166: Here and in several other places you are manually crafting the status field. This is a bit wordy (4 lines as opposed to 1) and doesn't explicitly set the `Result.Allowed` field. I would recommend using the webhook's `admission` package, which contains an `ResponseBadRequest` function [here](https://github.com/rancher/webhook/blob/276b2b953427d02f5ea2ae718ffa7e206399e725/pkg/admission/admission.go#L348)- this fills in a lot of these fields and sets the allowed flag on the response. You will need to return an `AdmissionResponse` object from your funciton, but I think that this will make it a lot more readable:

```suggestion
			return admission.ResponseBadRequest( `"CATTLE_AGENT_VAR_DIR" cannot be set within "cluster.Spec.RKEConfig.AgentEnvVars": use "cluster.Spec.RKEConfig.DataDirectories.SystemAgent"`)
```

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +126,108 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+			Code:    http.StatusBadRequest,
+		}
+	} else if oldCluster.Spec.RKEConfig.DataDirectories.Provisioning != newCluster.Spec.RKEConfig.DataDirectories.Provisioning {
+		return &metav1.Status{
+			Status:  "Failure",
+			Message: "Provisioning data directory cannot be changed after cluster creation",
+			Reason:  metav1.StatusReasonBadRequest,
+			Code:    http.StatusBadRequest,
+		}
+	} else if oldCluster.Spec.RKEConfig.DataDirectories.K8sDistro != newCluster.Spec.RKEConfig.DataDirectories.K8sDistro {
```
MbolotSuse, line 219: Have you considered simplifying this into a single case? Something like:

```suggestion
	if reflect.DeepEqual(oldCluster.Spec.RKEConfig.DataDirectories, newCluster.Spec.RKEConfig.DataDirectories){
```

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +126,108 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+			if envVar.Name == "CATTLE_AGENT_VAR_DIR" {
+				oldCount++
+			}
+		}
+		for _, envVar := range newCluster.Spec.AgentEnvVars {
+			if envVar.Name == "CATTLE_AGENT_VAR_DIR" {
+				newCount++
+			}
+		}
+		if newCount > oldCount {
```
MbolotSuse, line 160: I'm not sure that this works as intended. From what I can see here, I could change this value as long as I added/removed in a single step. For example, if I change `CATTLE_AGENT_VAR_DIR` from `/var/lib/rke2` to `/var/lib/new`, that would work.

That being said, I'm not 100% sure that this is intentional - open to hearing otherwise.

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -200,3 +202,337 @@ func TestValidateMachinePoolName(t *testing.T) {
...
+	t.Parallel()
+
+	tests := []struct {
+		name          string
+		request       *admission.Request
+		cluster       *v1.Cluster
+		oldCluster    *v1.Cluster
+		shouldSucceed bool
+	}{
+		{
```
MbolotSuse, line 216: One other test case that would be good to add: there is an RKEConfig with some agent env vars, but not the AGENT_VAR_DIR.

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -200,3 +202,337 @@ func TestValidateMachinePoolName(t *testing.T) {
...
```
MbolotSuse, line 334: One other case that would be good to add: the cluster exists, but RKEConfig is nil.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +126,108 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
 
+// validateAgentEnvVars will validate the provisioning cluster object's AgentEnvVars, ensuring that the
+// "CATTLE_AGENT_VAR_DIR" env var is not allowed to be updated.
+func (p *provisioningAdmitter) validateAgentEnvVars(request *admission.Request, oldCluster, newCluster *v1.Cluster) *metav1.Status {
+	if newCluster.Spec.RKEConfig == nil {
+		return nil
+	}
+	if request.Operation == admissionv1.Create {
+		if slices.ContainsFunc(newCluster.Spec.AgentEnvVars, func(envVar rkev1.EnvVar) bool {
+			return envVar.Name == "CATTLE_AGENT_VAR_DIR"
```
MbolotSuse, line 137: This env var sees a lot of re-use across this function and the related function. I think it would be a good idea to set it as a const at the top of the file to avoid typos.

MbolotSuse (changes requested):
Couple of suggestions on organization/function re-use. One other use case that I'm not sure that this addresses is changing the env var from one value to another - I think this permits that operation, but I'm not sure that's intentional.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
jakefhyde: Honestly it's not super clear but it is functionally correct: this function is making sure that new `CATTLE_AGENT_VAR_DIR` vars are only allowed to be added, not removed. The other function is supposed to do validation, which is covered by this case:
```
	oldEnvVars := slices.DeleteFunc(slices.Clone(oldCluster.Spec.AgentEnvVars), func(envVar rkev1.EnvVar) bool {
		return envVar.Name != systemAgentVarDirEnvVar
	})
	// will be true up until rancher performs one time migration to set SystemAgent data directory
	if len(oldEnvVars) > 0 {
		newEnvVars := slices.DeleteFunc(slices.Clone(newCluster.Spec.AgentEnvVars), func(envVar rkev1.EnvVar) bool {
			return envVar.Name != systemAgentVarDirEnvVar
		})
		if reflect.DeepEqual(newEnvVars, oldEnvVars) && oldCluster.Spec.RKEConfig.DataDirectories.SystemAgent == newCluster.Spec.RKEConfig.DataDirectories.SystemAgent {
			// rancher migration has not been performed yet
			return admission.ResponseAllowed()
		}
		if newCluster.Spec.RKEConfig.DataDirectories.SystemAgent != oldEnvVars[len(oldEnvVars)-1].Value {
			return admission.ResponseBadRequest(
				fmt.Sprintf(`"%s" cannot be set within "cluster.Spec.RKEConfig.AgentEnvVars": use "cluster.Spec.RKEConfig.DataDirectories.SystemAgent"`, systemAgentVarDirEnvVar))
		}
```
so if we were defining the env var previously, and we haven't changed either the env var or data directory field in this update, assume it's fine (actually as I look at it we shouldn't allow this without validating the other data directories but you get the idea).

MbolotSuse: I might be missing something, but the current method will only stop me from adding/removing the env var right? So if I just change it, it's allowed. From your reply we don't want that, it needs to be the same value as before correct?

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
jakefhyde: Still have to leave separate but will become clear as to why when I push.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -103,6 +113,14 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
 		if err := p.validateCloudCredentialAccess(request, response, oldCluster, cluster); err != nil || response.Result != nil {
 			return response, err
 		}
+
+		if response = p.validateAgentEnvVars(request, oldCluster, cluster); response.Result != nil {
```
MbolotSuse, line 117: I would handle the "ok" cases differently than this. Right now this depends on `response.Result` being nil for `ResponseAllowed()` which I think is very dangerous long term. I would recommend one of the following options:
- Only return a response if it fails validation and return nil otherwise. In this function, check `response != nil` to see if you should break with a response
- In this function, check if `!response.Allowed` - this way, if it fails we short-circuit and return early.
- Return a bool and an `*admissionv1.AdmissionResponse`. Use the bool to check if the response was allowed or not. 

Big thing here is not to closely couple this code to what the allowed response does like it's currently doing.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +131,83 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+// cluster.Spec.RKEConfig.DataDirectories.SystemAgent field for the cluster. validateAgentEnvVars will ensure
+// "CATTLE_AGENT_VAR_DIR" is not added, so this exception only applies to the one-time Rancher migration.
+func (p *provisioningAdmitter) validateDataDirectories(request *admission.Request, oldCluster, newCluster *v1.Cluster) *admissionv1.AdmissionResponse {
+	if request.Operation != admissionv1.Update {
+		return admission.ResponseAllowed()
+	}
+	if newCluster.Spec.RKEConfig == nil {
+		return admission.ResponseAllowed()
+	}
+	oldSystemAgentVarDirEnvVars := slices.DeleteFunc(slices.Clone(oldCluster.Spec.AgentEnvVars), func(envVar rkev1.EnvVar) bool {
```
MbolotSuse, line 183: Nit: It occurs to me that you could do this better with a simple extraction func to find the variable. Something like:

```
def getEnvVar(name string, envVars []rkev1.EnvVar) *rkev1.EnvVar{
    for _, envVar := range envVars{
        if envVar.Name == name{
            return &envVar
        }
    }
    return nil
}
```

Main benefit that we get here is that we are no longer cloning/modifying a slice, so I'd imagine the performance here is going to be a bit better.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -113,6 +131,83 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
...
+		return envVar.Name != systemAgentVarDirEnvVar
+	})
+	// will be true up until rancher performs one time migration to set SystemAgent data directory if env var was previously set
+	if len(oldSystemAgentVarDirEnvVars) > 0 {
+		newSystemAgentVarDirEnvVars := slices.DeleteFunc(slices.Clone(newCluster.Spec.AgentEnvVars), func(envVar rkev1.EnvVar) bool {
+			return envVar.Name != systemAgentVarDirEnvVar
+		})
+		if !reflect.DeepEqual(newSystemAgentVarDirEnvVars, oldSystemAgentVarDirEnvVars) || oldCluster.Spec.RKEConfig.DataDirectories.SystemAgent != newCluster.Spec.RKEConfig.DataDirectories.SystemAgent {
+			if newCluster.Spec.RKEConfig.DataDirectories.SystemAgent != oldSystemAgentVarDirEnvVars[len(oldSystemAgentVarDirEnvVars)-1].Value {
```
MbolotSuse, line 192: Nit: I think that this is a bit tough to follow. I think this would also be easier to write once the actual env var is extracted and you can move out of slices.

In addition, since the second if statement doesn't have an else (or any other additional logic). it can be combined into one if statement.

MbolotSuse (changes requested):
Couple of small comments and 1 unresolved thread.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
jakefhyde: The issue with this suggestion is that it returns the first occurrence of the env var, and we want the last, as it's possible to specify an env var multiple times.

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -200,3 +202,587 @@ func TestValidateMachinePoolName(t *testing.T) {
...
```
MbolotSuse, line 570: A few test cases that would be good to add:
- Create but there's the systemAgentVarDirEnvVar.
- Request other than create/update (delete), but the spec is filled in

pkg/resources/provisioning.cattle.io/v1/cluster/validator_test.go
```diff
@@ -200,3 +202,587 @@ func TestValidateMachinePoolName(t *testing.T) {
...
+func TestValidateSystemAgentDataDirectory(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name          string
+		cluster       *v1.Cluster
+		oldCluster    *v1.Cluster
+		shouldSucceed bool
+	}{
+		{
```
MbolotSuse, line 215: Some test cases that would be good to add:
- old env var is set, new DataDirectories.System agent is set, new env var is still provided
- old env var is set, new DataDirectories.System agent is set but doesn't agree with the old value

MbolotSuse (changes requested):
Some small test cases to add but outside of that LGTM

MbolotSuse (approved):

MbolotSuse:
@jakefhyde please squash the commits when merging.

Oats87 (approved):

PR #411, opened 2024-06-19
Remove PSP validation

raulcabello:
Remove PSP validation. PSP fields were removed from the `Cluster` resource in Rancher v2.9 in this [PR](https://github.com/rancher/rancher/pull/45380), this causes the webhook to fail if the Rancher version is bumped. 

The minimum k8s version that rancher v2.9 will support is 1.25 where PSP is removed from the upstream.


tomleb:
This should be fixed with https://github.com/rancher/webhook/pull/366/

> The minimum k8s version that rancher v2.9 will support is 1.25
I was under the impression the minimum version would be 1.27. Would you mind letting me know where you've seen 1.25? Thanks!

raulcabello:
Thanks @tomleb! superseded by https://github.com/rancher/webhook/pull/366

PR #414, opened 2024-06-25
Remove external-rules feature flag

raulcabello:
## Problem
`external-rules` feature flag was added in Rancher v2.7 and v2.8 to help admins with the transition of the new `RoleTemplate` field `ExternalRules`. That's no longer needed in Rancher v2.9
 
## Solution
Remove `external-rules` feature flag. It will work as it is always enabled, and can't be disabled.

### DIFF
--- a/pkg/auth/roleTemplate.go
+++ b/pkg/auth/roleTemplate.go
@@ -9,21 +9,17 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 )
 
-const ExternalRulesFeature = "external-rules"
-
 // RoleTemplateResolver provides an interface to flatten role templates into slice of rules.
 type RoleTemplateResolver struct {
 	roleTemplates v3.RoleTemplateCache
 	clusterRoles  v1.ClusterRoleCache
-	features      v3.FeatureCache
 }
 
 // NewRoleTemplateResolver creates a newly allocated RoleTemplateResolver from the provided caches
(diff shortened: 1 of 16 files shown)


pmatseykanets (approved):

pkg/auth/roleTemplate.go
```diff
@@ -63,25 +59,12 @@ func (r *RoleTemplateResolver) gatherRules(roleTemplate *rancherv3.RoleTemplate,
...
-				rules = append(rules, cr.Rules...)
-			}
-		} else if roleTemplate.Context == "cluster" {
+		if roleTemplate.ExternalRules != nil {
+			rules = append(rules, roleTemplate.ExternalRules...)
+		} else {
 			cr, err := r.clusterRoles.Get(roleTemplate.Name)
 			if err != nil {
-				return nil, fmt.Errorf("failed to get ClusterRole %q: %w", roleTemplate.Name, err)
+				return nil, fmt.Errorf("for external RoleTemplates, externalRules must be provided or a backing clusterRole must be installed to check for privilege escalations: failed to get ClusterRole %q: %w", roleTemplate.Name, err)
```
andreas-kupries, line 67: Explicitly noting that this looks ok, given that a similar change in the https://github.com/rancher/rancher/pull/45864#pullrequestreview-2138671649 got marked as problem.

andreas-kupries (approved):

PR #416, opened 2024-06-28
Add a check for the agent-tls-mode setting

maxsokolovsky:
## Problem

The webhook needs to validate updates to the `agent-tls-mode` setting given the logic below.

## Solution

If `agent-tls-mode` has `default` or `value` updated from `system-store` to `strict`, then all non-local clusters must
  have a status condition `AgentTlsStrictCheck` set to `True`, unless the new setting has an overriding
  annotation `cattle.io/force=true`.

## CheckList
  
- [x] Test
  
- [x] Docs


pkg/resources/management.cattle.io/v3/setting/validator.go
```diff
@@ -129,3 +164,21 @@ func validateDuration(value string) (time.Duration, error) {
 
 	return dur, err
 }
+
+func clusterConditionMatches(cluster *v3.Cluster, t v3.ClusterConditionType, status v1.ConditionStatus) bool {
```
JonCrowther, line 168: I appreciate the tests on `effectiveValue`, I think for the sake of completeness we could add a test for `clusterConditionMatches` too

JonCrowther (commented):
Everything looks good, just one comment.

ericpromislow (approved):
LGTM. Verified everything passes with the GHA-based CI

pkg/resources/management.cattle.io/v3/setting/validator.go
maxsokolovsky: I actually got rid of the tests of `effectiveValue`, as it will be tested by the exported `Admit` method - same with `clusterConditionMatches`.

ericpromislow (approved):
LGTM

JonCrowther (approved):

MbolotSuse (approved):

JonCrowther (dismissed):

MbolotSuse (approved):
Please squash when merging.

JonCrowther (approved):

PR #438, opened 2024-07-08
Trivial cleanup changes - no actual effect

ericpromislow:
Removing extra spaces, bogus comments, and a duplicate `--set replicas=1`


tomleb (approved):

nflynt (approved):
These look fine, but I don't actually see the `--set replicas=1` change. Did that not get included?

PR #442, opened 2024-07-09
Don't fail fast CI workflow

tomleb:
# Issue: https://github.com/rancher/rancher/issues/45231

Technically not required for https://github.com/rancher/rancher/issues/45231 but related.

With flaky tests or Docker rate limit, it's very annoying to have a job's failure cancel another job. Instead, let all jobs run to completion, this way we can restart only the failed one.


ericpromislow (approved):
Agreed

PR #443, opened 2024-07-09
[v0.4] update rke to v1.5.11-rc4

kinarashah:
https://github.com/rancher/rancher/issues/45950


maxsokolovsky (approved):

go.mod
```diff
@@ -40,23 +40,23 @@ require (
...
+	golang.org/x/tools v0.21.1-0.20240508182429-e35e4ccd0d2d
+	k8s.io/api v0.28.9
+	k8s.io/apimachinery v0.28.9
+	k8s.io/apiserver v0.28.9
 	k8s.io/client-go v12.0.0+incompatible
 	k8s.io/kubernetes v1.28.8
 	k8s.io/pod-security-admission v0.28.6
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
-	sigs.k8s.io/controller-runtime v0.15.3
+	sigs.k8s.io/controller-runtime v0.16.6
```
MbolotSuse, line 59: Can you clarify why all of these dependencies were bumped? In particular, I'm concerned about the controller-runtime bump since we do make use of it for some very important webhook functionality: https://github.com/rancher/webhook/blob/7db7696c660bb4c23d0bb6b6ce7b420ab808de3a/pkg/patch/patch.go#L8 .

kinarashah: @MbolotSuse So I didn't update them manually but `go mod tidy` picked them up. Looks like rancher v2.8 branch is using this version for controller-runtime 
`sigs.k8s.io/controller-runtime v0.16.6` 

For the k8s components, we replace them with `v0.28.6` in rancher and the require is at `v0.28.9` so my understanding is go picked it up from `require` section. I could add replace directive for k8s components so they match with rancher's go.mod version to keep them consistent. 

For controller-runtime, would we prefer to keep it in line with rancher's version or pinning it to `v0.15.3`?

MbolotSuse: I'd like to avoid a replace directive. I think we can go with this given the integration tests should cover this case and we are above this version on a future WH version. That being said, I'd like to avoid changes like this in the future when we are this close to a release.

MbolotSuse (approved):

PR #445, opened 2024-07-10
[v0.5] update rke to v1.6.0-rc10 and pkg/apis

kinarashah:
https://github.com/rancher/rancher/issues/45952

Context on why `replace` is required: https://github.com/rancher/rancher/pull/43630 

Will remove `replace` after RKE `v1.6.0` is tagged


maxsokolovsky (approved):

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.22.0
 toolchain go1.22.5
 
 replace (
+	github.com/rancher/rke => github.com/rancher/rke v1.6.0-rc10
```
MbolotSuse, line 8: @kinarashah can you elaborate on why you've added a replace directive for this import? I'd like to avoid adding new replacement directives wherever possible.

go.mod
```diff
@@ -41,17 +42,17 @@ require (
...
 	github.com/robfig/cron v1.2.0
 	github.com/sirupsen/logrus v1.9.3
 	github.com/stretchr/testify v1.9.0
 	golang.org/x/exp v0.0.0-20240416160154-fe59bbe5cc7f
 	golang.org/x/text v0.16.0
 	golang.org/x/tools v0.21.1-0.20240508182429-e35e4ccd0d2d
-	k8s.io/api v0.30.1
-	k8s.io/apimachinery v0.30.1
+	k8s.io/api v0.30.2
+	k8s.io/apimachinery v0.30.2
```
MbolotSuse, line 55: Can you elaborate on why this was bumped? If this is getting replaced above I'm not sure that I see the value.

go.mod
kinarashah: @MbolotSuse I understand the need to avoid replacement directives, note that this is needed only for RCs and will be removed in the final version after RKE is released. 

The reason behind this is that we don't tag RKE RCs with strict semver versioning, we need them to be dot separated (`rc.10` in this case and we'd be changing our RC tags in future). Without dots, go mod considers `rc10` < `rc9` so it doesn't pick the latest RC and stays at `rc9`. Once release is tagged, we can remove this line when we bump it to `v1.6.0`.

go.mod
kinarashah: I did not update this manually but this was automatically picked from rancher's require section with `go mod tidy`. Not sure why it ignores the `replace` directive.

go.mod
MbolotSuse: Alright, please make sure to file the follow-up PR before the release goes out.

MbolotSuse (approved):

PR #446, opened 2024-07-10
[v0.5] Set FailurePolicy to Ignore for settings validator

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/45841

## Problem

We have a "catch 22" situation with the settings validator as rancher may update some settings on startup while the webhook is starting up.

## Solution

Change the `FailurePolicy` to `Ignore` on settings' validator to avoid timeouts.

## CheckList
  
- [x] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/setting/validator.go
+++ b/pkg/resources/management.cattle.io/v3/setting/validator.go
@@ -57,9 +57,9 @@ func (v *Validator) Operations() []admissionregistrationv1.OperationType {
 
 // ValidatingWebhook returns the ValidatingWebhook.
 func (v *Validator) ValidatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) []admissionregistrationv1.ValidatingWebhook {
-	return []admissionregistrationv1.ValidatingWebhook{
-		*admission.NewDefaultValidatingWebhook(v, clientConfig, admissionregistrationv1.ClusterScope, v.Operations()),
-	}
+	valWebhook := admission.NewDefaultValidatingWebhook(v, clientConfig, admissionregistrationv1.ClusterScope, v.Operations())
+	valWebhook.FailurePolicy = admission.Ptr(admissionregistrationv1.Ignore)
+	return []admissionregistrationv1.ValidatingWebhook{*valWebhook}
 }
 
 // Admitters returns the admitter objects.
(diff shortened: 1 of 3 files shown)


tomleb:
Adding @maxsokolovsky to know his thoughts since this can affect the newest agent-tls-mode settings. From my POV it seems fine.

tomleb (dismissed):

maxsokolovsky (commented):
I'm not sure I fully understand the solution. This isn't the case of handling an unrecognized error, right? It's the case of the webhook not being available at all, so it's never even reached. What happens on the Rancher side with this change? Would you mind clarifying this part?

maxsokolovsky (dismissed):
Met with @pmatseykanets and @MbolotSuse to discuss this - everything makes sense.

MbolotSuse (changes requested):
Blocking merge until we decide on milestone.

MbolotSuse (approved):

MbolotSuse (approved):
LGTM, please squash when merging

ericpromislow (commented):
Two questions: 

1. Where is `<some-object>.FailurePolicy` used as an r-value? Ignoring the tests, I only see it set, never consulted. I looked in other modules that pull in webhook (it's possible I haven't pulled down all of them), and don't see it referenced in any of those modules either.

2. How would I test this?

pmatseykanets:
@ericpromislow 

> Where is <some-object>.FailurePolicy used as an r-value?
Handled by k8s. https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy

> How would I test this?
E.g. as shown in the integration test included in this PR:
- Scale webhook to 0
- Try to create or update a setting.
- With the FailurePolicy set to `Ignore` it will succeed whereas with `Fail` it will reject the request

ericpromislow (approved):
The changes make sense, and are actually in line with how I see other validators set the `FailurePolicy` to `Ignore` (like cluster/validator, clusterproxyconfig/validator, feature/validator and podsecurityadmissionconfigurationtemplate).

Verified by running `rancher/rancher:v2.8.3`, manually setting the webhook value to `Ignore`, stopping v2.8.3, and running `rancher/rancher:v2.8.4`

PR #449, opened 2024-07-12
Add workflow to release webhook in rancher/charts and rancher/rancher

tomleb:
# Issue https://github.com/rancher/rancher/issues/46704

# What is this

This adds two workflows that can be triggered manually to bump webhook RCs to rancher/charts and rancher/rancher.

Creating a new release in rancher/charts and rancher/rancher will be simple:
1. Go to the rancher/webhook repo
2. Click on the `Actions` tab
3. On the left sidebar, click on the workflow that you want to run.
4. Enter the necessary input. Example in the image below.
image
5. Click `Run workflow`

This will do the following.

**charts**

Create a new branch off of the branch you input, make some changes to bump webhook, push to rancher/charts directly (more details below), create a pull request to the target branch. You can see an example here: https://github.com/rancher/charts/pull/4211.

**rancher**

Create a new branch off of the branch you input, update webhook version in build.yaml, run go generate, push rancher/rancher, create a pull request to the target branch. You can see an example here: https://github.com/tomleb/rancher-rancher/pull/8.

# Support

Currently, this supports the following:
- Bumping from one RC to another
- Bumping from one RC to unRC

It does not yet support adding an entirely new version of webhook. I'd be okay with implementing it for this PR if asked. 

# Benefits

- No need to open a terminal to do this work (tedious)
- Faster to create PRs: I made sure to include some relevant information to the PR description when they are opened. Here's an example:
image. I'm lazy so I'd rather not have to fetch this information every time like I've been doing and it should make it easier for reviewer as well to review this.
- Easier to review due to the above PRs containing lots of information and links
- This will also mean that we're going to need only 2 people for an RC bump because the PR will be created by the bot. So the person running the workflow + another person will be enough to review a bump.

# Pushing to rancher/charts & rancher/rancher

EIO provided us with a GH app that can be used (through vault) to create tokens for rancher/rancher and rancher/charts. This allows us to push to those repositories directly and create PRs.

# Credits

A lot of it was taken from the fleet team:
- https://github.com/rancher/fleet/blob/main/.github/workflows/release-against-charts.yml
- https://github.com/rancher/fleet/blob/main/.github/workflows/release-against-rancher.yml
- https://github.com/rancher/fleet/blob/main/.github/scripts/release-against-charts.sh
- https://github.com/rancher/fleet/blob/main/.github/scripts/release-against-rancher.sh

Some notable changes:
- Removed support from unRC and creating a new release entirely. Mostly to simplify, let's start with just bumping RCs which is what we end up spending most of our times on.
- Removed the need to input the chart version number (`104.0.0` in `104.0.0+up0.5.0-rc13`). Instead, we just reuse the one for that release.
- Removed the usage of the [peter-evans/create-pull-request action](https://github.com/peter-evans/create-pull-request). This has been explicitly rejected by the security team so we're not allowed to use it. Instead, using `gh` cli tool is good enough.
- More information in the PR description generated as mentioned above.

### DIFF
--- a/.github/workflows/release-charts.yaml
+++ b/.github/workflows/release-charts.yaml
@@ -24,6 +24,7 @@ jobs:
   create-charts-pr:
     runs-on: ubuntu-latest
     permissions:
+      contents: read
       # Required for vault
       id-token: write
     steps:
@@ -60,9 +61,9 @@ jobs:
       - name: Configure the committer
         run: |
           cd charts
-          user_id=$(gh api "/users/$APP_USER" --jq .id)"
+          user_id=$(gh api "/users/$APP_USER" --jq .id)
           git config --global user.name "$APP_USER"
-          git config --global user.email "${user_id}+${APP_USER}@users.noreply.github.com>"
+          git config --global user.email "${user_id}+${APP_USER}@users.noreply.github.com"
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
           APP_USER: "${{ steps.app-token.outputs.app-slug }}[bot]"
--- a/.github/workflows/release-charts.yaml
+++ b/.github/workflows/release-charts.yaml
@@ -1,17 +1,17 @@
-name: Bump Webhook RC against rancher/charts
+name: Bump webhook in rancher/charts
 on:
   workflow_dispatch:
     inputs:
       charts_ref:
-        description: "Submit PR against the following rancher/charts branch (eg: dev-v2.9)"
+        description: "Submit PR against the following rancher/charts branch (eg: dev-v2.10)"
         required: true
-        default: "dev-v2.9"
+        default: "dev-v2.10"
       prev_webhook:
-        description: "Previous Webhook version (eg: 0.5.0-rc13)"
+        description: "Previous Webhook version (eg: 0.5.0-rc.13)"
         required: true
         default: ""
       new_webhook:
-        description: "New Webhook version (eg: 0.5.0-rc14)"
+        description: "New Webhook version (eg: 0.5.0-rc.14)"
         required: true
         default: ""
 
@@ -21,54 +21,75 @@ env:
   NEW_WEBHOOK: ${{ github.event.inputs.new_webhook }}
 
 jobs:
-  create-rancher-charts-pr:
+  create-charts-pr:
     runs-on: ubuntu-latest
+    permissions:
+      # Required for vault
+      id-token: write
     steps:
       - uses: actions/checkout@v4
         with:
           ref: ${{ github.event.inputs.ref }}
           path: webhook
 
-      # Checkout the fork first because it's simpler to setup the token for the
-      # push later on
-      - name: Checkout fork of rancher/charts
+      - uses: rancher-eio/read-vault-secrets@main
+        with:
+          secrets: |
+            secret/data/github/repo/${{ github.repository }}/github/app-credentials appId | APP_ID ;
+            secret/data/github/repo/${{ github.repository }}/github/app-credentials privateKey | PRIVATE_KEY
+
+      # Fetch github token just for the charts repository
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ env.APP_ID }}
+          private-key: ${{ env.PRIVATE_KEY }}
+          repositories: |
+            charts
+
+      - name: Checkout charts repository
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          repository: tomleb/rancher-charts
-          token: ${{ secrets.PUSH_TO_FORKS_TOKEN }}
+          repository: ${{ github.repository_owner }}/charts
+          token: ${{ steps.app-token.outputs.token }}
           path: charts
+          # Allow making git push request later on
+          persist-credentials: true
+
+      - name: Configure the committer
+        run: |
+          cd charts
+          user_id=$(gh api "/users/$APP_USER" --jq .id)"
+          git config --global user.name "$APP_USER"
+          git config --global user.email "${user_id}+${APP_USER}@users.noreply.github.com>"
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          APP_USER: "${{ steps.app-token.outputs.app-slug }}[bot]"
 
       - name: Install dependencies
         run: sudo snap install yq --channel=v4/stable
 
-      - name: Setup repo dir
-        run: |
-          cd charts
-          BRANCH="gha-test-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
-          echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
-          git config --global user.email "webhook@example.com"
-          git config --global user.name "tomleb webhook Bot"
-          git remote add upstream https://github.com/rancher/charts
-          git fetch upstream
-          git checkout -b $BRANCH "upstream/$CHARTS_REF"
-
       - name: Run release script
         run: |
-          ./webhook/.github/workflows/scripts/release-against-charts.sh charts "$PREV_WEBHOOK" "$NEW_WEBHOOK"
+          cd charts
+          BRANCH="bump-webhook-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
+          git checkout -b "$BRANCH" "$CHARTS_REF"
+          ./webhook/.github/workflows/scripts/release-against-charts.sh . "$PREV_WEBHOOK" "$NEW_WEBHOOK"
 
       - name: Push and create pull request
         env:
-           GITHUB_TOKEN: ${{ secrets.PUSH_TO_FORKS_TOKEN }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
           cd charts
-          git push origin $BRANCH
+          git push origin "$BRANCH"
 
           body=$(../webhook/.github/workflows/scripts/release-message.sh "$PREV_WEBHOOK" "$NEW_WEBHOOK")
 
           gh pr create \
             --title "[$CHARTS_REF] Bump rancher-webhook to v$NEW_WEBHOOK" \
             --body "$body" \
-            --repo rancher/charts \
-            --head "tomleb:$BRANCH" \
+            --repo ${{ github.repository_owner }}/charts \
+            --head "${{ github.repository_owner }}:$BRANCH" \
             --base "$CHARTS_REF"
--- a/.github/workflows/release-charts.yaml
+++ b/.github/workflows/release-charts.yaml
@@ -7,11 +7,11 @@ on:
         required: true
         default: "dev-v2.10"
       prev_webhook:
-        description: "Previous Webhook version (eg: 0.5.0-rc.13)"
+        description: "Previous Webhook version (eg: v0.5.0-rc.13)"
         required: true
         default: ""
       new_webhook:
-        description: "New Webhook version (eg: 0.5.0-rc.14)"
+        description: "New Webhook version (eg: v0.5.0-rc.14)"
         required: true
         default: ""
 
@@ -88,7 +88,7 @@ jobs:
           body=$(../webhook/.github/workflows/scripts/release-message.sh "$PREV_WEBHOOK" "$NEW_WEBHOOK")
 
           gh pr create \
-            --title "[$CHARTS_REF] Bump rancher-webhook to v$NEW_WEBHOOK" \
+            --title "[$CHARTS_REF] Bump rancher-webhook to $NEW_WEBHOOK" \
             --body "$body" \
             --repo ${{ github.repository_owner }}/charts \
             --head "${{ github.repository_owner }}:$BRANCH" \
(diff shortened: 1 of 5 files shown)


.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Install dependencies
+        run: sudo snap install yq --channel=v4/stable
+
+      - name: Setup repo dir
+        run: |
+          cd charts
+          BRANCH="gha-test-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
+          git config --global user.email "webhook@example.com"
+          git config --global user.name "tomleb webhook Bot"
```
tomleb, line 46: Will change this when we have token to something less personal

.github/workflows/release-rancher.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Install dependencies
+        run: sudo snap install yq --channel=v4/stable
+
+      - name: Setup repo dir
+        run: |
+          cd rancher
+          BRANCH="gha-test-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
+          git config --global user.email "webhook@example.com"
+          git config --global user.name "tomleb webhook Bot"
```
tomleb, line 46: Will change this when we have token to something less personal

.github/workflows/release-rancher.yaml
```diff
@@ -0,0 +1,71 @@
...
+          git status
+          git push origin $BRANCH
+
+          body=$(../webhook/.github/workflows/scripts/release-message.sh ${{github.event.inputs.prev_webhook}} ${{github.event.inputs.new_webhook}})
+
+          gh pr create \
+            --title "[${{github.event.inputs.rancher_ref}}] Test GHA: Bump rancher-webhook to v${{github.event.inputs.new_webhook}}" \
+            --body "$body" \
+            --repo tomleb/rancher-rancher \
+            --head "tomleb:$BRANCH" \
```
tomleb, line 68: Will change this when we have token to probably rancherbot

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,71 @@
...
+          cd charts
+          git push origin $BRANCH
+
+          body=$(../webhook/.github/workflows/scripts/release-message.sh ${{github.event.inputs.prev_webhook}} ${{github.event.inputs.new_webhook}})
+
+          gh pr create \
+            --title "[${{github.event.inputs.rancher_ref}}] Test GHA: Bump rancher-webhook to v${{github.event.inputs.new_webhook}}" \
+            --body "$body" \
+            --repo rancher/charts \
+            --head "tomleb:$BRANCH" \
```
tomleb, line 68: Will change this when we have token to probably rancherbot

.github/workflows/release-charts.yaml
```diff
@@ -0,0 +1,69 @@
...
+          echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
+          git config --global user.email "webhook@example.com"
+          git config --global user.name "tomleb webhook Bot"
+          git remote add upstream https://github.com/rancher/charts
+          git fetch upstream
+          git checkout -b $BRANCH upstream/${{github.event.inputs.charts_ref}}
+
+      - name: Run release script
+        run: |
+          ./webhook/.github/workflows/scripts/release-against-charts.sh charts ${{github.event.inputs.prev_webhook}} ${{github.event.inputs.new_webhook}}
```
ericpromislow, line 53: Feel free to correct me if I'm wrong, but
`prev_webhook` and `new_webhook` smell too much like user-supplied input and unless you know they can't contain semi-colons and other control characters, should probably be passed to `release-against-charts.sh` as environment vars

.github/workflows/release-charts.yaml
ericpromislow: I'm also wondering about script-injection problems here. What happens if you managed to set `body` to `foo " ; not-a-command-so-we-should-see-an-error-message ; " bar` ?

tomleb: Discussed on slack but I'll summarize here.

For something like `body`, where I do `--body "$body"`, there isn't any problem if body contains a `"` (as an example). Here's a simple poc in bash:
```bash
test() {
	echo "$2"
}

body="\" && echo hi"
// output: " && echo hi
test "foo" "$body"
```

However there are issues the way `--title` is currently set. In that case, it would be possible to do injection (either by mistake or for malicious reasons).

I've found two sources about those kind of injections:
- https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#good-practices-for-mitigating-script-injection-attacks
- https://www.kenmuse.com/blog/github-actions-injection-attacks/ 

It seems like the best practice is to set environment variables with those GH variables (`${{ foo }}`). I'll work on doing this.

tomleb: Updated to use environment variables instead. Let me know what you think.

.github/workflows/release-charts.yaml
tomleb: Updated to use environment variables instead, let me know what you think!

ericpromislow:
Regarding the env vars, once this is out of draft and ready to play with I'll see if I can trigger an exploit

.github/workflows/release-rancher.yaml
```diff
@@ -0,0 +1,95 @@
...
+          # Allow making git push request later on
+          persist-credentials: true
+
+      - name: Configure the committer
+        run: |
+          cd rancher
+          user_id=$(gh api "/users/$APP_USER" --jq .id)"
+          git config --global user.name "$APP_USER"
+          git config --global user.email "${user_id}+${APP_USER}@users.noreply.github.com>"
+        env:
```
tomleb, line 66: Adapted from https://github.com/actions/create-github-app-token, unsure how necessary this is, probably not, but I guess that's probably the "idiomatic" values to put there so why not

tomleb:
Updated the workflow so they use a github apps credential provided by EIO instead. No need for a fork anymore, can directly push to our repos.

.github/workflows/scripts/release-against-charts.sh
```diff
@@ -0,0 +1,66 @@
...
+set -ue
+
+pushd "${CHARTS_DIR}" > /dev/null
+
+# Validate the given webhook version (eg: 0.5.0-rc.13)
+if ! grep -q "${PREV_WEBHOOK_VERSION}" ./packages/rancher-webhook/package.yaml; then
+    echo "Previous Webhook version references do not exist in ./packages/rancher-webhook/. The content of the file is:"
+    cat ./packages/rancher-webhook/package.yaml
+    exit 1
+fi
```
tomleb, line 36: We could also avoid requiring the old version of webhook and instead fetch it directly from that same file. I'd be okay with that. This way you'd just need to know the version you want to go to.

.github/workflows/scripts/release-against-charts.sh
```diff
@@ -0,0 +1,66 @@
...
+Usage:
+  $0 <path to charts repo> <prev webhook release> <new webhook release>
+EOF
+}
+
+if [ -z "$CHARTS_DIR" ] || [ -z "$PREV_WEBHOOK_VERSION" ] || [ -z "$NEW_WEBHOOK_VERSION" ]; then
+    usage
+    exit 1
+fi
+
```
pmatseykanets, line 26: We may want to check that `$PREV_WEBHOOK_VERSION != $NEW_WEBHOOK_VERSION`

pmatseykanets (approved):

.github/workflows/scripts/release-against-rancher.sh
```diff
@@ -0,0 +1,56 @@
...
+fi
+NEW_CHART_VERSION=$PREV_CHART_VERSION
+
+yq --inplace ".webhookVersion = \"${NEW_CHART_VERSION}+up${NEW_WEBHOOK_VERSION}\"" ./build.yaml
+
+# Downloads dapper
+make .dapper
+
+# DAPPER_MODE=bind will make sure we output everything that changed
+DAPPER_MODE=bind ./.dapper go generate ./... || true
```
pmatseykanets, line 50: Do we need to `go generate` everything here or intent was only to call `go run ./pkg/codegen/buildconfig/...`?

pmatseykanets (dismissed):

.github/workflows/scripts/release-against-rancher.sh
tomleb: I could extract the commands that `go generate` runs but I think it's more future proof to use `go generate` directly. The tradeoff is that this takes longer to run (which also means preventing other jobs to run).

.github/workflows/scripts/release-message.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
```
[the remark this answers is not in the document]
tomleb: Yeah, added in 0d17e9c57f3991ad948082a121fef4444c877285.

.github/workflows/scripts/release-against-rancher.sh
```diff
@@ -0,0 +1,56 @@
...
+DAPPER_MODE=bind ./.dapper rm -rf go
```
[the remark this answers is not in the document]
tomleb: Discussed this on slack but for visibility:

`go generate` might use a few tools like `go` and `controller-gen`. Those tools version are currently only defined in the Dockerfile for dapper in r/r. I want to be able to generate using the right tools, so using dapper for now is convenient because we don't need to care about those tools, no need to keep maintaining and syncing them with r/r, etc.

I don't really have bandwidth for making a change in r/r to allow that (eg: by extracting the versions of tools outside).

.github/workflows/scripts/release-against-charts.sh
tomleb: Addressed in a7c70c3299b2f955c33ce7b01873c1e63a0d5df4.

.github/workflows/scripts/release-against-rancher.sh
pmatseykanets: It also has potential of failing doing the irrelevant work.

.github/workflows/scripts/release-against-charts.sh
```diff
@@ -0,0 +1,75 @@
...
+git add packages/rancher-webhook
+git commit -m "Bump rancher-webhook to $NEW_WEBHOOK_VERSION"
+
+PACKAGE=rancher-webhook make charts
+git add ./assets/rancher-webhook ./charts/rancher-webhook
+git commit -m "make charts"
+
+CHART=rancher-webhook VERSION=${PREV_CHART_VERSION}+up${PREV_WEBHOOK_VERSION_SHORT} make remove
+git add ./assets/rancher-webhook ./charts/rancher-webhook ./index.yaml
+git commit -m "make remove"
```
nflynt, line 66: For new RC builds where the PREV_WEBHOOK_VERSION is not an `rc` tag, this would end up inadvertently blasting out the release chart. We should almost certainly check for and explicitly skip this step (maybe disallow the whole run?), lest someone unfamiliar with the workflow miss that this has occurred and delete a production build of the chart.

nflynt (changes requested):
Looks mostly okay to me. If we can make the new RC handling a tiny bit more robust I think I have no other changes to add. Mostly we want to make it impossible/difficult for newly onboarded engineers to accidentally blast build artifacts. :)

.github/workflows/scripts/release-against-rancher.sh
tomleb: I'll leave as is for now.

tomleb:
Updated the release against charts script to support going from stable release to new rc (eg: v0.5.2 to v0.5.3-rc.1). I have tested it locally and that seems to work fine.

nflynt (approved):
This looks good to me. Now that the RC guards are in place, nothing's really jumping out. We can improve it further as we go. :+1:

.github/workflows/scripts/release-against-charts.sh
```diff
@@ -0,0 +1,110 @@
...
+EOF
+}
+
+bump_patch() {
+    version=$1
+    major=$(echo "$version" | cut -d. -f1)
+    minor=$(echo "$version" | cut -d. -f2)
+    patch=$(echo "$version" | cut -d. -f3)
+    new_patch=$((patch + 1))
+    echo "${major}.${minor}.${new_patch}"
```
ericpromislow, line 28: If perl is available, this is a one-liner:
```
echo $1 | perl -pe 's/(.*)(\d+)$/$1 . ($2 + 1)/e'
```

tomleb: Yeah it's also possible to do a one liner with awk, eg:

```sh
echo "v1.2.3" | awk -F. '{$3=$3+1;print}' OFS=.
```

but I opted to go with simple commands that I think more people are familiar with (eg: `cut` and addition).

ericpromislow: Disagreements accepted. Because of its heritage Perl is ubiquitous, but the only reason I would ever use it now is for its substitute-with-eval feature. I also learned last week that Perl is still on version 5, which shipped about 30 years ago, and probably always will be.

tomleb:
Rebased, fixed an issue with quotes, that's pretty much it. Ready for a re-review.

crobby (approved):
The bash looks really good and actions are best experienced rather than thought about.  
lgtm, let's see it in action

nflynt (approved):

PR #451, opened 2024-07-18
[v0.5] Use the proper config when fetching mutating webhook configuration

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/42611

## Problem

A wrong variable is used when fetching the mutating webhook configuration, that may lead to the following errors on startup:
```
time="2024-07-18T14:01:43Z" level=error msg="Failed to ensure configuration: failed to create mutating configuration: MutatingWebhookConfiguration.admissionregistration.k8s.io \"\" is invalid: metadata.name: Required value: name or generateName is required"
time="2024-07-18T14:01:43Z" level=error msg="error syncing 'cattle-system/cattle-webhook-ca': handler secrets: failed to create mutating configuration: MutatingWebhookConfiguration.admissionregistration.k8s.io \"\" is invalid: metadata.name: Required value: name or generateName is required, requeuing"
```

## Solution

Use the right/intended variable.

## CheckList
  
- [x] Test
  
- [ ] Docs

### DIFF
--- a/.github/workflows/scripts/integration-test-ci
+++ b/.github/workflows/scripts/integration-test-ci
@@ -40,6 +40,9 @@ set -e
 echo "Shutting down core rancher"
 kubectl scale deploy rancher -n cattle-system --replicas=0
 kubectl wait pods -l app=rancher --for=delete -n cattle-system
+# Make sure the webhook recreates configurations on startup
+kubectl delete validatingwebhookconfiguration rancher.cattle.io
+kubectl delete mutatingwebhookconfiguration rancher.cattle.io
 
 echo "Uploading new webhook image"
 
--- a/scripts/ci
+++ b/scripts/ci
@@ -3,10 +3,10 @@ set -e
 
 cd $(dirname $0)
 
-./validate
-./validate-ci
 ./build
 ./test
+./validate
+./validate-ci
 ./package
 ./package-helm
 ./test-helm
--- a/scripts/ci
+++ b/scripts/ci
@@ -3,10 +3,10 @@ set -e
 
 cd $(dirname $0)
 
-./build
-./test
 ./validate
 ./validate-ci
+./build
+./test
 ./package
 ./package-helm
 ./test-helm
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -262,9 +262,9 @@ func (s *secretHandler) ensureWebhookConfiguration(validatingConfig *v1.Validati
 		}
 	}
 
-	currMutation, err := s.mutatingController.Get(validatingConfig.Name, metav1.GetOptions{})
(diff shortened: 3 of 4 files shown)


JonCrowther (dismissed):

nflynt (dismissed):
I **wondered** why only one of these was getting created when I was doing a docker upgrade reproduction earlier. This is probably why! :+1:

ericpromislow:
I independently came up with the same solution. Was about to figure out how to write an integration test, but given the approvals I assume it's not needed.

Related to issue [#46189](https://github.com/rancher/rancher/issues/46189)

nflynt:
Admittedly I did not think about integration tests, but those are a _really good idea_ since this issue is causing problems in the wild. (It confounds workarounds for the recent docker upgrade kerfuffle, among other things.) We should probably not merge until we have those; I'll gladly re-review when we do.

scripts/ci
```diff
@@ -3,10 +3,10 @@ set -e
 
 cd $(dirname $0)
 
-./build
-./test
```
pmatseykanets, line 7: Arguably static analysis and other forms of validation should come before building and running tests.

ericpromislow (approved):
And the update to the test looks fine

scripts/ci
```diff
@@ -3,10 +3,10 @@ set -e
...
 ./validate
 ./validate-ci
+./build
```
ericpromislow, line 8: This change looks like it's breaking GHA, although I don't see why.

ericpromislow (changes requested):
The amd64 tests are breaking when the validators happen after `build` and `test`. This is probably fixable, but it's not required for this PR.

scripts/ci
pmatseykanets: Agreed, reverted. We can look at this outside of this PR.

.github/workflows/scripts/integration-test-ci
```diff
@@ -40,6 +40,9 @@ set -e
 echo "Shutting down core rancher"
 kubectl scale deploy rancher -n cattle-system --replicas=0
 kubectl wait pods -l app=rancher --for=delete -n cattle-system
+# Make sure the webhook recreates configurations on startup
+kubectl delete validatingwebhookconfiguration rancher.cattle.io
+kubectl delete mutatingwebhookconfiguration rancher.cattle.io
```
pmatseykanets, line 45: This is to make sure we exercise `ensureWebhookConfiguration` when the webhook starts for unit tests.

ericpromislow (approved):
Looks fine. Why the validators have to happen in the CI after building and testing needs to be left for another PR

nflynt (approved):
:+1:

PR #456, opened 2024-07-30
Stop testing the 'ok' part of a map access.

ericpromislow:
Verify that the actual value of the annotation is string "true" when checking that it's ok to change the tls-mode from `system-store` to `strict`.

Related to issue [#46189](https://github.com/rancher/rancher/issues/46189)


ericpromislow:
`TestIntegrationTest/TestClusterProxyConfig` is a flaky test. It failed first time on amd64 but not arm64 on this branch. I manually ran both on `release/v0.5`, and it passed on amd64, failed on arm64. But the problem isn't introduced on this branch.

ericpromislow:
Dup of PR #451

MbolotSuse (approved):

PR #468, opened 2024-08-07
Current dev work is now for rancher 2.10

ericpromislow:
## Issue: 

No related issue. Now that we're doing trunk-based development, the version in the trunk has to specify the version that current trunk work is targeting.


MbolotSuse (approved):
Thanks for making this change.

tomleb (approved):

PR #469, opened 2024-08-08
WIP - test for race condition

ericpromislow:
# DO NOT MERGE

- I can't run these tests locally right now. So using the CI to run them and check log output


tests/integration/clusterProxyConfig_test.go
```diff
@@ -26,6 +28,36 @@ func (m *IntegrationSuite) TestClusterProxyConfig() {
 	// Creating the first CPC should succeed
 	err = client.Create(ctx, validCreateObj.Namespace, validCreateObj, result, metav1.CreateOptions{})
 	m.NoError(err, "Error returned during the creation of a valid clusterProxyConfig")
+
+	// Verify the API knows about the first object before we try to create a second one.
+	// Wait 2 minutes for the object to be created (should be more than ample time)
```
nflynt, line 33: nit: the comment states 2 minutes, but the actual code waits for 5
second nit: 5 minutes is a long time. I kinda prefer that to the 30+ minutes it takes to restart the entire testing suite, but is there no better way? could we poll for the API to be aware of the object in a loop with smaller sleeps or something?

tests/integration/clusterProxyConfig_test.go
```diff
@@ -26,6 +28,36 @@ func (m *IntegrationSuite) TestClusterProxyConfig() {
...
+	}
+	watcher, err := client.Watch(ctx, validCreateObj.Namespace, listOptions)
+	m.NoError(err, "Error returned trying to watch the clusterProxyConfig")
+	if watcher != nil {
+		// If `m.NoError` fails, code still keeps running and we have a nil watcher
+		defer watcher.Stop()
+		for {
+			receivedEvent, ok := <-watcher.ResultChan()
+			if !ok {
+				logrus.Infof("Got a non-ok event from the watcher; breaking")
```
nflynt, line 47: Should a non-ok event here fail the test?

tests/integration/clusterProxyConfig_test.go
```diff
@@ -26,6 +28,36 @@ func (m *IntegrationSuite) TestClusterProxyConfig() {
...
+				break
+			}
+			if receivedEvent.Object.GetObjectKind().GroupVersionKind().Kind == "ClusterProxyConfig" {
+				err = client.Get(ctx, validCreateObj.Namespace, cpcName, result, metav1.GetOptions{})
+				if err == nil {
+					break
```
nflynt, line 53: likewise, if we fail to parse the watched ClusterProxyConfig, shouldn't that also fail the test?

tests/integration/clusterProxyConfig_test.go
ericpromislow: At least I added math to the code to make it clearer how long we're willing to wait.

tests/integration/clusterProxyConfig_test.go
ericpromislow: Yes, and I figured out what to do.

tests/integration/clusterProxyConfig_test.go
ericpromislow: Yes

nflynt (approved):
Excellent, that test logic is _much_ cleaner. :+1:

PR #471, opened 2024-08-12
Check that {disable|delete}-inactive-user-after setting is not less than auth-user-session-ttl-minutes

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/46648

## Problem

It doesn't make sense to have `{disable|delete}-inactive-user-after` setting values less than  `auth-user-session-ttl-minutes` as users won't have a chance to log in outside of manually/deliberately re-authenticating.

## Solution

Validate that `{disable|delete}-inactive-user-after` is not less than `auth-user-session-ttl-minutes`.

## CheckList
  
- [x] Test
  
- [x] Docs


pkg/resources/management.cattle.io/v3/setting/validator.go
```diff
@@ -96,20 +99,58 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+	return time.Duration(minutes) * time.Minute, nil
+}
+
+func (a *admitter) isLessThanUserSessionTTL(dur time.Duration) bool {
+	authUserSessionTTLDuration, getErr := a.getAuthUserSessionTTLDuration()
+	if getErr != nil {
+		logrus.Warnf("failed to get auth-user-session-ttl-minutes setting: %v", getErr) // Log the error and move on.
+	}
+
+	return dur < authUserSessionTTLDuration
```
pmatseykanets, line 126: We may potentially want to make sure the `dur` is reasonably larger than the `auth-user-session-ttl-minutes` (e.g. 2x). 🤔

crobby: Having a dur that is 1 min longer than ttl isn't likely to be useful, but it's not something I feel passionate about when it comes to "we must enforce this".

crobby (approved):

JonCrowther (approved):

tomleb (approved):
LGTM.

Fyi, it will be possible for users to update `auth-user-session-ttl-minutes` to a value such that `disable-inactive-user-after` and `delete-inactive-user-after` is then lower than `auth-user-session-ttl-minutes`. Not sure if that's something you've considered / is important for your use case.

pkg/resources/management.cattle.io/v3/setting/validator.go
```diff
@@ -69,71 +82,243 @@ func (v *Validator) Admitters() []admission.Admitter {
...
+func parseMinutes(value string) (time.Duration, error) {
+	minutes, err := strconv.ParseInt(value, 10, 64)
+	if err != nil {
+		return 0, err
+	}
+
+	return time.Duration(minutes) * time.Minute, nil
+}
+
+func (a *admitter) validateAgentTLSMode(oldSetting, newSetting *v3.Setting) error {
```
pmatseykanets, line 321: This change is deliberate to avoid unnecessarily dereferencing Setting objects.
The only case when `oldSetting` and `newSetting` can be `nil` is when `objectsv3.SettingOldAndNewFromRequest()` returns an err, which is checked.

pkg/resources/management.cattle.io/v3/setting/validator.go
```diff
@@ -155,33 +340,25 @@ func (a *admitter) validateAgentTLSMode(oldSetting, newSetting v3.Setting) error
...
 			return true
 		}
 	}
+
 	return false
 }
 
-func effectiveValue(s v3.Setting) string {
+// effectiveValue returns the effective value of the setting.
+func effectiveValue(s *v3.Setting) string {
```
pmatseykanets, line 354: This change is deliberate to avoid unnecessarily dereferencing Setting objects. See above.

pkg/resources/management.cattle.io/v3/setting/validator.go
```diff
@@ -69,71 +82,243 @@ func (v *Validator) Admitters() []admission.Admitter {
...
+	if s.Value == "" {
+		return nil
+	}
+
+	userSessionDuration, err := parseMinutes(s.Value)
+	if err != nil {
+		return field.TypeInvalid(valuePath, s.Value, err.Error())
+	}
+	if userSessionDuration < 1 {
+		return field.TypeInvalid(valuePath, s.Value, "negative value")
```
tomleb, line 167: nit: might confuse people that reads `negative` as meaning strictly less than zero. Also I think it's more common to see `<= 0` instead of `< 1`.

```suggestion
		return field.TypeInvalid(valuePath, s.Value, "non-positive value")
```

pkg/resources/management.cattle.io/v3/setting/validator_test.go
```diff
@@ -37,151 +37,504 @@ var (
...
+
+func (s *SettingSuite) TestValidateDeleteInactiveUserAfterOnCreate() {
+	s.validateDeleteInactiveUserAfter(v1.Create)
 }
-func (t *retentionTest) toOldSetting() ([]byte, error) {
-	return json.Marshal(v3.Setting{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: t.setting,
+
+func (s *SettingSuite) validateDeleteInactiveUserAfter(op v1.Operation) {
```
tomleb, line 143: Could use a test for explicit `0s`

pkg/resources/management.cattle.io/v3/setting/validator_test.go
```diff
@@ -37,151 +37,504 @@ var (
...
-	return t.setting + "_" + t.value
+func (s *SettingSuite) TestValidateDisableInactiveUserAfterOnCreate() {
+	s.validateDisableInactiveUserAfter(v1.Create)
 }
 
-func (t *retentionTest) toSetting() ([]byte, error) {
-	return json.Marshal(v3.Setting{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: t.setting,
+func (s *SettingSuite) validateDisableInactiveUserAfter(op v1.Operation) {
```
tomleb, line 48: Could use a test for explicit `0s`

tomleb (commented):
As discussed, I'm curious if the validation for auth-user-session-ttl-minutes can be considered a breaking change. Other than that, the change in how to deal with setting is appreciated.

pkg/resources/management.cattle.io/v3/setting/validator.go
tomleb: Resolving as we're allowing 0 now.

pmatseykanets: Agreed. But it's back to being strictly checking for a negative.

docs.md
```diff
@@ -311,25 +311,20 @@ RoleTemplate can not be deleted if they are referenced by other RoleTemplates vi
...
-When a Setting is created, the following checks take place:
+When settings are created or updated, the following common checks take place:
 
-- If set, `disable-inactive-user-after` must be zero or a positive duration and can't be less than `auth-user-session-ttl-minutes` (e.g. `240h`).
-- If set, `delete-inactive-user-after` must be zero or a positive duration and can't be less than `336h` or `auth-user-session-ttl-minutes`, whichever is greater (e.g. `336h`).
+- If set, `disable-inactive-user-after` must be zero or a positive duration (e.g. `240h`).
+- If set, `delete-inactive-user-after` must be zero or a positive duration and can't be less than `336h` (e.g. `336h`).
 - If set, `user-last-login-default` must be a date time according to RFC3339 (e.g. `2023-11-29T00:00:00Z`).
 - If set, `user-retention-cron` must be a valid standard cron expression (e.g. `0 0 * * 0`).
+- The `auth-user-session-ttl-minutes` .
```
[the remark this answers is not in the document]
pmatseykanets: 👍 Duh, edit remnants. Removed.

pkg/resources/management.cattle.io/v3/setting/validator_test.go
pmatseykanets: Agreed. Done.

pkg/resources/management.cattle.io/v3/setting/validator_test.go
pmatseykanets: Agreed. Done.

tomleb (approved):

pmatseykanets:
> As discussed, I'm curious if the validation for auth-user-session-ttl-minutes can be considered a breaking change.
Agreed, this needs to be discussed with a larger group.

pmatseykanets:
image





image
image
image

tomleb (approved):

crobby (approved):

PR #474, opened 2024-08-14
[46381] [2.10] Add validation to data directories on creation

jakefhyde:
## Problem

Data directories were not validated on creation. The should be validated based on the following criteria:
- Absolute paths (start with /)
- Clean (not contain env vars, shell expressions, ., or ..)
- Not set to the same path (no duplicates are allowed between any pairs of paths out of 3 configurable)
- Not nested one within another

## Solution

Implement the above checks with the provisioning cluster validator

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -379,12 +379,24 @@ When a UserAttribute is updated, the following checks take place:
 
 ### Validation Checks
 
+#### On Create
+
+##### Data Directories
+
+Prevent the creation of new objects with an env var (under `spec.agentEnvVars`) with a name of `CATTLE_AGENT_VAR_DIR`.
+Prevent the creation of new objects with an invalid data directory. An invalid data directory is defined as the 
+following:
+- Is not an absolute path (i.e. does not start with `/`)
+- Attempts to include environment variables (e.g. `$VARIABLE` or `${VARIABLE}`)
+- Attempts to include shell expressions (e.g. `$(command)` or `` `command` ``)
+- Equal to another data directory
+- Attempts to nest another data directory
+
 #### On Update
 
 ##### Data Directories
 
-Prevent the creation of new objects with an env var (under `spec.agentEnvVars`) with a name of `CATTLE_AGENT_VAR_DIR`. 
-On update, also prevent new env vars with this name from being added but allow them to be removed. Rancher will perform 
+On update, prevent new env vars with this name from being added but allow them to be removed. Rancher will perform 
 a one-time migration to move the system-agent data dir definition to the top level field from the `AgentEnvVars` 
 section. A secondary validator will ensure that the effective data directory for the `system-agent` is not different 
 from the one chosen during cluster creation. Additionally, the changing of a data directory for the `system-agent`, 
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/Cluster.md
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/Cluster.md
@@ -1,11 +1,23 @@
 ## Validation Checks
 
+### On Create
+
+#### Data Directories
+
+Prevent the creation of new objects with an env var (under `spec.agentEnvVars`) with a name of `CATTLE_AGENT_VAR_DIR`.
+Prevent the creation of new objects with an invalid data directory. An invalid data directory is defined as the 
+following:
+- Is not an absolute path (i.e. does not start with `/`)
+- Attempts to include environment variables (e.g. `$VARIABLE` or `${VARIABLE}`)
+- Attempts to include shell expressions (e.g. `$(command)` or `` `command` ``)
+- Equal to another data directory
+- Attempts to nest another data directory
+
 ### On Update
 
 #### Data Directories
 
-Prevent the creation of new objects with an env var (under `spec.agentEnvVars`) with a name of `CATTLE_AGENT_VAR_DIR`. 
-On update, also prevent new env vars with this name from being added but allow them to be removed. Rancher will perform 
+On update, prevent new env vars with this name from being added but allow them to be removed. Rancher will perform 
 a one-time migration to move the system-agent data dir definition to the top level field from the `AgentEnvVars` 
 section. A secondary validator will ensure that the effective data directory for the `system-agent` is not different 
 from the one chosen during cluster creation. Additionally, the changing of a data directory for the `system-agent`, 
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
@@ -5,8 +5,10 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net/http"
+	"path/filepath"
 	"regexp"
 	"slices"
+	"strings"
 
 	v1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
 	rkev1 "github.com/rancher/rancher/pkg/apis/rke.cattle.io/v1"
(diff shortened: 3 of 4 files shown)


brandond (commented):
lgtm!

Request: can we also validate that they don't have a trailing slash? Recent updates to the killall script to properly kill pods when the data dir does not end with /rke2 or /k3s require that the custom data dir does not have a trailing slash. We could probably fix enhance the killall scripts to normalize the data dir environment variable, but it'd be great to enforce it here as well.

brandond (commented):
Another thought: how does this work on clusters with Windows nodes, where the data dir will have a drive letter prefix? Does the current custom data dir support allow for different paths for windows and Linux nodes? Probably should have covered that in the ADR.

jakefhyde:
> Another thought: how does this work on clusters with Windows nodes, where the data dir will have a drive letter prefix? Does the current custom data dir support allow for different paths for windows and Linux nodes? Probably should have covered that in the RFC.

@HarrisonWAffel Feel free to chime in (or to correct me if I get any of this wrong) but to my understanding, Windows will use a default drive letter if one is not supplied. Since the data dir fields are applied cluster wide, the choice was either between relying on the default drive letter, or duplicating all fields for Windows nodes. Windows support hasn't been a super huge concern for this feature (i.e. wasn't a formal part of RFC), but I have been doing my best to ensure that it won't be horribly broken or require massive amounts of work to get working for Windows nodes. I'm not aware of any testing done for this feature on Windows nodes as of yet, as there are still some issues being worked out with the current Windows implementation.

brandond:
idk, we default the data-dir to `c:/var/lib/rancher/rke2` so I'm not sure what happens if you leave off the drive letter when customizing the path. Ref:
https://github.com/rancher/system-agent-installer-rke2/blob/main/package/run.ps1#L51

Either way... sounds like Windows is out of scope for the moment. I hope noone was expecting to be able to stick things on a different drive letter using custom data-dir support. We should probably do a v2 of the data dir support RFC that specifically addresses windows.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -194,11 +215,100 @@ func (p *provisioningAdmitter) validateDataDirectories(request *admission.Reques
...
+	if strings.Contains(dir, "/./") || strings.HasSuffix(dir, "/.") {
+		return admission.ResponseBadRequest(
+			fmt.Sprintf("%s data directory cannot contain reference to current directory \"(./)\"", name))
+	}
+
+	// As with the above, a directory like "/var/lib/rancher/..-rke2" is technically a valid folder name
+	if strings.Contains(dir, "/../") || strings.HasSuffix(dir, "/..") {
+		return admission.ResponseBadRequest(
+			fmt.Sprintf("%s data directory cannot contain reference to parent directory \"(../)\"", name))
+	}
```
brandond, line 268: Can we just use https://pkg.go.dev/path/filepath#IsAbs and https://pkg.go.dev/path/filepath#Clean? Make sure it's absolute and is unchanged by cleaning.  Could probably just use strings.ContainsAny for the shell char checks too.

Example: https://go.dev/play/p/6bY3TQ89idh

HarrisonWAffel:
@jakefhyde 

> Windows will use a default drive letter if one is not supplied
This is my understanding as well. When running the `New-Directory` function within the `run.ps1` script the value `/etc2/test2` correctly resolves to the default `C:` drive. Though, it would be much better to be explicit about the drive, if only just for consistencies sake. 

> I'm not aware of any testing done for this feature on Windows nodes as of yet, as there are still some issues being worked out with the current Windows implementation.
QA should do a smoke test on Windows soon, I'm not aware of a specific issue that would block it.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
jakefhyde: Updated. I kept the 3 checks as 3 separate if statements for error handling purposes, moving `filepath.Clean` to the last since it's the most esoteric.

brandond (approved):

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -194,11 +215,78 @@ func (p *provisioningAdmitter) validateDataDirectories(request *admission.Reques
...
+// dir is the path of the data directory, and name corresponds to a print friendly name for this data directory.
+func validateDataDirectoryFormat(dir, name string) *admissionv1.AdmissionResponse {
+	if dir == "" {
+		return admission.ResponseAllowed()
+	}
+	if !filepath.IsAbs(dir) {
+		return admission.ResponseBadRequest(
+			fmt.Sprintf("%s data directory must be an absolute path", name))
+	}
+	if strings.ContainsAny(dir, "${}()`") {
```
ericpromislow, line 239: This implies that these names are run through shell evaluation at some point. Are we also concerned about shell expansion on `*` or `?`?

brandond: I don't know if the scripts are run with `noglob` or not.

If we wanted to be more comprehensive we could probably go with something like 
```golang
strings.ContainsAny(dir, "*?#~=%$|&;<>{}[]()`")
```

I looked around and I couldn't find an existing golang definition of shell metacharacters for us to use here.

docs.md
```diff
@@ -379,12 +379,24 @@ When a UserAttribute is updated, the following checks take place:
...
 
+#### On Create
+
+##### Data Directories
+
+Prevent the creation of new objects with an env var (under `spec.agentEnvVars`) with a name of `CATTLE_AGENT_VAR_DIR`.
+Prevent the creation of new objects with an invalid data directory. An invalid data directory is defined as the 
+following:
+- Is not an absolute path (i.e. does not start with `/`)
+- Attempts to include environment variables (e.g. `$VARIABLE` or `S{VARIABLE}`)
```
jiaqiluo, line 390:
```suggestion
- Attempts to include environment variables (e.g. `$VARIABLE` or `${VARIABLE}`)
```

pkg/resources/provisioning.cattle.io/v1/cluster/Cluster.md
```diff
@@ -1,11 +1,23 @@
...
 
+### On Create
+
+#### Data Directories
+
+Prevent the creation of new objects with an env var (under `spec.agentEnvVars`) with a name of `CATTLE_AGENT_VAR_DIR`.
+Prevent the creation of new objects with an invalid data directory. An invalid data directory is defined as the 
+following:
+- Is not an absolute path (i.e. does not start with `/`)
+- Attempts to include environment variables (e.g. `$VARIABLE` or `S{VARIABLE}`)
```
jiaqiluo, line 11:
```suggestion
- Attempts to include environment variables (e.g. `$VARIABLE` or `${VARIABLE}`)
```

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -194,11 +215,78 @@ func (p *provisioningAdmitter) validateDataDirectories(request *admission.Reques
...
+	}
+
+	for i := range paths {
+		for j := i + 1; j < len(paths); j++ {
+			path1 := paths[i]
+			path2 := paths[j]
+
+			if path1 == path2 {
+				return admission.ResponseBadRequest(
+					fmt.Sprintf("%s data directory cannot be equal to %s data directory", dataDirs[path1], dataDirs[path2]))
```
jiaqiluo, line 271: The keys in the map `dataDirs` are "Distro", "Provisioning" and "System Agent", not the path, which will result in the empty string being placed in the produced message.

jiaqiluo (changes requested):

ericpromislow (approved):
ok with me

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
brandond: good catch!

jakefhyde: Good catch, I had it the other way around but flipped them because if multiple keys were identical it would overwrite before validating, I'll update.

brandond (approved):

jiaqiluo (approved):

brandond (approved):

ericpromislow (approved):
Looks good after latest commit

PR #478, opened 2024-08-15
[v0.4] Bump both rke and dymamiclistener to their release versions

nflynt:
This updates both dependencies to their release tags, in prep for a webhook v0.4.10 release.

PR #481, opened 2024-08-16
[v0.4] bump rancher apis dependency in preparation for v0.4.10

nflynt:
Notably `rancher/rancher` `release/v2.8` has updated its go version to 1.22, so pulling in the updated pkg/apis has updated webhook's go version and toolchain (1.22.5) and a couple of upstream x module versions to match. We should probably re-QA in light of this change.

(Note: This PR is based on https://github.com/rancher/webhook/pull/478)


ericpromislow (dismissed):
LG

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/rancher/webhook
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.5
```
maxsokolovsky, line 5: May be unimportant, but the latest 1.22 is `go1.22.6`.

nflynt: My goal is to match rancher/rancher, which is tracking `go01.22.5` on its release v2.8 branch https://github.com/rancher/rancher/blob/release/v2.8/go.mod

nflynt: (After the code freeze, perhaps we should bump both. But for release purposes, I think it's mostly  just important to have them in sync.)

maxsokolovsky (dismissed):

prachidamle:
What did the merge-base change from @nflynt ?

maxsokolovsky (dismissed):

ericpromislow (dismissed):
LGTM

PR #482, opened 2024-08-16
[v0.5] Bump rancher/rancher package apis to un-rc hosted operators, in prep for v0.5.1 release

nflynt:
As far as I can tell, the only hosted operator with significant changes is `gke-operator` with a rather unremarkable charts update to propagate a debug logging flag. The other two are just upstream CI changes and should have no impact.


ericpromislow (approved):
lgtm

maxsokolovsky (approved):

go.mod
```diff
@@ -41,15 +41,15 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/rancher/dynamiclistener v0.6.0
 	github.com/rancher/lasso v0.0.0-20240705194423-b2a060d103c1
-	github.com/rancher/rancher/pkg/apis v0.0.0-20240710151157-3a6de11e4de9
+	github.com/rancher/rancher/pkg/apis v0.0.0-20240816071246-0c3b39a95054
```
tomleb, line 44: Isn't baaa516fe13cc4abefd0f8a1553bdfe087be8e2c the latest version of pkg/apis in release/v2.9?

nflynt: They're fundamentally the same commit; baaa516fe13cc4abefd0f8a1553bdfe087be8e2c is the merge, and 0c3b39a950540ce43035e2530fd1f6ed97c0be0d is the single commit that merge contained. That's what shows up in github's history for the pkg/apis folder, which is why I grabbed it initially. Do we care enough to want to reference the merge commit hash instead?

PR #501, opened 2024-09-11
Validate creator-principal-name annotation when creating cluster/project

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/46828

Related Rancher changes https://github.com/rancher/rancher/pull/46856

## CheckList
  
- [x] Test
  
- [x] Docs


pkg/resources/common/validation.go
```diff
@@ -58,3 +61,35 @@ func ValidateRules(rules []rbacv1.PolicyRule, isNamespaced bool, fldPath *field.
...
+func CheckCreatorPrincipalName(userCache controllerv3.UserCache, obj metav1.Object) (*field.Error, error) {
+	annotations := obj.GetAnnotations()
+	principalName := annotations[auth.CreatorPrincipalNameAnn]
+	if principalName == "" { // Nothing to check.
+		return nil, nil
+	}
+
+	annotationsFieldPath := field.NewPath("metadata").Child("annotations")
+
+	creatorID := obj.GetAnnotations()[auth.CreatorIDAnn]
```
enrichman, line 74:
```suggestion
	creatorID := annotations[auth.CreatorIDAnn]
```

pkg/resources/common/validation.go
```diff
@@ -58,3 +61,35 @@ func ValidateRules(rules []rbacv1.PolicyRule, isNamespaced bool, fldPath *field.
...
+	}
+
+	creatorPrincipalName := annotations[auth.CreatorPrincipalNameAnn]
+	for _, principal := range user.PrincipalIDs {
+		if principal == creatorPrincipalName {
+			return nil, nil
+		}
+	}
+
+	return field.Invalid(annotationsFieldPath, auth.CreatorPrincipalNameAnn, fmt.Sprintf("creator user %s doesn't have principal %s", creatorID, creatorPrincipalName)), nil
```
enrichman, line 94: We already have the creatorPrincipalName from L67:

```go
principalName := annotations[auth.CreatorPrincipalNameAnn]
```

```suggestion
	for _, principal := range user.PrincipalIDs {
		if principal == principalName {
			return nil, nil
		}
	}

	return field.Invalid(annotationsFieldPath, auth.CreatorPrincipalNameAnn, fmt.Sprintf("creator user %s doesn't have principal %s", creatorID, principalName)), nil
```

pkg/resources/common/validation_test.go
```diff
@@ -60,26 +67,116 @@ func TestValidateRules(t *testing.T) {
 		},
 	}
 
-	for _, testcase := range tests {
-		t.Run("global/"+testcase.name, func(t *testing.T) {
-			err := ValidateRules([]v1.PolicyRule{
-				testcase.data,
+	for _, test := range tests {
+		test := test
```
enrichman, line 71: From Go v1.22 this shouldn't be required anymore, but I'm fine keeping it.

Ref: https://go.dev/blog/loopvar-preview

pkg/resources/common/validation_test.go
```diff
@@ -60,26 +67,116 @@ func TestValidateRules(t *testing.T) {
...
+			creatorID:     "u-12345",
+			principalName: "keycloak_user://12345",
+		},
+		{
+			desc:          "no creatorId annotation",
+			principalName: "keycloak_user://12345",
+			fieldErr:      true,
+		},
+		{
+			desc:          "principle doesn't belong to creator",
```
enrichman, line 141:
```suggestion
			desc:          "principal doesn't belong to creator",
```

enrichman (commented):
Some small comments.

pkg/resources/management.cattle.io/v3/cluster/validator.go
```diff
@@ -78,26 +85,36 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
 	}
 
-	if request.Operation == admissionv1.Create || request.Operation == admissionv1.Update {
-		cluster, err := objectsv3.ClusterFromRequest(&request.AdmissionRequest)
+	cluster, err := objectsv3.ClusterFromRequest(&request.AdmissionRequest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get cluster from request: %w", err)
+	}
+
+	if request.Operation == admissionv1.Create && cluster.Name != "local" {
```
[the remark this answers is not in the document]
crobby: I think that makes sense.  Relying solely on cluster.Name seems a bit flimsy.

pkg/resources/management.cattle.io/v3/project/validator_test.go
```diff
@@ -906,17 +1047,18 @@ func TestProjectValidation(t *testing.T) {
 	for _, test := range tests {
 		test := test
 		t.Run(test.name, func(t *testing.T) {
-			t.Parallel()
+			// t.Parallel()
```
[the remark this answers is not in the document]
pmatseykanets: We are 👍. This wasn't reverted back after experiments. Thanks.

pkg/resources/common/validation.go
pmatseykanets: Duh, of course. The no-op check came later.

pkg/resources/management.cattle.io/v3/cluster/validator.go
```diff
@@ -78,26 +85,36 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
+		fieldErr, err := common.CheckCreatorPrincipalName(a.userCache, cluster)
```
raulcabello, line 94: why do we validate this just on create? shouldn't we also check on updates?

pkg/resources/common/validation_test.go
```diff
@@ -60,26 +67,116 @@ func TestValidateRules(t *testing.T) {
...
 		})
 	}
 }
+
+func TestCheckCreatorPrincipalName(t *testing.T) {
+	t.Parallel()
+
+	ctrl := gomock.NewController(t)
+	userCache := fake.NewMockNonNamespacedCacheInterface[*v3.User](ctrl)
+	userCache.EXPECT().Get(gomock.Any()).DoAndReturn(func(name string) (*v3.User, error) {
```
raulcabello, line 104: nit: It'd be nice to remove `gomock.Any()` and replace it with the expected parameter if possible

pmatseykanets: This is a perfect example where you can't use anything other than `gomock.Any()` as it expects multiple values. In fact ANY value is acceptable in the context of this test.

pkg/resources/management.cattle.io/v3/cluster/validator.go
pmatseykanets: Under what circumstances we expect the owner of the cluster/project to change? Shouldn't it be immutable? I mean it only make sense when we create a cluster/project and even then if you already have necessary permissions you don't even necessarily need owner RBAC objects. There is even an idea to add an annotation to skip the creation of owner RBAC.
Having said that, we can make sure that both creator id and principal annotations are in fact immutable on update. Though it might be considered a kind of breaking change as we didn't check for that before.

pkg/resources/management.cattle.io/v3/cluster/validator.go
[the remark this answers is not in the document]
pmatseykanets: What do you think about 9d4436ef823a45816e582acdf12261166ce1841e?

pkg/resources/common/validation_test.go
raulcabello: I don't agree with you here. `userCache.Get()` always expects the `creatorId` (`annotations[auth.CreatorIDAnn]`), and we don't verify this parameter because we are using `gomock.any()`. For example, if you change this [line](https://github.com/pmatseykanets/webhook/blob/validate-creator-principal-name/pkg/resources/common/validation.go#L79) to `user, err := userCache.Get("wrong-id")` application would stop working, but the test would still pass.

In order to get rid of `gomock.any()`, you'd need to create a mock for each test case inside the tests struct. I'd create a func that accept a string (the  `creatorId`) and returns the mock.

Something like:

```
tests := []struct {
		desc          string
		creatorID     string
		principalName string
		userCache func(string) controllerv3.UserCache
		fieldErr      bool
		err           bool
	}{

		{
			desc:          "creator id and principal name match",
			creatorID:     "u-12345",
			principalName: "keycloak_user://12345",
			userCache: func(creatorId string) controllerv3.UserCache {
				userCache := fake.NewMockNonNamespacedCacheInterface[*v3.User](ctrl)
				userCache.EXPECT().Get(creatorId).DoAndReturn(func(name string) (*v3.User, error) {
					return &v3.User{
						ObjectMeta: metav1.ObjectMeta{
							Name: "u-12345",
						},
						PrincipalIDs: []string{"local://12345", "keycloak_user://12345"},
					}, nil
				}
				return userCache
			},
		},
		{
			desc:          "error getting creator user",
			creatorID:     "u-error",
			principalName: "keycloak_user://12345",
			userCache: func(creatorId string) controllerv3.UserCache {
				userCache := fake.NewMockNonNamespacedCacheInterface[*v3.User](ctrl)
				userCache.EXPECT().Get(creatorId).DoAndReturn(func(name string) (*v3.User, error) {
					return nil, fmt.Errorf("some error")
				}
				return userCache
			},
			err:           true,
		},
	}

	for _, test := range tests {
		test := test
		t.Run(test.desc, func(t *testing.T) {
			t.Parallel()

			annotations := map[string]string{}
			if test.creatorID != "" {
				annotations[auth.CreatorIDAnn] = test.creatorID
			}
			if test.principalName != "" {
				annotations[auth.CreatorPrincipalNameAnn] = test.principalName
			}

			fieldErr, err := CheckCreatorPrincipalName(test.userCache(test.creatorID), &v3.Project{
				ObjectMeta: metav1.ObjectMeta{
					Annotations: annotations,
				},
			})
			require.Equal(t, test.fieldErr, fieldErr != nil)
			require.Equal(t, test.err, err != nil)
		})
	}
```

You can specify different mock for each test case by using the previous approach. Then you can get rid of the [switch](https://github.com/pmatseykanets/webhook/blob/validate-creator-principal-name/pkg/resources/common/validation_test.go#L105) case that returns a different value based on the name, which can become difficult to read if this test grows in the future.

raulcabello: You an find an example of this approach in https://github.com/raulcabello/rancher/blob/134a10ce985571b9856814e26011766ab0a7867f/pkg/auth/requests/impersonate_test.go#L27

pkg/resources/management.cattle.io/v3/cluster/validator.go
raulcabello: > we can make sure that both creator id and principal annotations are in fact immutable on update

I like this idea. It'll stop working if the creator id or principal annotations are manually changed at the moment. Making them immutable makes sense to me.

pkg/resources/common/validation_test.go
pmatseykanets: While it's possible to do this in this particular case given a relatively small number of test cases it definitely doesn't help with readability and amount of additional code for tests organized this way.

> and we don't verify this parameter because we are using 

Actually we do indirectly with the `switch`.

> For example, if you change this [line](https://github.com/pmatseykanets/webhook/blob/validate-creator-principal-name/pkg/resources/common/validation.go#L79) to user, err := userCache.Get("wrong-id") application would stop working, but the test would still pass.

Not really, no. The two main test cases break because of that indirect assertion in the `DoAndReturn()`.

That being said I think matching arguments on mock expectations is a very useful tool , especially for simple type arguments, and help to have a more comprehensive tests. But as everything else it comes with a cost.

crobby (approved):
lg

pkg/resources/management.cattle.io/v3/cluster/Cluster.md
```diff
@@ -0,0 +1,5 @@
+## Cluster
```
JonCrowther, line 1: We don't need to specify cluster here because that gets generated in docs.md
```suggestion
## Validation Checks
```

pkg/resources/management.cattle.io/v3/cluster/Cluster.md
```diff
@@ -0,0 +1,5 @@
...
+
+### Annotations validation
+
+When a cluster is created if `field.cattle.io/creator-principal-name` annotation is set then `field.cattle.io/creatorId` annotation must be set as well and the user's principal name in the former should match the creator's user principal id.
```
JonCrowther, line 5: These are two separate checks so they should be different sentences for clarity
```suggestion
When a cluster is created and `field.cattle.io/creator-principal-name` annotation is set then `field.cattle.io/creatorId` annotation must be set as well. 
The value of `field.cattle.io/creator-principal-name` should match the creator's user principal id.
```

pkg/resources/management.cattle.io/v3/cluster/validator.go
JonCrowther: Just make sure that you can change from no annotation -> annotation, just not annotation -> new annotation. Otherwise it'll block the controller that adds the annotation.

pkg/resources/management.cattle.io/v3/project/Project.md
```diff
@@ -19,6 +19,10 @@ The container default resource configuration must have properly formatted quanti
 
 Limits for any resource must not be less than requests.
 
+### Annotations validation
+
+When a project is created if `field.cattle.io/creator-principal-name` annotation is set then `field.cattle.io/creatorId` annotation must be set as well and the user's principal name in the former should match the creator's user principal id.
```
JonCrowther, line 24: Similarly, these are 2 different validations and could be 2 sentences
```suggestion
When a project is created and `field.cattle.io/creator-principal-name` annotation is set then `field.cattle.io/creatorId` annotation must be set as well.
The value of `field.cattle.io/creator-principal-name` should match the creator's user principal id.
```

pkg/resources/common/validation.go
```diff
@@ -58,3 +61,50 @@ func ValidateRules(rules []rbacv1.PolicyRule, isNamespaced bool, fldPath *field.
 	}
 	return returnErr
 }
+
+var annotationsFieldPath = field.NewPath("metadata").Child("annotations")
+
+func CheckCreatorPrincipalName(userCache controllerv3.UserCache, obj metav1.Object) (*field.Error, error) {
```
JonCrowther, line 67: Since these functions are exported and in a common validation file, could you add a comment at the top of the functions describing them?

pkg/resources/management.cattle.io/v3/cluster/Cluster.md
pmatseykanets: Agreed.

pkg/resources/management.cattle.io/v3/cluster/Cluster.md
pmatseykanets: Nice rewording!

pkg/resources/common/validation.go
pmatseykanets: Of course.

pkg/resources/common/validation.go
```diff
@@ -58,3 +61,50 @@ func ValidateRules(rules []rbacv1.PolicyRule, isNamespaced bool, fldPath *field.
...
+	for _, principal := range user.PrincipalIDs {
+		if principal == principalName {
+			return nil, nil
+		}
+	}
+
+	return field.Invalid(annotationsFieldPath, auth.CreatorPrincipalNameAnn, fmt.Sprintf("creator user %s doesn't have principal %s", creatorID, principalName)), nil
+}
+
+func CheckCreatorAnnotationsOnUpdate(oldObj, newObj metav1.Object) *field.Error {
```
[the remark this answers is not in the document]
pmatseykanets: Under what circumstances we create a cluster/project object and then add annotations as a separate update operation?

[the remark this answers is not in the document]
JonCrowther: My bad, looking through the issue and rancher changes, I see that the UI is responsible for specifying the annotation on the create. I had assumed the annotation was added after (which is how a lot of RBAC works IIRC). Resolved 👍

pkg/resources/management.cattle.io/v3/cluster/validator.go
pmatseykanets: > Just make sure that you can change from no annotation -> annotation, just not annotation -> new annotation.

We check for immutability only on update [145d37a](https://github.com/rancher/webhook/pull/501/commits/145d37a76aad5cda7e90ab8094f271c5a1c2a843) and the only allowed change to these annotations is to remove them.
These annotations make sense only on create and should be present on the object. Meaning you can't create a cluster and then set these annotations with update.

JonCrowther (approved):

crobby (approved):

raulcabello (approved):

pkg/resources/common/validation.go
```diff
@@ -58,3 +61,55 @@ func ValidateRules(rules []rbacv1.PolicyRule, isNamespaced bool, fldPath *field.
...
+// This function should only be called for the update operation.
+func CheckCreatorAnnotationsOnUpdate(oldObj, newObj metav1.Object) *field.Error {
+	oldAnnotations := oldObj.GetAnnotations()
+	newAnnotations := newObj.GetAnnotations()
+
+	for _, annotation := range []string{auth.CreatorIDAnn, auth.CreatorPrincipalNameAnn} {
+		if _, ok := newAnnotations[annotation]; ok {
+			// If the annotation exists on the new object it must be the same as on the old object.
+			if oldAnnotations[annotation] != newAnnotations[annotation] {
+				return field.Invalid(annotationsFieldPath, annotation, "annotation is immutable")
```
bigkevmcd, line 109: What if both of them are changed?

Maybe we could use the `field.ErrorList`  which you can `.Aggregate()` or combine with `apierrors.NewInvalid` ?

It's usually nicer to report all validation errors in a single pass?

pmatseykanets: This is deliberate. None of the checks for cluster and/or project are aggregated or any other object (except for PSA) for that matter. In general I agree. I'd like to have my validation errors altogether. But IMHO consistency is the key.

tomleb (approved):

PR #511, opened 2024-09-24
Add checks for annotation to opt out of cluster owner RBAC

JonCrowther:
## Issue: https://github.com/rancher/rancher/issues/45591
Requires rancher/rancher changes: https://github.com/rancher/rancher/pull/47259

## Problem
When a service account creates a cluster, the project handler and cluster handler both attempt to create ClusterOwner and ProjectOwner roles and bind those roles to the service account. The problem is that we don't support those roles for service accounts, so the logs get flooded with errors as it keeps retrying to add those roles.

## Solution
To avoid the errors and the re-queuing, we have added a new annotation `field.cattle.io/no-creator-rbac`. When it is set the webhook does not set the `field.cattle.io/creatorId` annotation to clusters because the controllers will no longer be creating RBAC for the creator of the cluster.

I moved some annotations to the `common` package as I felt it was more apt. I also cleaned up some functions in the `common` package to make them more consistent.

## CheckList
  
- [X] Test
  
- [X] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/cluster/validator.go
+++ b/pkg/resources/management.cattle.io/v3/cluster/validator.go
@@ -107,7 +107,12 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 	}
 
 	if request.Operation == admissionv1.Create || request.Operation == admissionv1.Update {
-		// no need to validate the local cluster, or imported cluster which represents a KEv2 cluster (GKE/EKS/AKS) or v1 Provisioning Cluster
+		if fieldErr := common.CheckCreatorIDAndNoCreatorRBAC(newCluster); fieldErr != nil {
+			return admission.ResponseBadRequest(fieldErr.Error()), nil
+		}
+
+		// no need to validate the PodSecurityAdmissionConfigurationTemplate on a local cluster,
+		// or imported cluster which represents a KEv2 cluster (GKE/EKS/AKS) or v1 Provisioning Cluster
 		if newCluster.Name == "local" || newCluster.Spec.RancherKubernetesEngineConfig == nil {
 			return admission.ResponseAllowed(), nil
 		}
--- a/pkg/resources/management.cattle.io/v3/cluster/validator.go
+++ b/pkg/resources/management.cattle.io/v3/cluster/validator.go
@@ -92,6 +92,9 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 	if a.userCache != nil {
 		// The following checks don't make sense for downstream clusters (userCache == nil)
 		if request.Operation == admissionv1.Create {
+			if fieldErr := common.CheckCreatorIDAndNoCreatorRBAC(newCluster); fieldErr != nil {
+				return admission.ResponseBadRequest(fieldErr.Error()), nil
+			}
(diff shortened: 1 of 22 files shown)


pmatseykanets (commented):
I think we need to add checks for `creatorId` in the presence of `no-creator-rbac` in cluster and project validators. Unfortunately the `common. CheckCreatorID()` can't be used in those cases as it also checks if the `creatorId` matches `request.UserInfo.Username` which is not going to work (for projects for sure).

JonCrowther:
> I think we need to add checks for `creatorId` in the presence of `no-creator-rbac` in cluster and project validators. Unfortunately the `common. CheckCreatorID()` can't be used in those cases as it also checks if the `creatorId` matches `request.UserInfo.Username` which is not going to work (for projects for sure).

I've added a new function to common to check that only one of `creatorId` and `no-creator-rbac` is set for use by project and cluster validators.

pkg/resources/management.cattle.io/v3/cluster/validator.go
```diff
@@ -107,7 +107,12 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 	}
 
 	if request.Operation == admissionv1.Create || request.Operation == admissionv1.Update {
-		// no need to validate the local cluster, or imported cluster which represents a KEv2 cluster (GKE/EKS/AKS) or v1 Provisioning Cluster
+		if fieldErr := common.CheckCreatorIDAndNoCreatorRBAC(newCluster); fieldErr != nil {
+			return admission.ResponseBadRequest(fieldErr.Error()), nil
+		}
+
```
pmatseykanets, line 113: I think this should go before or after [L95](https://github.com/rancher/webhook/pull/511/files/1d2da326e32215e9baefa122eea248ce3425e547..6bd7ee78378c33405ecfe7769d40deae81f2d8b1#diff-d7e4efad65a61bc47176fc39ab03add1805d4effff6cade04a89c7eec05c36d7R95)
IMHO Creator annotation (any) checks only make sense for Create operation and this case only in local/upstream cluster.

pkg/resources/management.cattle.io/v3/project/validator.go
```diff
@@ -141,6 +141,10 @@ func (a *admitter) admitUpdate(oldProject, newProject *v3.Project) (*admissionv1
 }
 
 func (a *admitter) admitCommonCreateUpdate(oldProject, newProject *v3.Project) (*admissionv1.AdmissionResponse, error) {
+	if fieldErr := common.CheckCreatorIDAndNoCreatorRBAC(newProject); fieldErr != nil {
+		return admission.ResponseBadRequest(fieldErr.Error()), nil
+	}
```
pmatseykanets, line 146: Similarly this should probably go before or after L118.

pkg/resources/management.cattle.io/v3/cluster/validator.go
pmatseykanets: IMHO similar to creatorId and creator-principal-name, no-creator-rbac annotation should be immutable. The only allowed update should be removing the annotation. 
So we should add it ot this check https://github.com/rancher/webhook/blob/6b74a9a9006ccb27e74146a91899a02dd03a7d3b/pkg/resources/common/validation.go#L101

pmatseykanets (approved):

crobby (approved):

PR #513, opened 2024-09-27
Prevents dropping unknown cluster fields

dharmit:
## Issue: 

https://github.com/rancher/rancher/issues/47200

## Problem

Presently mutating webhook drops fields not present in `Cluster` struct in RKE1. This creates problem for the users of ACI CNI provider.

## Solution

We pass the marshaled `newCluster` object created by the function `ClusterOldAndNewFromRequest` instead of current approach of using `request.Object.Raw`.

## CheckList
  
- [ ] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/cluster/mutator.go
+++ b/pkg/resources/management.cattle.io/v3/cluster/mutator.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 
@@ -59,6 +60,10 @@ func (m *ManagementClusterMutator) Admit(request *admission.Request) (*admission
 	if err != nil {
 		return nil, fmt.Errorf("failed to get old and new clusters from request: %w", err)
 	}
+	newClusterRaw, err := json.Marshal(newCluster)
+	if err != nil {
+		return nil, fmt.Errorf("unable to re-marshal new cluster: %w", err)
+	}
 	// no need to mutate the local cluster, or imported cluster which represents a KEv2 cluster (GKE/EKS/AKS) or v1 Provisioning Cluster
 	if newCluster.Name == "local" || newCluster.Spec.RancherKubernetesEngineConfig == nil {
 		return admission.ResponseAllowed(), nil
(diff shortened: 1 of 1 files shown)


pkg/resources/management.cattle.io/v3/cluster/mutator.go
```diff
@@ -94,7 +99,9 @@ func (m *ManagementClusterMutator) Admit(request *admission.Request) (*admission
 	}
 
 	response := &admissionv1.AdmissionResponse{}
-	if err := patch.CreatePatch(request.Object.Raw, newCluster, response); err != nil {
+	// we use the re-marshalled old cluster to make sure that the patch doesn't drop "unknown" fields which were
```
snasovich, line 102:
```suggestion
	// we use the re-marshalled new cluster to make sure that the patch doesn't drop "unknown" fields which were
```

snasovich (commented):
Minor nit on the comment wording. Otherwise LGTM.
However, please populate PR description with more information - currently it looks like a placeholder.
Also CI is failing - please look into it.

dharmit:
> Minor nit on the comment wording. Otherwise LGTM. 
> Also CI is failing - please look into it.

Fixed. Thanks.

> However, please populate PR description with more information - currently it looks like a placeholder. 

Ack. Will do.

@snasovich there's a commit I have added for unit tests. It's WIP and heavily based on the unit test for validator in the same package. I'm going to build up on that tomorrow, but some early feedback and/or suggestions might help. My best case hope is to add more unit tests, and worst case readiness is to remove unit tests completely.

snasovich:
@dharmit , thank you. Please prioritize adding extended description to this PR to give enough context to the reviewers.

Obviously like the idea of adding a unit test and looks like you're on the right track - though the main challenge is crafting `Object` and `OldObject` in a way to test this scenario as well as asserting that generated response's patch doesn't have entries to drop unknown fields. It's very much doable but given the time constraints we may want to create a follow up tech debt issue to add unit test to this file.
Let's see where you're by end of tomorrow (Oct 1st) and go from there.

jakefhyde (approved):
Seeing as this is pretty much exactly what was done for provisioning clusters with the same problem, approving.

dharmit:
@snasovich I wasn't able to make enough progress on unit test to keep this PR waiting for it. Can you suggest where should I create a "tech debt" issue about it?

@jakefhyde my force push has removed your approval from the PR. Can you please help add it again?

MbolotSuse (changes requested):
The code LGTM. However, I think we should add some tests for this. As @jakefhyde noted, there's something similar done for provisioning clusters, and there does appear to be a test for that: https://github.com/rancher/webhook/blob/6b74a9a9006ccb27e74146a91899a02dd03a7d3b/pkg/resources/provisioning.cattle.io/v1/cluster/mutator_test.go#L591. Might be a quick way to test what you are looking for.

MbolotSuse:
@dharmit If you can also add a `replace` directive in the go.mod for the RKE version: https://github.com/rancher/webhook/blob/6b74a9a9006ccb27e74146a91899a02dd03a7d3b/go.mod#L45 that would be good. Idea here is to make sure that we don't get this updated in the future when we update for other types (which could pull in an RC version).

kinarashah:
@MbolotSuse Thanks much for the link, that was helpful in adding a simple test to validate our most common use case - introducing new ACI fields. Added test and updated go.mod in https://github.com/rancher/webhook/pull/515.

PR #520, opened 2024-10-08
Validate LastUsedAt for Token and ClusterAuthToken

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/45732

Related work
- https://github.com/rancher/rancher/pull/46313
- https://github.com/rancher/kube-api-auth/pull/28

## Problem

A new optional field `lastUsedAt` was added to Token and ClusterAuthToken of type `metav1.Time`.

## Solution

Validate that if set the `lastUsedAt` field of Token and ClusterAuthToken contains a valid RFC3339 date time value.

## CheckList
  
- [ ] Test
  
- [ ] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -1,6 +1,6 @@
-# catalog.cattle.io/v1 
+# catalog.cattle.io/v1
 
-## ClusterRepo 
+## ClusterRepo
 
 ### Validation Checks
 
@@ -16,9 +16,27 @@ Users cannot update a ClusterRepo which violates the following constraints:
 
 - Fields GitRepo and URL are mutually exclusive and so both cannot be filled at once.
 
-# core/v1 
+# cluster.cattle.io/v3
 
-## Namespace 
+## ClusterAuthToken
+
+### Validation Checks
+
+#### Invalid Fields - Create
+
+When a ClusterAuthToken is created, the following checks take place:
+
+- If set, `lastUsedAt` must be a valid date time according to RFC3339 (e.g. `2023-11-29T00:00:00Z`).
+
+#### Invalid Fields - Update
+
+When a ClusterAuthToken is updated, the following checks take place:
+
+- If set, `lastUsedAt` must be a valid date time according to RFC3339 (e.g. `2023-11-29T00:00:00Z`).
+
+# core/v1
+
+## Namespace
 
 ### Validation Checks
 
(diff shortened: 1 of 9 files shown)


crobby (dismissed):

andreas-kupries (approved):

docs.md
```diff
@@ -126,7 +144,7 @@ Users can update the following fields if they have not been set, but after they
 
 In addition, as in the create validation, both a user subject and a group subject cannot be specified.
```
ericpromislow, line 145: Not part of this change per se, but I stumbled over it. Should be "`create` validation" (because 'create' is a standard English word).

docs.md
```diff
@@ -308,7 +326,7 @@ changed:
...
```
ericpromislow, line 327: Same note about tt-quoting `create`

pkg/server/handlers.go
```diff
@@ -68,16 +71,17 @@ func Validation(clients *clients.Clients) ([]admission.ValidatingAdmissionHandle
 		roles := role.NewValidator()
 		rolebindings := rolebinding.NewValidator()
 		setting := setting.NewValidator(clients.Management.Cluster().Cache(), clients.Management.Setting().Cache())
```
ericpromislow, line 73: I'm not big on local variables shadowing package names, but I see here that they're just being used to avoid coming up with a very long 1-liner. And the assignment to `token` is the only change in this PR. How about a compromise, call it `tokenValidator`, and we'll call it even.

ericpromislow (approved):
This is fine. Feel free to ignore all my remarks, as they're all nits.

docs.md
pmatseykanets: Better do it outside of this PR.

pkg/server/handlers.go
pmatseykanets: Shadowing is part of day-to-day life with Go and package shadowing is the least problematic type. But I agree here it could've been avoided simply by going with the plural `settings := ...`.
One the other hand that one liner is much better IMHO when broken into separate lines because that `append` that follows is already out-of-hand and resource packages are self explanatory.

ericpromislow (approved):
LGTM

andreas-kupries (approved):
changes looking ok

PR #528, opened 2024-10-28
Bump to k8s 1.31

ericpromislow:


go.mod
```diff
@@ -6,33 +6,33 @@ toolchain go1.22.7
...
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.30.1
-	k8s.io/kubectl => k8s.io/kubectl v0.30.1
-	k8s.io/kubelet => k8s.io/kubelet v0.30.1
-	k8s.io/kubernetes => k8s.io/kubernetes v1.30.1
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.30.1
-	k8s.io/metrics => k8s.io/metrics v0.30.1
-	k8s.io/mount-utils => k8s.io/mount-utils v0.30.1
-	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.30.1
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.30.1
+	k8s.io/api => k8s.io/api v0.31.0
```
tomleb, line 9: r/r appears to be using `.1` patch version so I think we should too: https://github.com/rancher/rancher/blob/main/go.mod#L28

go.mod
```diff
@@ -6,33 +6,33 @@ toolchain go1.22.7
...
+	k8s.io/kubelet => k8s.io/kubelet v0.31.0
+	k8s.io/kubernetes => k8s.io/kubernetes v1.31.1
+	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.31.0
+	k8s.io/metrics => k8s.io/metrics v0.31.0
+	k8s.io/mount-utils => k8s.io/mount-utils v0.31.0
+	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.31.0
+	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.31.0
 )
 
 require (
```
tomleb, line 38: The current version of wrangler does not support v1.31 so needs to be updated. you can use the tag `v3.1.0-rc.1`. 
The version of lasso should probably be updated too to the one that wrangler is using: https://github.com/rancher/wrangler/blob/master/go.mod#L12.

tomleb (changes requested):

go.mod
ericpromislow: done

go.mod
ericpromislow: done

go.mod
```diff
@@ -6,93 +6,94 @@ toolchain go1.22.7
...
 	github.com/evanphx/json-patch v5.9.0+incompatible
-	github.com/golang/mock v1.6.0
 	github.com/gorilla/mux v1.8.1
 	github.com/rancher/dynamiclistener v0.6.1-rc.1
-	github.com/rancher/lasso v0.0.0-20240705194423-b2a060d103c1
+	github.com/rancher/lasso v0.0.0-20240924233157-8f384efc8813
 	github.com/rancher/rancher/pkg/apis v0.0.0-20240903164338-21e4787cd0b3
 	github.com/rancher/rke v1.6.0
-	github.com/rancher/wrangler/v3 v3.0.0
+	github.com/rancher/wrangler/v3 v3.0.1-rc.3
```
tomleb, line 46: As per my other comment, this should be [v3.1.0-rc.1](https://github.com/rancher/wrangler/releases/tag/v3.1.0-rc.1). They point to the same commit but unRCing will be less confusing if we use the correct minor version.

tomleb (approved):

gehrkefc (approved):
LGTM

PR #532, opened 2024-10-30
[v0.6] Populate backing namespace field for projects

JonCrowther:
## Issue: 
Builds on https://github.com/rancher/rancher/pull/47723

PR #535, opened 2024-10-31
[0.6] Make sure to update the name in the mutator

JonCrowther:
Fixes an issue with https://github.com/rancher/webhook/pull/532 where the name wasn't being updated

### DIFF
--- a/pkg/resources/management.cattle.io/v3/project/mutator.go
+++ b/pkg/resources/management.cattle.io/v3/project/mutator.go
@@ -126,10 +126,9 @@ func (m *Mutator) createProjectNamespace(project *v3.Project) (*v3.Project, erro
 	// for the backing namespace, we need to generate it ourselves
 	if project.Name == "" {
 		// If err is nil, (meaning "project exists", see below) we need to repeat the generation process to find a project name and backing namespace that isn't taken
-		var newName string
 		for err == nil {
-			newName = names.SimpleNameGenerator.GenerateName(project.GenerateName)
-			_, err = m.projectClient.Get(newProject.Spec.ClusterName, newName, v1.GetOptions{})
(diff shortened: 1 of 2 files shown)


MbolotSuse (commented):
Idea looks ok to me (assuming that if you set both name and generate name k8s won't just re-make the name based on generate name, though I would test that). I would definitely add a test for this overall logic if we can.

nflynt (approved):

pkg/resources/management.cattle.io/v3/project/mutator.go
```diff
@@ -126,8 +126,9 @@ func (m *Mutator) createProjectNamespace(project *v3.Project) (*v3.Project, erro
 	// for the backing namespace, we need to generate it ourselves
 	if project.Name == "" {
 		// If err is nil, (meaning "project exists", see below) we need to repeat the generation process to find a project name and backing namespace that isn't taken
+		var newName string
 		for err == nil {
-			newName := names.SimpleNameGenerator.GenerateName(project.GenerateName)
+			newName = names.SimpleNameGenerator.GenerateName(project.GenerateName)
```
pmatseykanets, line 131: I'd probably just assign it directly instead and use it in code that follows
```suggestion
			newProject.Name = names.SimpleNameGenerator.GenerateName(project.GenerateName)
```

pmatseykanets (dismissed):

JonCrowther:
> Idea looks ok to me (assuming that if you set both name and generate name k8s won't just re-make the name based on generate name, though I would test that). I would definitely add a test for this overall logic if we can.

I have tested the generateName vs name, and as long as name is populated, k8s won't use generate name to overwrite it

pmatseykanets (approved):

nflynt (approved):
:+1: Tests look like they'd cover the documented business requirements for the generative name prefix.

PR #549, opened 2024-11-25
[0.7] Remove references to Restricted Admin

JonCrowther:
## Problem
Restricted admin is fully deprecated and removed in rancher 2.11. Need to reflect those changes in webhook.

## Solution
Removed all the references to restricted-admin, which is exclusively in GlobalRoles that provides some special rules.

Doesn't need the 2.11 changes since it just looks for the GlobalRole name `restricted-admin`. Removed all related tests as well since they no longer are valid.

## CheckList
- [x] Test
  
- [X] Docs


raulcabello (approved):

crobby (approved):
lgtm

PR #550, opened 2024-11-26
[main/2.10.2] Add resource request and limit validation when creating a namespace

crobby:
## Issue: 
[#44825](https://github.com/rancher/rancher/issues/44825)

## Problem

There is no validation of the cpu/memory request/limits when creating namespaces, which leads to Rancher attempting to create invalid objects.

## Solution

The webhook now validates cpu/memory request/limits, if provided.
If a request or limit is provided by itself, we just validate that it is indeed a properly formatted value.
If both a cpu request/limit or a memory request/limit pair is provided, we also check to be sure that the request is less than or equal to the limit.
If parsing any of the provided values fails or if a request is greater than the limit, the webhook will refuse the request.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -60,6 +60,10 @@ The following labels are considered relevant for PSA enforcement:
 - pod-security.kubernetes.io/warn
 - pod-security.kubernetes.io/warn-version
 
+#### Namespace resource limit validation
+
+Validation ensures that the limits for cpu/memory must not be less than the requests for cpu/memory.
+
 ## Secret
 
 ### Validation Checks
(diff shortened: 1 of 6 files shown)


ericpromislow (approved):
Looks ok to me. Some might suggest you could fold the 2 comparisons in `validateResourceLimitsWithUnits` into a separate function, because they have identical structures, but given that those are most likely the only two attributes that we'll ever compare I would leave the code as is, favoring repetition over encapsulation and more complexity (and even more complexity if you use reflection on the field names).

nflynt (approved):
:+1: This looks good. I might nitpick and suggest tests for an exact match between limit and requests, just to cover the edge of what's acceptable, but I won't block on that.

PR #551, opened 2024-12-05
Prevent deletion of `local` cluster

dharmit:
## Issue: 

https://github.com/rancher/rancher/issues/32745
https://github.com/rancher/rancher/issues/48303

## Problem

Presently, it's possible to delete the `local` cluster using:
```sh
$ k delete cluster.management.cattle.io local
```

Other ways the `local` cluster, one on which Rancher is deployed, could be corrupted are mentioned in https://github.com/rancher/rancher/issues/48303.

## Solution

* Prevent `local` cluster deletion:

  ```sh
  $  k delete cluster.management.cattle.io local
  Error from server (BadRequest): admission webhook "rancher.cattle.io.clusters.management.cattle.io" denied the request: cannot delete the local cluster
  ```

* Prevent deletion of `local` and `fleet-local` namespaces:
  ```sh
  $ k delete ns local
  Error from server (BadRequest): admission webhook "rancher.cattle.io.namespaces" denied the request: deletion of namespace "local" is not allowed

  $ k delete ns fleet-local
  Error from server (BadRequest): admission webhook "rancher.cattle.io.namespaces" denied the request: deletion of namespace "fleet-local" is not allowed
  ```

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/pkg/resources/core/v1/namespace/projectannotations.go
+++ b/pkg/resources/core/v1/namespace/projectannotations.go
@@ -15,6 +15,8 @@ import (
 )
 
 const (
+	fleetLocalNs        = "fleet-local"
+	localNs             = "local"
 	manageNSVerb        = "manage-namespaces"
 	projectNSAnnotation = "field.cattle.io/projectId"
 )
@@ -23,8 +25,10 @@ type projectNamespaceAdmitter struct {
 	sar authorizationv1.SubjectAccessReviewInterface
 }
 
-// Admit ensures that the user has permission to change the namespace annotation for
-// project membership, effectively moving a project from one namespace to another.
+// Admit ensures that the:
+//   - user has permission to change the namespace annotation for project membership, effectively moving a project from
+//     one namespace to another.
+//   - deletion of `local` and `fleet-local` namespace is not allowed
 func (p *projectNamespaceAdmitter) Admit(request *admission.Request) (*admissionv1.AdmissionResponse, error) {
 	listTrace := trace.New("Namespace Admit", trace.Field{Key: "user", Value: request.UserInfo.Username})
 	defer listTrace.LogIfLong(admission.SlowTraceDuration)
@@ -36,6 +40,11 @@ func (p *projectNamespaceAdmitter) Admit(request *admission.Request) (*admission
 		return nil, fmt.Errorf("failed to decode namespace from request: %w", err)
 	}
 
+	if request.Operation == admissionv1.Delete {
+		if oldNs.Name == localNs || oldNs.Name == fleetLocalNs {
+			return admission.ResponseBadRequest(fmt.Sprintf("deletion of namespace %q is not allowed\n", request.Name)), nil
+		}
+	}
 	projectAnnoValue, ok := newNs.Annotations[projectNSAnnotation]
 	if !ok {
 		// this namespace doesn't belong to a project, let standard RBAC handle it
--- a/pkg/resources/common/psa-validation.go
+++ b/pkg/resources/common/psa-validation.go
@@ -23,7 +23,7 @@ var psaLabels = []string{
 
 // IsUpdatingPSAConfig will indicate whether or not the labels being passed in
 // are attempting to update PSA-related configuration.
-func IsUpdatingPSAConfig(old map[string]string, new map[string]string) bool {
+func IsUpdatingPSAConfig(old, new map[string]string) bool {
 	for _, label := range psaLabels {
 		if old[label] != new[label] {
 			return true
--- a/pkg/resources/core/v1/namespace/projectannotations_test.go
+++ b/pkg/resources/core/v1/namespace/projectannotations_test.go
@@ -30,6 +30,7 @@ func TestValidateProjectNamespaceAnnotations(t *testing.T) {
 		sarError                  bool
 		wantError                 bool
 		wantAllowed               bool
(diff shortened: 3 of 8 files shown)


ericpromislow (dismissed):
Given that `"local"` already appears in that file, I would set it to a constant called `localCluster`, like in `pkg/resolvers/grbRuleResolvers.go`

pkg/resources/management.cattle.io/v3/node/validator.go
```diff
@@ -0,0 +1,63 @@
...
+
+type admitter struct{}
+
+// Admit is the entrypoint for the validator. It lets a request through or prevent user from executing it.
+func (a admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResponse, error) {
+	oldNode, _, err := objectsv3.NodeOldAndNewFromRequest(&request.AdmissionRequest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get old and new node from request: %w\n", err)
+	}
+	if request.Operation == admissionv1.Delete && oldNode.Namespace == localNs {
```
jakefhyde, line 33: What if a user rotates all of the nodes in their local cluster? I don't think we should be leaving all of the nodes that have ever existed around. If they rotate nodes say once a month, after a year you could have 96 left over node objects. Rotate enough nodes, and you could have resource exhaustion in the local cluster. I would recommend updating the approach to maybe account for accidental deletion, but allow intentional deletion. i.e. if the corev1.Node object really doesn't exist anymore, then perhaps annotate the v3.Node object and gate the deletion validator behind the presence of the annotation for the local cluster.

jakefhyde (changes requested):

pkg/resources/management.cattle.io/v3/node/validator.go
dharmit: Thanks for the review, @jakefhyde. Your observation is spot on and something I hadn't thought of. If I delete a v1.Node, the change proposed in this PR keeps preventing the deletion of corresponding v3.Node object indefinitely.

In terms of the solution to this problem, are you suggesting annotating the v3.Node when its corresponding v1.Node is deleted, and if a v3.Node with the said annotation is marked for deletion, the webhook should let that through? If yes, this change would go into the rancher code, right? I'm trying to find and understand the relevant code bits for this. I think the file we're interested in is `pkg/controllers/management/node/controller.go`, but let me know if you have any pointers.

jakefhyde: This is where we delete the machine if the downstream node no longer exists: https://github.com/rancher/rancher/blob/b435a2d786c50b03bd1ba7279a6a621ebcd19c84/pkg/controllers/managementuser/nodesyncer/nodessyncer.go#L496. Can I ask why we want to prevent deleting the machine? Reading https://github.com/rancher/rancher/issues/48303:
> At a glance it seems that the issue happens because deletion of local namespace (naturally) triggers deletion of management.cattle.io.nodes
I feel like if we prevent deletion of the namespace, then we don't need to change how deleting machines works.

dharmit: > I feel like if we prevent deletion of the namespace, then we don't need to change how deleting machines works.

Agree. But if the admin directly deletes a node using `kubectl`, that would render the cluster unstable, no? In argument to that, we can always say that an admin should know what they are doing and what its consequences would be.

@snasovich can you chime in on this one?

jakefhyde: > Agree. But if the admin directly deletes a node using `kubectl`, that would render the cluster unstable, no?

This would only be a problem if they deleted all of their nodes, not just one. I also think that's somewhat outside the scope of this issue, it seems to me like it's just about preventing the deletion of the local namespace. I think if you incorporate a change to how node objects are deleted on such a short timetable, the regression chance is going to be too high.

jakefhyde (changes requested):
LGTM, but just wanted to confirm that the agent's `CLUSTER_CLEANUP` is unimpacted.

jakefhyde (approved):

pkg/resources/core/v1/namespace/projectannotations.go
```diff
@@ -23,8 +25,10 @@ type projectNamespaceAdmitter struct {
 	sar authorizationv1.SubjectAccessReviewInterface
 }
 
-// Admit ensures that the user has permission to change the namespace annotation for
-// project membership, effectively moving a project from one namespace to another.
+// Admit ensures that the:
+//   - user has permission to change the namespace annotation for project membership, effectively moving a project from
+//     one namespace to another.
+//   - deletion of `local` and `fleet-local` namespace is not allowed as it could
```
crobby, line 31: as it could......what?

crobby (dismissed):
One small non-blocking nit on a comment, otherwise, lgtm

pkg/resources/core/v1/namespace/projectannotations.go
dharmit: Lol. 😂 
I thought I had removed those words.

dharmit: Done. Removed.

jakefhyde (approved):

crobby (approved):

PR #563, opened 2024-12-16
Fix PR not triggering CI

tomleb:
A PR created with a Github token from a GHA workflow run will not trigger the CI pull-request event so we end up with PR being opened but no CI running. Instead, let's use an app token. (https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#triggering-a-workflow-from-a-workflow)


crobby (approved):
looks reasonable, worth a shot

gehrkefc (approved):
LGTM

PR #572, opened 2024-12-17
[v0.7|main] Make admission rejection message more useful for namespace creation

crobby:
## Issue: 

https://github.com/rancher/rancher/issues/47585

## Problem

Webhook rejection message was not useful in this situation

## Solution

Include user, verb, resource in the rejection message.

## Testing notes
To make this situation show up, you can create a roletemplate like the one below

```
builtin: false
context: project
description: ""
displayName: nomanagenamespace
external: false
hidden: false
kind: RoleTemplate
metadata:
  generateName: rt-
rules:
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - '*'
- apiGroups:
  - management.cattle.io
  resources:
  - projects
  verbs:
  - get
  - list
  - watch
  - updatepsa

```
Then you can create a clusterroletemplatebinding to bind that role to a user that has no other permissions on a given cluster.

With that in place, login as the user to which the crtb was bound and navigate to the cluster and try to create a namespace in a project.

Prior to this fix, the error message provided just said "Unauthorized".
After this fix, the error message will look like:  `admission webhook "rancher.cattle.io.namespaces.create-non-kubesystem" denied the request: User "u-7qhg2" does not have permission "manage-namespaces" on project "p-w7252"`

### DIFF
--- a/pkg/resources/core/v1/namespace/projectannotations.go
+++ b/pkg/resources/core/v1/namespace/projectannotations.go
@@ -87,10 +87,13 @@ func (p *projectNamespaceAdmitter) Admit(request *admission.Request) (*admission
 		return response, nil
 	}
 
+	constructedReason := fmt.Sprintf("User %q does not have permission %q on project %q",
+		request.UserInfo.Username, manageNSVerb, projectName)
+
 	response.Allowed = false
 	response.Result = &metav1.Status{
 		Status:  "Failure",
-		Message: sarResponse.Status.Reason,
+		Message: constructedReason,
 		Reason:  metav1.StatusReasonUnauthorized,
 		Code:    http.StatusForbidden,
 	}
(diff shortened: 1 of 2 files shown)


nflynt (approved):

ericpromislow (changes requested):
Would be good if there could be a unit test for this. I tried messing with the test file, but always get an actual error "error when creating sar, server unavailable" instead of a fabricated one. Not sure if the underlying function is properly mocked.

Looks fine otherwise.

crobby:
> Would be good if there could be a unit test for this. I tried messing with the test file, but always get an actual error "error when creating sar, server unavailable" instead of a fabricated one. Not sure if the underlying function is properly mocked.
> 
> Looks fine otherwise.

I've updated an existing test to check the message for this case.

joshmeranda (approved):

ericpromislow (approved):
LGTM

PR #651, opened 2025-02-10
Prevent deletion of namespaces and cluster

dharmit:
## Issue: 

https://github.com/rancher/rancher/issues/32745
https://github.com/rancher/rancher/issues/48303

## Problem

* Presently, it's possible to delete `local` cluster (both kinds `clusters.provisioning.cattle.io` and `cluster.management.cattle.io`) using `kubectl delete` commands.
* Presently, it's possible to delete `local` and `fleet-local` namespaces using `kubectl delete`.
## Solution

All aforementioned deletions can be prevented using the proposed PR.

```sh
$ k delete clusters.provisioning.cattle.io -n fleet-local local
Error from server (BadRequest): admission webhook "rancher.cattle.io.clusters.provisioning.cattle.io" denied the request: can't delete local cluster

$ k delete cluster.management.cattle.io local
Error from server (BadRequest): admission webhook "rancher.cattle.io.clusters.management.cattle.io" denied the request: cannot delete the local cluster

$ k delete ns local
Error from server (BadRequest): admission webhook "rancher.cattle.io.namespaces.delete-namespace" denied the request: deletion of namespace "local" is not allowed

$ k delete ns fleet-local
Error from server (BadRequest): admission webhook "rancher.cattle.io.namespaces.delete-namespace" denied the request: deletion of namespace "fleet-local" is not allowed
```
## CheckList
  
- [x] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/cluster/validator.go
+++ b/pkg/resources/management.cattle.io/v3/cluster/validator.go
@@ -25,6 +25,7 @@ import (
 )
 
 const (
+	localCluster             = "local"
 	VersionManagementAnno    = "rancher.io/imported-cluster-version-management"
 	VersionManagementSetting = "imported-cluster-version-management"
 )
@@ -87,6 +88,11 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 	oldCluster, newCluster, err := objectsv3.ClusterOldAndNewFromRequest(&request.AdmissionRequest)
 	if err != nil {
 		return nil, fmt.Errorf("failed get old and new clusters from request: %w", err)
+	}
+
+	if request.Operation == admissionv1.Delete && oldCluster.Name == localCluster {
+		// deleting "local" cluster could corrupt the cluster Rancher is deployed in
+		return admission.ResponseBadRequest("local cluster may not be deleted"), nil
 	}
 
 	response, err := a.validateFleetPermissions(request, oldCluster, newCluster)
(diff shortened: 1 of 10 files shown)


pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -97,6 +98,11 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
 		return nil, err
 	}
 
+	if request.Operation == admissionv1.Delete && oldCluster.Name == localCluster {
+		// deleting "local" cluster could corrupt the cluster Rancher is deployed in
+		return admission.ResponseBadRequest("can't delete local cluster"), nil
```
crobby, line 103: non-blocking nit:  "can't delete local cluster" isn't maybe as informative as it could be.  Maybe also include why in the error since I think this could be displayed to the user.  Or maybe change "can't" to "refusing".

pkg/resources/management.cattle.io/v3/cluster/validator.go
```diff
@@ -89,6 +90,11 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		return nil, fmt.Errorf("failed get old and new clusters from request: %w", err)
 	}
 
+	if request.Operation == admissionv1.Delete && oldCluster.Name == localCluster {
+		// deleting "local" cluster could corrupt the cluster Rancher is deployed in
+		return admission.ResponseBadRequest("cannot delete the local cluster"), nil
```
crobby, line 95: non-blocking nit: "cannot delete local cluster" isn't maybe as informative as it could be. Maybe also include why in the error since I think this could be displayed to the user. Or maybe change "cannot" to "refusing".

Error message should be consistent either way.

crobby (commented):
Just a couple comments on the error messages.  If you're passionate about the way they are, I can live with them, but I think they could be better.

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
dharmit: How about copy-pasting what k8s complains when we try to delete `kube-system`:
```
Error from server (Forbidden): namespaces "kube-system" is forbidden: this namespace may not be deleted
```

dharmit: Of course, modify the message for cluster resource instead of namespace.

dharmit:
@crobby I modified the error message for all deletions. PTAL. :)

crobby (approved):

tomleb (approved):

PR #666, opened 2025-02-13
Fix validation.

bigkevmcd:
## Problem

Allows any of userName, userPrincipalName or groupPrincipalName for GRBs.

## Solution

userPrincipalName has been added to allow you to reference non-existent users in the same way as ClusterRoleBindings can.

## CheckList
  
- [ ] Test
  
- [ ] Docs


alegrey91 (approved):

ericpromislow (approved):
Looks good. New test fails when I revert the underlying change, as expected

PR #672, opened 2025-02-17
Update dependencies for k8s 1.32

vatsalparekh:
## CheckList
  
- [ ] Test
  
- [ ] Docs


tomleb (approved):
LGTM, going to assign someone from Collie for the rule resolver changes just in case.

tomleb (changes requested):
Bleh, forgot about the `kube-version` field in the charts. It needs to be bumped as well here: https://github.com/rancher/webhook/blob/main/charts/rancher-webhook/Chart.yaml#L14.

tomleb:
@JonCrowther added you because we're updating the RuleResolving code, I think that's because the new K8s version added a `context.Context` to all of these libraries.

JonCrowther (commented):
The RuleResolver stuff looks good to me, but the unused parameters need to be fixed

crobby (commented):
Looks pretty good.  Seems like there are a few unused imports that are preventing the tests from running.  Everything else seems good.

crobby (dismissed):

JonCrowther (approved):

tomleb (approved):

PR #682, opened 2025-02-19
Validate LDAP/AD authconfigs

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/48071

This PR is complementary to https://github.com/rancher/rancher/pull/48091

## CheckList
  
- [x] Test
  
- [x] Docs


pkg/resources/management.cattle.io/v3/authconfig/validator.go
```diff
@@ -0,0 +1,272 @@
...
+		err = errors.Join(err, field.Forbidden(field.NewPath("groupDNAttribute"), "invalid value"))
+	}
+	if config.GroupMemberUserAttribute != "" && !isValidAttr(config.GroupMemberUserAttribute) {
+		err = errors.Join(err, field.Forbidden(field.NewPath("groupMemberUserAttribute"), "invalid value"))
+	}
+	if config.GroupMemberMappingAttribute != "" && !isValidAttr(config.GroupMemberMappingAttribute) {
+		err = errors.Join(err, field.Forbidden(field.NewPath("groupMemberMappingAttribute"), "invalid value"))
+	}
+
+	if config.UserLoginFilter != "" {
```
bigkevmcd, line 242: I wonder if these shouldn't be moved up to where we check the values so that the errors come out together when logged?

bigkevmcd (dismissed):
Looks ok, probably needs some tests to reflect the documented behaviour?

pkg/resources/management.cattle.io/v3/authconfig/validator.go
```diff
@@ -0,0 +1,272 @@
...
+		err = errors.Join(err, field.Forbidden(field.NewPath("servers"), "at least one server is required"))
+	}
+	if config.TLS && config.Certificate == "" {
+		err = errors.Join(err, field.Forbidden(field.NewPath("certificate"), "certificate is required"))
+	}
+
+	if config.UserSearchAttribute != "" {
+		for _, attr := range strings.Split(config.UserSearchAttribute, "|") {
+			if !isValidAttr(attr) {
+				err = errors.Join(err, field.Forbidden(field.NewPath("userSearchAttribute"), "invalid value"))
```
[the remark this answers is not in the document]
pmatseykanets: Not that tricky really as it's usually two or three separated attributes at most. At the same time it's not a problem to add it.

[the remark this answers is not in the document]
pmatseykanets: We don't trim the whitespace on the rancher side and I don't believe we should as it's a slipper sloap.

pkg/resources/management.cattle.io/v3/authconfig/validator.go
pmatseykanets: I'm not sure I understand what you mean. All of these are independent fields e.g. LoginUserAttribute has nothing to do with UserLoginFilter.

pkg/resources/management.cattle.io/v3/authconfig/validator.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// According to RFC4512, attribute names (descriptors) should adhere to the following syntax
+// descr = keystring
+// keystring = leadkeychar *keychar
+// leadkeychar = ALPHA
+// keychar = ALPHA / DIGIT / HYPHEN
+// See https://datatracker.ietf.org/doc/html/rfc4512#section-1.4
+var validAttr = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\-]*$`)
+
```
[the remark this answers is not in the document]
pmatseykanets: Yep, tests are coming as this is a draft PR.

pkg/resources/management.cattle.io/v3/authconfig/validator.go
bigkevmcd: To clarify, you're using `errors.Join()` which under the covers returns an `errors.joinError` the `.Error()` method basically appends the `.Error()` of each error with `\n`.

It _might_ be nice if the two possible `UserSearchFilter` errors came out on consecutive lines?

pmatseykanets: Right, this is exactly how it's ordered right now. If you have validation errors, at the top you'll see all errors regarding different attributes, while at the bottom all filter errors.

>  if the two possible UserSearchFilter errors came out on consecutive lines?

There can only be one possible error related to UserSearchFilter (or any other field for that matter), the one that says that it's invalid if the expression it contains can't be successfully compiled.
The only exception is `UserSearchAttribute` which contains a delimited list of attribute names but all possible error messages for this field are grouped together(going one after another) L207.

pkg/resources/management.cattle.io/v3/authconfig/validator.go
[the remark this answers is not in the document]
pmatseykanets: Yes we should and we do. It was a draft PR being actively worked on.

pkg/resources/management.cattle.io/v3/authconfig/validator.go
```diff
@@ -0,0 +1,272 @@
...
+		config v3.LdapConfig
+	)
+
+	if err = json.Unmarshal(request.Object.Raw, &config); err != nil {
+		return fmt.Errorf("failed to get %T from request: %w", config, err)
+	}
+
+	if len(config.Servers) < 1 {
+		err = errors.Join(err, field.Forbidden(field.NewPath("servers"), "at least one server is required"))
+	}
```
pmatseykanets, line 122: FYI This check we have in testAndApply on the Rancher side, that's why it was included.

pkg/resources/management.cattle.io/v3/authconfig/validator.go
```diff
@@ -0,0 +1,272 @@
...
+	if config.TLS && config.Certificate == "" {
+		err = errors.Join(err, field.Forbidden(field.NewPath("certificate"), "certificate is required"))
+	}
```
pmatseykanets, line 125: This one doesn't necessarily have to be added in the context of this PR as it (the PR) doesn't aim to cover all possible validations. That can/should be done as part of migrating authconfigs to the public api.
So I'm open to removing this if there are any concerns about it.

go.mod
```diff
@@ -5,6 +5,9 @@ go 1.23.0
 toolchain go1.23.6
 
 replace (
+	// Remove and bump github.com/rancher/rancher/pkg/apis once the corresponding rancher PR is merged.
+	github.com/rancher/rancher/pkg/apis => github.com/pmatseykanets/rancher/pkg/apis v0.0.0-20250218142204-0faa2dd1e363
```
pmatseykanets, line 9: Naturally, this will be handled before merging this PR.

pkg/resources/management.cattle.io/v3/authconfig/validator.go
```diff
@@ -0,0 +1,272 @@
...
+
+// According to RFC4512, attribute names (descriptors) should adhere to the following syntax
+// descr = keystring
+// keystring = leadkeychar *keychar
+// leadkeychar = ALPHA
+// keychar = ALPHA / DIGIT / HYPHEN
+// See https://datatracker.ietf.org/doc/html/rfc4512#section-1.4
+var validAttr = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\-]*$`)
+
+// IsValidAttr returns true is the given attribute name conforms with the syntax defined in RFC4512.
```
JonCrowther, line 269:
```suggestion
// IsValidAttr returns true if the given attribute name conforms with the syntax defined in RFC4512.
```

andreas-kupries (approved):

crobby (approved):
lg.  Seems to do what it says it does and it's nicely documented and tested.

PR #685, opened 2025-02-19
Update renovate.json to ignore packages

rohitsakala:
Issue https://github.com/rancher/rancher/issues/49156

Implemented a part of it as I faced that specific problem.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -13,5 +13,17 @@
     "helm-values",
     "regex",
     "gomod"
+  ],
+  "packageRules": [
+    {
+      "matchPackageNames": [
+        "k8s.io/**",
+        "sigs.k8s.io/**",
+        "go.opentelemetry.io/**",
+        "github.com/prometheus/**",
+        "github.com/rancher/rancher/pkg/apis"
+      ],
+      "enabled": false
+    }
   ]
 }


crobby:
Quick question here, not necessarily a change request:   for webhook, we get norman pulled-in as an indirect dep.  Do we also need to protect against indirect norman bumps to major/minor versions with this?

tomleb:
> Quick question here, not necessarily a change request: for webhook, we get norman pulled-in as an indirect dep. Do we also need to protect against indirect norman bumps to major/minor versions with this?

I'm not sure what the behavior of renovate is wrt indirect dependencies, however looking at the list of renovate PR, I don't see one targeting norman alone. As such, it's likely we don't **need** any rules here for this.

tomleb (approved):
Yeah let's try it out and we'll see.

tomleb:
@rohitsakala yes, I see https://github.com/rancher/webhook/pull/697 is still opened though :thinking: 

`gomod-k8sio-dependencies` is this group: https://github.com/rancher/renovate-config/blob/release/default.json#L304. I wonder if we have to do:

```json
    {
      "enabled": false,
      "groupName": "gomod-k8sio-dependencies"
    },
```

to turn it off

tomleb:
@rohitsakala maybe, we should be able to test it locally. Looking at my bash history, I have the following command I was using for testing renovate:

```bash
docker run --rm -v $PWD:/repo:ro -e LOG_LEVEL=debug   -e GITHUB_COM_TOKEN="$(gh auth token)"  -it --workdir /repo renovate/renovate:slim renovate --platform=local  | tee ok.log
```

IIRC it has some limitations but in the past I was able to see what renovate was going to do.

PR #738, opened 2025-02-27
Validate authconfig only if the auth provider is enabled

pmatseykanets:
## Problem

With #682 we introduced some required validations for LDAP/AD auth providers but we keep authconfigs objects for all auth providers even for disabled ones, which leads to errors like this

```
2025/02/27 18:02:34 [ERROR] failed to call leader func: failed to add authconfig data: admission webhook "rancher.cattle.io.authconfigs.management.cattle.io" denied the request: servers: Forbidden: at least one server is required
```

## Solution

Validate the authconfig only if the auth provider is enabled.

## CheckList
  
- [x] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/authconfig/validator.go
+++ b/pkg/resources/management.cattle.io/v3/authconfig/validator.go
@@ -92,6 +92,11 @@ func (a *admitter) admitUpdate(request *admission.Request, oldAuthConfig, newAut
 func (a *admitter) admitCommonCreateUpdate(request *admission.Request, _, newAuthConfig *v3.AuthConfig) (*admissionv1.AdmissionResponse, error) {
 	var err error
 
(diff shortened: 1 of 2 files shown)


bigkevmcd (approved):
This looks good to me.

I have some changes for rancher that will log out which of the AuthConfigs this refers to:

```
2025/02/27 18:02:34 [ERROR] failed to call leader func: failed to add authconfig data: admission webhook "rancher.cattle.io.authconfigs.management.cattle.io" denied the request: servers: Forbidden: at least one server is required
```

Because otherwise, it's not clear from the error which AuthConfig is causing problems.

Maybe we could include this in the response?

crobby (approved):
Change lgtm.
The integration test fails are definitely not related to this PR.

alegrey91 (approved):

PR #765, opened 2025-03-05
Do not validate PC and PDB settings if they have not been customized

HarrisonWAffel:
## Problem
When the backup restore operator attempts to restore a snapshot taken from a rancher version including the PC and PDB feature, the webhook will attempt to validate the `cluster-agent-default-priority-class` and `cluster-agent-default-pod-disruption-budget` settings even if they have not been customized. This results in a json marshaling error, blocking the request. This issue will only reproduce when attempting to reapply the default settings json through BRO or manually. Manual reproduction steps can be found in the linked BRO issue. 

## Solution
Add a conditional to ensure that the setting is only validated if it has been customized. 

## CheckList
  
- [x] Test
  
- [ ] Docs
   - No documentation updates required


mallardduck (approved):

jiaqiluo (approved):
It seems impossible for `newSetting` to be a nil point. 
LGMT.

crobby (approved):

PR #798, opened 2025-03-12
feat: updatepsa for project level

alegrey91:
## Problem
Currently, updatepsa is only available to Cluster Roles and not Project Roles due to security concern. Customer wants the ability to assign this to a Project Role.

## Solution
Allow user to update PSA when admin grant access to do it.
In this PR, we are adding information to craft a better SAR (SubjectAccessReview) request.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/pkg/resources/core/v1/namespace/psalabels.go
+++ b/pkg/resources/core/v1/namespace/psalabels.go
@@ -73,7 +73,10 @@ func (p *psaLabelAdmitter) Admit(request *admission.Request) (*admissionv1.Admis
 	// so that if we are not able to get them,
 	// the SAR request will be done in any case.
 	if ns.Annotations[projectID] != "" {
-		projectInfo := strings.Split(ns.Annotations[projectID], ":")
+		projectNamespace, projectName, found := strings.Cut(ns.Annotations[projectID], ":")
+		if found {
+		...
+		}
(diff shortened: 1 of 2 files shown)


JonCrowther (approved):
LGTM

pkg/resources/core/v1/namespace/psalabels.go
```diff
@@ -60,13 +68,27 @@ func (p *psaLabelAdmitter) Admit(request *admission.Request) (*admissionv1.Admis
 		extras[k] = v1.ExtraValue(v)
 	}
 
+	var projectNamespace, projectName string
+	// here we are filling the variables above with the projectId,
+	// so that if we are not able to get them,
+	// the SAR request will be done in any case.
+	if ns.Annotations[projectID] != "" {
+		projectInfo := strings.Split(ns.Annotations[projectID], ":")
```
JonCrowther, line 76: Not necessary, but I generally prefer to use `strings.Cut` for this operation:
```suggestion
		projectNamespace, projectName, found := strings.Cut(ns.Annotations[projectID], ":")
		if found {
		...
		}
```

crobby (approved):

JonCrowther (approved):

PR #868, opened 2025-04-08
Don't remove previous RCs

tomleb:
## Issue: https://github.com/rancher/charts-build-scripts/issues/100

Charts should now support multiple RCs so we no longer need to delete old RCs. This will prevent breaking the CI whenever we bump webhook for example (or worse when both webhook and fleet are bumped at the same time).


crobby (approved):
lg, pretty sure it does what it says it does

PR #912, opened 2025-05-12
fix: Invoke webhook for cloud-credential secrets only

pratikjagrut:
## Issue:
 https://github.com/rancher/rancher/issues/49750

## Problem
The `MutatingWebhookConfiguration` for secrets was configured to invoke the Rancher webhook for all secret creation and update operations. However, the webhook's internal mutation logic for secrets is specifically designed to act only upon secrets of type `provisioning.cattle.io/cloud-credential` (e.g., to add the `field.cattle.io/creatorId` annotation).

This broad invocation scope meant that if the webhook service was temporarily unavailable (for instance, during Rancher startup, upgrades, or pod restarts), it could inadvertently block the creation or update of *any* secret. This included critical internal secrets, such as the `api-extension` secret required by the remotedialer-proxy (RDP), leading to Rancher startup failures or other operational issues.

## Solution
This PR modifies the `MutatingWebhookConfiguration` for secrets by adding `matchConditions`. A CEL (Common Expression Language) expression is now used to instruct the Kubernetes API server to only forward admission requests to this webhook if the secret object being processed is of type `provisioning.cattle.io/cloud-credential`.

The specific CEL expression implemented is:
```cel
object != null && object.type == "provisioning.cattle.io/cloud-credential"
```
This ensures that:
1.  The webhook is only invoked for secrets of the intended type.
2.  The condition correctly handles `DELETE` operations where `object` would be `null`, preventing evaluation errors.


crobby (changes requested):
Seems like a pretty straightforward change, but can a test be added for this?

pratikjagrut:
@crobby I added the tests.

crobby (approved):

ericpromislow (approved):
LGTM

pkg/resources/core/v1/secret/mutator.go
```diff
@@ -72,17 +72,18 @@ func (m *Mutator) MutatingWebhook(clientConfig admissionregistrationv1.WebhookCl
 	mutatingWebhook := admission.NewDefaultMutatingWebhook(m, clientConfig, admissionregistrationv1.NamespacedScope, m.Operations())
 	mutatingWebhook.SideEffects = admission.Ptr(admissionregistrationv1.SideEffectClassNoneOnDryRun)
 	mutatingWebhook.TimeoutSeconds = admission.Ptr(int32(15))
+	mutatingWebhook.MatchConditions = []admissionregistrationv1.MatchCondition{
```
tomleb, line 75: The code only checks for cloud-credential for the `CREATE` case:

```go
func (m *Mutator) admitCreate(secret *corev1.Secret, request *admission.Request) (*admissionv1.AdmissionResponse, error) {
	if secret.Type != "provisioning.cattle.io/cloud-credential" {
			Allowed: true,
		}, nil
	}
```

but there's no such check for the delete. However this match condition will be for `CREATE` and `DELETE` operations. Are we sure that the `DELETE` behavior was only for cloud credentials? The tests in mutator_test.go doesn't have a `type` field set so it feels like it could be for other secret types as well.

tomleb (changes requested):
One concern with the `DELETE` case since it's not clear to me that it's meant to be skipped here.

pkg/resources/core/v1/secret/mutator.go
pratikjagrut: Hi Tom,
Thanks for pointing that out.
The delete operation could apply to any secret type.
So, I’ve updated the match expression and added a test to verify that the delete operation behaves as expected.

tomleb (approved):

PR #927, opened 2025-05-27
Don't require certificate if tls is on for ldap/ad authconfigs

pmatseykanets:
## Issue: 

https://github.com/rancher/rancher/issues/50470

## Problem

In #682 we made the certificate required for LDAP/AD auth providers if the TLS is set to on, but it's only required for self-signed certs.

## Solution

Don't require the certificate if the TLS is on for LDAP/AD auth providers.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -97,7 +97,6 @@ If yes, the webhook redacts the role, so that it only grants a deletion permissi
 When an LDAP (`openldap`, `freeipa`) or ActiveDirectory (`activedirectory`) authconfig is created or updated, the following checks take place:
 
 - The field `servers` is required.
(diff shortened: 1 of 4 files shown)


pmatseykanets:
> Surprised no tests needed changes?

They did change. The test cases were already there. The outcome changed for these.
- https://github.com/rancher/webhook/pull/927/files#diff-8468a668b39e083c53040b492f543a9abf46299a3a19a112802f748f4e276ec0R84
- https://github.com/rancher/webhook/pull/927/files#diff-8468a668b39e083c53040b492f543a9abf46299a3a19a112802f748f4e276ec0R288

PR #932, opened 2025-06-02
:seedling:  Allow creation of `fleetworkspace` for existing namespace when annotation is present and set to true

yiannistri:
This PR is part of a fix for https://github.com/rancher/turtles/issues/1329
## Problem
When a `fleetworkspace` gets created, the webhook code in this repo also creates a corresponding namespace. This presents a problem when there is a need to create a `fleetworkspace` for an existing namespace as the webhook does not currently allow this. This PR adds an annotation that allows a `fleetworkspace` to be created for an existing namespace when this annotation is present.

## Solution

When a specific annotation is present and set to true for an existing namespace, then the webhook allows the creation of a `fleetworkspace` resource.

## CheckList
  
- [x] Test
  
- [ ] Docs


crobby (approved):
lgtm

PR #943, opened 2025-06-10
Add update validation for User.UserName

JonCrowther:
## Problem

UserName was immutable in Norman, but with public API we need a webhook to do that validation.

## Solution

Added a new validation. It's possible to add a UserName if none existed, but not to change an existing one.

## CheckList
  
- [x] Test
  
- [x] Docs


crobby (approved):
lgtm

PR #985, opened 2025-07-17
[2.13] add validation for etcd s3 cloud credential 

jiaqiluo:
## Issue: 

https://github.com/rancher/rancher/issues/50987

## Problem

While creating a node driver or custom RKE2/k3s cluster, if the secret used for saving etcd backups to S3 does not exist, then the cluster will be stuck waiting for node ref, and Rnacher will generate empty plans for nodes.

Note that this bug can only be triggered when creating or updating the cluster programmatically or via editing YAML.

## Solution

Add validation for the etcd S3 cloud credential to reject the request for creating or updating the provisioning cluster if the provided secret does not exist. 

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
@@ -157,7 +157,7 @@ func (p *provisioningAdmitter) Admit(request *admission.Request) (*admissionv1.A
 			return response, err
 		}
 
-		if response, err = p.validateS3Secret(cluster); err != nil || !response.Allowed {
+		if response, err = p.validateS3Secret(oldCluster, cluster); err != nil || !response.Allowed {
 			return response, err
 		}
 	}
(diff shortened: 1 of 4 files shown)


jakefhyde (changes requested):

pkg/resources/provisioning.cattle.io/v1/cluster/validator.go
```diff
@@ -697,6 +703,51 @@ func (p *provisioningAdmitter) validatePodDisruptionBudget(oldCluster, cluster *
 	return admission.ResponseAllowed(), nil
 }
 
+// validateS3Secret checks if the S3 cloud credential secret referenced in the cluster exists.
+func (p *provisioningAdmitter) validateS3Secret(oldCluster, cluster *v1.Cluster) (*admissionv1.AdmissionResponse, error) {
+	if cluster.Name == localCluster {
```
jakefhyde, line 708: I think you need a check here `if oldCluster == nil`, so that you still validate on create, otherwise it will likely panic.

jakefhyde (changes requested):

jakefhyde (approved):

PR #998, opened 2025-07-23
[main] Label admission-configuration-psact secret to be backed up

jbiers:
## Issue: https://github.com/rancher/backup-restore-operator/issues/663


crobby (approved):
Change looks fine.  Could use a test, but I won't block on that.

PR #1013, opened 2025-08-02
[v0.7] Remove use of Dapper for go generate

tomleb:
# Context

We used to need dapper to be able to run `go generate ./...` in the release action of rancher/rancher. This was to make sure we install the same dependencies (controller-gen) as the one used for r/r. With https://github.com/rancher/rancher/pull/51308, this is no longer needed, we can just run `go generate ./...` and `go tools` will install the tool at the appropriate version.


tomleb:
Closing and re-opening to unstuck CI.

joshmeranda (approved):

PR #1015, opened 2025-08-06
validate password for local users

raulcabello:
## Problem
Users have been migrated to the public API which means that users can be created with kubectl bypassing norman validations.
User passwords are encrypted with pbkfd2. This encryption will be done by the webhook unless the user or Rancher already encrypted it. In this case the `cattle.io/password-hash` annotation must be provided.

## Solution
Move Norman validations to the webhook:
- User can't delete or deactivate himself.
- 2 users can't be created with the same username
- Password must have min length
- Password can't match the user name

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/pkg/resources/core/v1/secret/Secret.md
+++ b/pkg/resources/core/v1/secret/Secret.md
@@ -12,6 +12,11 @@ places a `field.cattle.io/creatorId` annotation with the name of the user as the
 
 If `field.cattle.io/no-creator-rbac` annotation is set, `field.cattle.io/creatorId` does not get set.
 
+For secrets that holds local users passwords, which are stored in the `cattle-local-user-passwords` namespace:
+- Verifies the password has the minimum required length.
+- The password is not the same as the username.
+- It encrypts the password using pbkdf2 
+
 ### On delete
 
 Checks if there are any RoleBindings owned by this secret which provide access to a role granting access to this secret.
--- a/pkg/resources/core/v1/secret/mutator.go
+++ b/pkg/resources/core/v1/secret/mutator.go
@@ -1,9 +1,15 @@
 package secret
 
 import (
+	"crypto/pbkdf2"
+	"crypto/rand"
+	"crypto/sha3"
 	"fmt"
+	"strconv"
+	"unicode/utf8"
 
 	"github.com/rancher/webhook/pkg/admission"
+	ctrlv3 "github.com/rancher/webhook/pkg/generated/controllers/management.cattle.io/v3"
 	objectsv1 "github.com/rancher/webhook/pkg/generated/objects/core/v1"
 	"github.com/rancher/webhook/pkg/patch"
 	"github.com/rancher/webhook/pkg/resources/common"
@@ -14,6 +20,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/utils/trace"
 )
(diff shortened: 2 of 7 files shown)


pkg/resources/core/v1/secret/Secret.md
```diff
@@ -12,6 +12,11 @@ places a `field.cattle.io/creatorId` annotation with the name of the user as the
 
 If `field.cattle.io/no-creator-rbac` annotation is set, `field.cattle.io/creatorId` does not get set.
 
+For secrets that holds local users passwords, which are stored in the `cattle-local-user-passwords` namespace:
+- Verifies the password has the minimum required length.
+- The password is not the same as the username.
+- It encrypts the password using pbkdf2 
+
```
JonCrowther, line 19:
```suggestion
For secrets stored in the `cattle-local-user-passwords` namespace containing local users passwords:
- Verifies the password has the minimum required length.
- Verifies the password is not the same as the username.
- Encrypts the password using pbkdf2 
```

pkg/resources/core/v1/secret/mutator.go
```diff
@@ -75,7 +97,7 @@ func (m *Mutator) MutatingWebhook(clientConfig admissionregistrationv1.WebhookCl
 	mutatingWebhook.MatchConditions = []admissionregistrationv1.MatchCondition{
 		{
 			Name:       "filter-by-secret-type-cloud-credential",
-			Expression: `request.operation == 'DELETE' || (object != null && object.type == "provisioning.cattle.io/cloud-credential")`,
+			Expression: `request.operation == 'DELETE' || (object != null && object.type == "provisioning.cattle.io/cloud-credential" && request.operation == 'CREATE') || (object != null && object.metadata.namespace == "` + localUserPasswordsNamespace + `")`,
```
JonCrowther, line 100: Maybe some line breaks to make this a little more readable?
```suggestion
			Expression: `request.operation == 'DELETE' ||
			(object != null && object.type == "provisioning.cattle.io/cloud-credential" && request.operation == 'CREATE') || 
			(object != null && object.metadata.namespace == "` + localUserPasswordsNamespace + `")`,
```

pkg/resources/core/v1/secret/mutator.go
```diff
@@ -179,3 +208,98 @@ func amendRulesToOnlyPermitDelete(rules []rbacv1.PolicyRule, secretName string)
 	}
 	return rules, amended
 }
+
+// admitLocalUserPassword handle the secrets that contains the local user passwords, which are stored in the cattle-local-user-passwords namespace.
+// If the annotation ccattle.io/password-hash is not present in the secret, the webhook will encrypt it using pbkdf2. The secret is mutated to include the hashed password, the salt and the user as owner reference.
```
JonCrowther, line 213:
```suggestion
// admitLocalUserPassword handles the secrets that contains the local user passwords, which are stored in the cattle-local-user-passwords namespace.
// If the annotation cattle.io/password-hash is not present in the secret, the webhook will encrypt it using pbkdf2. The secret is mutated to include the hashed password, the salt and the user as owner reference.
```

pkg/resources/core/v1/secret/mutator_test.go
```diff
@@ -439,3 +443,268 @@ func addRoleRefToBinding(role rbacv1.Role, binding rbacv1.RoleBinding) *rbacv1.R
...
+			mockUserCache: func() ctrlv3.UserCache {
+				mock := fake.NewMockNonNamespacedCacheInterface[*v3.User](ctrl)
+				mock.EXPECT().Get("test-user").Return(fakeUser, nil)
+
+				return mock
+			},
+			wantPatch:   `[{"op":"add","path":"/metadata/ownerReferences","value":[{"apiVersion":"","kind":"","name":"test-user","uid":""}]},{"op":"add","path":"/metadata/annotations","value":{"cattle.io/password-hash":"pbkdf2sha3512"}},{"op":"replace","path":"/data/password","value":"aGFzaGVkUGFzc3dvcmQ="},{"op":"add","path":"/data/salt","value":"c2FsdA=="}]`, // aGFzaGVkUGFzc3dvcmQ= -> hashedPassword base64 encoded, and c2FsdA==" => salt base64 encoded
+			wantAllowed: true,
+		},
+		"password was already hashed": {
```
JonCrowther, line 519: Could you add a test that also checks that a password hashed with `bcrypt` is allowed?

JonCrowther (commented):
Logic LGTM. A few grammar nits and test suggestions

pkg/resources/core/v1/secret/mutator_test.go
raulcabello: done in https://github.com/rancher/webhook/pull/1015/commits/30d85f0840b6629c47837ac3df7d2cfa0afa08c4

PR #1111, opened 2025-09-19
Fix crash in mutator when rancherd patches local cluster

jakefhyde:
## Problem

https://github.com/rancher/webhook/pull/410 introduced data directory validation to the webhook, designed to prevent users from changing data directories once the cluster has been created. Harvester leverages rancherd, which is effectively a self-bootstrapping mechanism for creating the Rancher local cluster. Part of this bootstrapping involves patching the local cluster to set the rkeConfig: {}, which result in Rancher reconciling it like any other normal CAPI cluster.

Previously, the webhook did not come up quick enough in the rancherd process, but now the webhook and it's related mutatingwebhookconfiguration exists before the cluster can be patched. The data directory validation causes the webhook to panic (and reject the request), as it was not expected that a cluster would go from an rkeConfig with a nil value, to one without one.

## Solution

if `oldCluster.Spec.RKEConfig == nil`, accept the request.

## CheckList
  
- [x] Test
  
- [x] Docs - No need to update docs on this one I believe.

### DIFF
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
@@ -194,7 +194,7 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 // dynamicSchemaSpec field for a machine pool if the "provisioning.cattle.io/allow-dynamic-schema-drop" annotation is
 // not present and true on the cluster. If the value of the annotation is true, no mutation is performed.
 func (m *ProvisioningClusterMutator) handleDynamicSchemaDrop(request *admission.Request, oldCluster, cluster *v1.Cluster) *admissionv1.AdmissionResponse {
-	if cluster.Name == "local" || cluster.Spec.RKEConfig == nil {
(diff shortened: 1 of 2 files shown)


pkg/resources/provisioning.cattle.io/v1/cluster/mutator.go
```diff
@@ -194,7 +194,7 @@ func (m *ProvisioningClusterMutator) Admit(request *admission.Request) (*admissi
 // dynamicSchemaSpec field for a machine pool if the "provisioning.cattle.io/allow-dynamic-schema-drop" annotation is
 // not present and true on the cluster. If the value of the annotation is true, no mutation is performed.
 func (m *ProvisioningClusterMutator) handleDynamicSchemaDrop(request *admission.Request, oldCluster, cluster *v1.Cluster) *admissionv1.AdmissionResponse {
-	if cluster.Name == "local" || cluster.Spec.RKEConfig == nil {
+	if cluster.Name == "local" || oldCluster.Spec.RKEConfig == nil || cluster.Spec.RKEConfig == nil {
```
crobby, line 197: Looks like the new test causes this to panic.  Do we need to also check oldCluster.Spec == nil?

crobby (approved):

PR #1118, opened 2025-09-25
validate the new project state for consistency, ignore old state

andreas-kupries:
## Issue: 

https://github.com/rancher/rancher/issues/49041

## Problem

The webhook checked new quotas against old used limit information.
This ignored the possibility that the users may supply bogus used-limit data, for example on project creation.
And further ignored that parts of the system make changes to used-limits.
With bogus used-limit in the project setting this back to good information fails because of the validation failure triggered by the use of the old bad data.

## Solution

The webhook now checks the entire new project state, quotas and used-limit for consistency.

## CheckList
  
- [x] Test - additional unit tests
  
- [x] Docs - no change, behaviour was wrong, not matching the docs


crobby (approved):
looks sensible

PR #1141, opened 2025-10-15
Add GHA to publish head images

crobby:
Issue:  https://github.com/rancher/rancher/issues/52337


tomleb (approved):
LGTM.

This shouldn't be able to mistakenly override an image tag already existing since we're hardcoding head or a -head suffix.

One nit but that's it, feel free to ignore.

PR #1152, opened 2025-10-30
[backport v2.12] bug webhook error msg

marytlf:
Issue: https://github.com/rancher/rancher/issues/48484
### Problem
The Rancher webhook can generate a large volume of http: TLS handshake error logs when a client attempts to connect with a missing or invalid certificate. While these are often non-critical, they can overwhelm the logs, making it difficult to debug genuine issues. The current implementation doesn't provide a way to suppress these logs without impacting other important error messages. The community has requested a solution to reduce log noise caused by these specific errors, as detailed in issue #48484.

### Solution
`http: TLS handshake error` messages are downgraded from error to debug level. This means the messages will only appear in the logs if the CATTLE_DEBUG environment variable is set to true

### Test Scenarios
**Scenario 1**:` CATTLE_DEBUG = false`
Result: No TLS handshake error messages are logged. The messages are logged at the debug level but are not displayed because CATTLE_DEBUG is false.

**Scenario 2**: `CATTLE_DEBUG = true`

Result: TLS handshake error messages are logged at the debug level, as expected.

Log Example: level=debug msg="2025/08/21 18:03:25 http: TLS handshake error from 172.31.7.103:40900: EOF"

**How to reproduce and force the error:**
NOTE: The error happens when using rke2, not k3s.

Set the environment variables on rancher-webhook deployment (kubectl -n cattle-system edit deploy rancher-webhook)
CATTLE_DEBUG = "true"/"false"
Login on Rancher UI (or create/delete a new user)
Check the logs from rancher-webhook (kubectl logs -l app=rancher-webhook -n cattle-system)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,7 +42,7 @@ require (
 	github.com/evanphx/json-patch v5.9.11+incompatible
 	github.com/go-ldap/ldap/v3 v3.4.10
 	github.com/gorilla/mux v1.8.1
-	github.com/rancher/dynamiclistener v0.7.0
+	github.com/rancher/dynamiclistener v0.7.2-rc1
 	github.com/rancher/jsonpath v0.0.0-20250620213443-ad24535cf0c1
 	github.com/rancher/lasso v0.2.4
 	github.com/rancher/rancher/pkg/apis v0.0.0-20250919165148-49b3f67fb245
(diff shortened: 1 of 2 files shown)


crobby (approved):

marytlf:
backport v2.12: https://github.com/rancher/rancher/issues/52536

PR #1192, opened 2025-11-25
bring quota conversion function up to par with work done in rancher/rancher

andreas-kupries:
## Issue: 

Companion to https://github.com/rancher/rancher/pull/52773

Brings the changes to the conversion functions in r/r to the webhook as well.
The code is currently a copy instead of calling the unified function in r/r.
This works, is however also a tech debt.
Unification requires figuring out how to import `rancher/pkg/resourcequota` into the webhook.

## Problem

The webhook has to properly convert a rancher resource quota into a kube resource list, taking the new field `extended` into account. The webhook's copy of equivalent functions in r/r was behind the times here.

## Solution

For now made a copy of the unified r/r function in the webhook.

## CheckList
- [x] Test
- [x] Docs - No behavioral change. Extension to handle new field properly

### DIFF
--- a/pkg/resources/management.cattle.io/v3/project/mutator_test.go
+++ b/pkg/resources/management.cattle.io/v3/project/mutator_test.go
@@ -29,6 +29,14 @@ var (
 		},
 		Spec: v3.ProjectSpec{
 			ClusterName: "testcluster",
+			ResourceQuota: &v3.ProjectResourceQuota{
+				Limit: v3.ResourceQuotaLimit{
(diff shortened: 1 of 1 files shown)


bigkevmcd:
I think we should have a test that asserts that the resource is passed through correctly?

andreas-kupries:
> > I think we should have a test that asserts that the resource is passed through correctly?
> 
> Yes, that is one of the todos

And done (copied from the equivalent r/r tests)

PR #702, opened 2025-02-20
(feat): Cattle Cluster Agent PDB and PC Validation

HarrisonWAffel:
## Issue: https://github.com/rancher/rancher/issues/48995
depends on https://github.com/rancher/rancher/pull/49107

## Problem
A new feature is being introduced to Rancher which allows users to define a Pod Disruption Budget and Priority Class on the v1.Cluster and v3.Cluster objects to better ensure HA of the cattle cluster agent. By default this feature will be disabled in v2.11. In order to prevent use of the new fields added to the cluster objects when the feature is disabled, as well as to prevent incorrect configuration of the PC and PDB, additional validation needs to be added to the webhook.

## Solution
Add additional logic to the validators for the v1.Cluster, v3.Cluster, and settings resources. 

## CheckList
  
- [x] Test
  
- [x] Docs

PR #1195, opened 2025-12-09
Validate PodDisruptionBudget and PriorityClass for fleetAgentDeploymentCustomization

p-se:
## Problem

A new feature is being introduced to Rancher which allows users to define a Pod Disruption Budget and Priority Class on the v1.Cluster and v3.Cluster objects to better ensure HA of the fleet agent. In order to prevent use of the new fields added to the cluster objects when the feature is disabled, as well as to prevent incorrect configuration of the PC and PDB, additional validation needs to be added to the webhook.

https://github.com/rancher/webhook/pull/702 implemented the same checks for the cattle cluster agent.

## Solution

Add additional logic to the validators for the v1.Cluster, v3.Cluster, and settings resources.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -544,10 +544,10 @@ When settings are updated, the following additional checks take place:
   annotation `cattle.io/force=true`.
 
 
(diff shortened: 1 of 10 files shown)


crobby:
> @crobby please add (someone from) Fleet team as reviewer. I can't for some reason 🤷🏻 😞

Any suggestions?  I'm not sure who would be the best choice.

PR #1208, opened 2025-12-29
[main] add validating webhook for `machinedeployments/scale` resource for autoscaling side-effect 

thatmidwesterncoder:
## Summary

This PR implements a validating webhook with a side effect to handle scale-down operations for Rancher-managed machine deployments, eliminating a race condition that can occur when a `scale` operation occurs on a CAPI MachineDeployment object.

## Problem

During scale-down operations, a race condition can occur when the controller that synchronizes the CAPI MachineDeployment object's `replicas` field back to the provisioning cluster's matching rkeMachinePool happens too late. This causes the following problematic behavior:

1. A scale-down operation is initiated on a machine deployment
2. The CAPI MachineDeployment is updated in etcd
3. Before the sync controller can propagate the replica count change to the Rancher Provisioning cluster's rkeMachinePool, Rancher provisioning controllers may fire
4. This results in new nodes being spun up immediately after node deletion
5. Eventual consistency is eventually achieved, but only after multiple failed attempts
6. This creates a poor user experience with delayed and redundant node provisioning

## Solution

__tl;dr:__ moving the logic [removed here](https://github.com/rancher/rancher/pull/53213) to the webhook, it just looks a little different due to validating webhook-isms + not being able to use the CAPI controllers due to the lack of a deferred start mechanism in this codebase.

---

The PR moves the synchronization logic that updates the Rancher Provisioning cluster object from a controller-based approach to a validating webhook with a side effect. This approach:

1. Intercepts scale requests before the MachineDeployment is committed to etcd
2. Updates the Rancher Provisioning cluster's matching rkeMachinePool to match the CAPI MachineDeployment replica count
3. Only admits the scale request after confirming the machinedeployment either doesn't match a rancher provisioning cluster OR updating the provisisioning cluster to match

By processing the scale request through a validating webhook with a side effect, we guarantee that the Rancher object matches the CAPI object before the MachineDeployment even hits etcd. This completely prevents the race condition from occurring.

## Notes:
- Scale requests will fail if the Rancher Provisioning cluster update cannot be completed
    - This may be the correct behavior in scenarios where nodes are provisioned by Rancher - as we are the infra provider in CAPI terms
- There is a slight update to the way a URL is generated to handle subresources such as `machinedeployments/scale` in this case
- [this rancher PR](https://github.com/rancher/rancher/pull/53213) will need to be merged AFTER this is merged, webhook tagged and bumped in Rancher

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
+++ b/pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
@@ -1,6 +1,7 @@
 package machinedeployment
 
 import (
+	"encoding/json"
 	"fmt"
 
 	provv1 "github.com/rancher/rancher/pkg/apis/provisioning.cattle.io/v1"
@@ -8,13 +9,14 @@ import (
 	provcontrollers "github.com/rancher/webhook/pkg/generated/controllers/provisioning.cattle.io/v1"
 	scaling "github.com/rancher/webhook/pkg/generated/objects/autoscaling/v1"
 	"github.com/sirupsen/logrus"
+	jsonpatch "gopkg.in/evanphx/json-patch.v4"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/util/retry"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/trace"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 )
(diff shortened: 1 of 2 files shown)


HarrisonWAffel (approved):
This makes sense. Though I'm not a big fan of having a validating webhook with a side effect like this, I can see how it prevents machine thrashing. Is there anywhere we can document this behavior outside of the webhook to help future debugging? Maybe in `autoscaler.go` in r/r?

pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
```diff
@@ -0,0 +1,296 @@
...
+//
+// Returns:
+// - (*capi.Cluster, nil) if the CAPI cluster is found
+// - (nil, error) if the cluster is not found or lookup fails
+func (v *ReplicaValidator) fetchCAPICluster(md *capi.MachineDeployment) (*capi.Cluster, error) {
+	clusterName := md.Spec.Template.ObjectMeta.Labels[capi.ClusterNameLabel]
+
+	if clusterName == "" {
+		logrus.Debugf("MachineDeployment %s/%s has no CAPI cluster name label", md.Namespace, md.Name)
+		return nil, apierrors.NewNotFound(schema.GroupResource{Group: "cluster.x-k8s.io", Resource: "clusters"}, "")
```
jakefhyde: Somewhat pedantic but you can just use this: https://github.com/kubernetes-sigs/cluster-api/blob/main/api/core/v1beta1/groupversion_info.go#L27

pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
```diff
@@ -0,0 +1,296 @@
...
+
+	logrus.Debugf("Machine pool %s not found in cluster %s/%s, skipping sync", machinePoolName, cluster.Namespace, cluster.Name)
+	return cluster, false
+}
+
+// fetchMachinePoolName extracts the RKE machine pool name from the MachineDeployment's labels.
+//
+// Returns:
+// - (string) the RKE machine pool name label
+func (v *ReplicaValidator) fetchMachinePoolName(md *capi.MachineDeployment) string {
```
jakefhyde: Nit: I don't know if this really needs to be a function if it's only used in one place.

pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// setMachinePoolQuantity finds the machine pool by name and updates its quantity
+// if it differs from the target replicas.
+//
+// Returns:
+// - (modifiedCluster, true) if update was needed and performed
+// - (cluster, false) if no update was needed or pool not found
+func (v *ReplicaValidator) setMachinePoolQuantity(md *capi.MachineDeployment, cluster *provv1.Cluster, targetReplicas int32) (*provv1.Cluster, bool) {
+	cluster = cluster.DeepCopy()
```
jakefhyde: Nit: move this closer to where it's required, i.e. just above the `for` loop.

pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
```diff
@@ -0,0 +1,296 @@
...
+		if apierrors.IsNotFound(err) {
+			return admission.ResponseAllowed(), nil
+		}
+
+		logrus.Errorf("Failed to sync machine pool replicas for MachineDeployment %s/%s: %v",
+			machineDeployment.Namespace, machineDeployment.Name, err)
+		return admission.ResponseBadRequest(err.Error()), err
+	}
+
+	return admission.ResponseAllowed(), nil
```
jakefhyde:
```suggestion
	err = v.reconcileMachinePoolReplicas(machineDeployment, scale.Spec.Replicas)
	if err == nil || apierrors.IsNotFound(err) {
		// prov/CAPI cluster do not exist or the machinedeployment isn't managed by rancher
		return admission.ResponseAllowed(), nil
	}

	logrus.Errorf("Failed to sync machine pool replicas for MachineDeployment %s/%s: %v", machineDeployment.Namespace, machineDeployment.Name, err)
	return admission.ResponseBadRequest(err.Error()), err
```

pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
```diff
@@ -0,0 +1,296 @@
...
+//  1. Looks up the CAPI cluster
+//  2. Finds the Rancher Provisioning Cluster through owner references on the CAPI Cluster
+//  3. Updates the machine pool's quantity to match the target replicas
+//  4. Updates the cluster object with the new machinepool in k8s
+//
+// Returns:
+// - nil if update was successful
+// - error if cluster or machine pool not found, or update fails after retries
+func (v *ReplicaValidator) reconcileMachinePoolReplicas(md *capi.MachineDeployment, targetReplicas int32) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
jakefhyde: In concept you should be able to avoid this with a `PATCH`.

pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
```diff
@@ -0,0 +1,296 @@
...
+			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstr.Object, md); err != nil {
+				logrus.Errorf("Failed to convert unstructured to MachineDeployment: %v", err)
+				return admission.ResponseBadRequest(fmt.Sprintf("failed to convert object: %v", err)), nil
+			}
+			machineDeployment = md
+		} else {
+			logrus.Errorf("Unexpected type from dynamic Get: %T", machineDeploymentObj)
+			return admission.ResponseBadRequest(fmt.Sprintf("unexpected object type: %T", machineDeploymentObj)), nil
+		}
+	}
```
jakefhyde: This too. Honestly maybe we should wrap this whole thing in a generic function (I can help with that if needed).

pkg/resources/cluster.x-k8s.io/v1beta1/machinedeployment/validator.go
jakefhyde: Ah my bad, probably more relevant for line 61 then.

crobby (approved):
Looks good.  I think we're ok deferring the integration tests until we have a mechanism in place to deal with non-Rancher CRDs

PR #1209, opened 2025-12-30
Fix webhook configuration not being applied

tomleb:
# Issue https://github.com/rancher/rancher/issues/53219

This PR fixes a couple of issues with the integration tests and webhook:
- It's possible for the controllers to start, run for all secrets (even the webhook one) and not be leader yet, and then once the pod becomes leader... nothing, the configuration aren't created/updated
- The integration test re-deploy webhook a couple times but doesn't wait for the potentially new webhook configuration (or even the first one) to be properly created/updated.

I also removed the `set -e` and `set +e` in the integration test script. `set -e` is called once and try will not fail on first try if the command it needs to run fails.

Also, to make it easier to debug future issues, I added a step at the end that runs on failure and outputs logs + the configurations. This can be expanded further in another PR to add more things we want to debug.

### DIFF
--- a/.github/workflows/scripts/integration-test-ci
+++ b/.github/workflows/scripts/integration-test-ci
@@ -15,7 +15,6 @@ set -o pipefail
 
 source ./.github/workflows/scripts/try.sh
 
-set +e
 
 # Wait for rancher to start up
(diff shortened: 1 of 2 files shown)


crobby (approved):
Evening here seems reasonable to me

jferrazbr (approved):

PR #1225, opened 2026-01-12
add `dnsPolicy:ClusterFirstWithHostNet` to hostNetwork

susesamu:


ericpromislow (approved):
Looks good, tests pass (eventually). I did get an unrelated failure on a "dirty" k8s deployment, so I refreshed and tests pass.

Note that the tests are stateful -- they take much more time to run on a clean k8s deployment to build the docker images.

After that both first and subsequent runs pass each time.

PR #1340, opened 2026-04-06
fix: Prevent duplicate ProjectRoleTemplateBinding creation

pratikjagrut:
## Problem
Multiple ProjectRoleTemplateBindings (PRTBs) are created for the same user and role within the same project. There is no backend validation to enforce uniqueness of the following tuple: (principalId, roleTemplateId, projectId)
Repeated requests (via UI or API) to assign the same user the same role in the same project result in multiple identical PRTBs being created.

## Solution
Added a new mutating admission webhook for ProjectRoleTemplateBinding CREATE requests that computes a deterministic metadata.name based on the binding's content, replacing whatever name or generateName the client provided:
```
name = "prtb-" + lowercase(base32(sha256(subject + "/" + roleTemplateName + "/" + projectName))[:10])
```
- The subject is resolved using the same priority order as the existing validator: UserPrincipalName → UserName → GroupPrincipalName → GroupName → ServiceAccount.
- Uses base32.StdEncoding without padding, lowercased — all characters are valid in K8s resource names.
10 characters of base32 encode 50 bits of entropy (~1/2^50 collision probability).
- Two identical concurrent requests now produce the same resource name, so the K8s API server rejects the second with a 409 Conflict at the etcd level — a race-free guarantee.
- If the name already matches the deterministic value (idempotent), no patch is emitted.
- If no subject can be resolved, the mutator passes through and lets the validating webhook reject the request.

This PR also  introduces a backend validation mechanism to prevent the creation of duplicate PRTB.

- It efficiently lists all existing PRTBs for the target cluster using a cached lister.
- It iterates through the existing bindings and checks if one already exists with the same roleTemplateName and the same subject (matching by userName, userPrincipalID, groupName, or groupPrincipalName).
- If a non-deleted duplicate is found, the API rejects the request with a 409 Conflict error and a descriptive message.
- To solve a race condition observed in the UI (where a user is removed and re-added in the same transaction), the validator checks the DeletionTimestamp of any conflicting binding. If a duplicate is found but is already marked for deletion, it is ignored (continue), allowing the creation to proceed and preventing a transient UI error.

## CheckList
- [x] Test — Added unit tests for GenerateDeterministicName (7 subtests) and Mutator.Admit (11 subtests) covering all mutation paths, subject priority, idempotency, and error handling. Fixed existing flaky setting validator tests.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -495,6 +495,48 @@ changed:
 
 In addition, as in the create validation, both a user subject and a group subject cannot be specified.
 
+#### Duplicate ProjectRoleTemplateBinding Prevention
+
+On creation, the webhook prevents the creation of a `ProjectRoleTemplateBinding` if another one already exists with the same subject and role in the same project.
+This ensures that a user or group is not granted the same project-level role multiple times.
+
+A binding is considered a duplicate if another non-deleting `ProjectRoleTemplateBinding` exists in the same namespace with the exact same values for:
+- `roleTemplateName`
+- The subject, which is determined by one of the following fields:
+  - `userName`
+  - `userPrincipalName`
+  - `groupName`
+  - `groupPrincipalName`
+
+If a conflicting binding exists but is already marked for deletion (has a non-nil `DeletionTimestamp`), it is ignored, allowing the creation to proceed.
+
+### Mutation Checks
+
+#### On create
+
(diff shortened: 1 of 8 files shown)


crobby (changes requested):
Looks pretty reasonable, but the tests are failing.
Also, you should update the doc in ProjectRoleTemplateBinding.md and run the generator to update the overall docs.

crobby (approved):
LGTM, thanks for adding the docs and having tests.

andreas-kupries (approved):

PR #1361, opened 2026-04-10
feat: add mutating webook for generating names for CRTB

pratikjagrut:
## Problem

Duplicate CRTBs created if the same user is assigned the same role multiple times during downstream cluster provisioning.

## Solution

This PR adds a **mutating admission webhook** for `ClusterRoleTemplateBinding` resources that replaces the server-side random name generation with a **deterministic name** derived from the binding's content.

On `CREATE`, when `metadata.generateName` is set and `metadata.name` is empty (i.e. the caller expects a generated name), the mutator:

1. Determines the **subject** using the priority order: `userPrincipalName` → `userName` → `groupPrincipalName` → `groupName`.
2. Computes a deterministic name: `<generateName prefix> + lowercase(base32(sha256(subject + "/" + roleTemplateName + "/" + clusterName))[:10])`.
3. Sets `metadata.name` to this value and clears `metadata.generateName`.
4. Returns a JSON patch in the mutating response.

Two identical concurrent requests now produce the **same `metadata.name`**, causing the Kubernetes API server to reject the second with a **409 Conflict** — preventing duplicate CRTBs at the API level.

**Behavior boundaries:**
- If `metadata.name` is already set (i.e. the caller explicitly chose a name), **no mutation is performed**. This preserves backward compatibility for automations and the public API.
- If both `metadata.name` and `metadata.generateName` are empty, the request is **rejected** with a descriptive error.
- If no subject field is set, the request is **rejected** (the downstream validating webhook would also catch this, but failing early is clearer).

## CheckList

- [x] Test — Unit tests added in `mutator_test.go` covering: deterministic name generation, different inputs producing different names, valid K8s name characters, subject priority ordering, pass-through for explicit names, rejection for missing subjects, rejection for empty name+generateName, pass-through for non-CREATE operations, and idempotency of identical requests.
- [x] Docs — `ClusterRoleTemplateBinding.md` and `docs.md` updated to document the new mutating webhook behavior (deterministic name generation) and the duplicate prevention validation.

### DIFF
--- a/pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/mutator.go
+++ b/pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/mutator.go
@@ -1,0 +1,107 @@
+package clusterroletemplatebinding
+
+import (
+	"crypto/sha256"
+	"encoding/base32"
+	"fmt"
+	"strings"
+
+	"github.com/rancher/webhook/pkg/admission"
+	objectsv3 "github.com/rancher/webhook/pkg/generated/objects/management.cattle.io/v3"
+	"github.com/rancher/webhook/pkg/patch"
+	admissionv1 "k8s.io/api/admission/v1"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/utils/trace"
+)
+
+// Mutator implements admission.MutatingAdmissionWebhook.
+type Mutator struct{}
+
+// NewMutator returns a new mutator for ClusterRoleTemplateBindings.
+func NewMutator() *Mutator {
+	return &Mutator{}
+}
+
+// GVR returns the GroupVersionResource for this CRD.
+func (m *Mutator) GVR() schema.GroupVersionResource {
+	return gvr
+}
+
+// Operations returns list of operations handled by this mutator.
+func (m *Mutator) Operations() []admissionregistrationv1.OperationType {
+	return []admissionregistrationv1.OperationType{admissionregistrationv1.Create}
+}
+
+// MutatingWebhook returns the MutatingWebhook used for this CRD.
+func (m *Mutator) MutatingWebhook(clientConfig admissionregistrationv1.WebhookClientConfig) []admissionregistrationv1.MutatingWebhook {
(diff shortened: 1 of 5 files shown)


crobby (approved):
Looks pretty similar to the PRTB one, has tests and docs, lgtm

pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/mutator.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := patch.CreatePatch(request.Object.Raw, crtb, response); err != nil {
+		return nil, fmt.Errorf("failed to create patch: %w", err)
+	}
+	response.Allowed = true
+	return response, nil
+}
+
+// getSubject returns the binding subject using the following priority:
+// UserPrincipalName > UserName > GroupPrincipalName > GroupName.
+func getSubject(userPrincipalName, userName, groupPrincipalName, groupName string) string {
```
andreas-kupries, line 83: This seems to be the same as for PRTB. I yes, I would recommend to have a look into merging the duplicate code into a single function. Could be done in a followup PR to address tech debt.

pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/mutator.go
```diff
@@ -0,0 +1,107 @@
...
+	default:
+		return ""
+	}
+}
+
+// GenerateDeterministicName computes a deterministic resource name from the given prefix and
+// binding content. The name is: prefix + lowercase(base32(sha256(subject/roleTemplateName/clusterName))[:10]).
+// This produces names that are valid K8s resource names and ensures that two identical binding
+// requests yield the same name, causing the API server to reject the second with a 409 Conflict.
+func GenerateDeterministicName(prefix, subject, roleTemplateName, clusterName string) string {
```
andreas-kupries, line 102: This looks like duplicate code too.

andreas-kupries (approved):
LGTM, with some code duplication (I believe), to address

pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/mutator.go
pratikjagrut: I'll do it in another PR.

PR #1410, opened 2026-04-23
Reject operations on local users when the local auth provider is marked disabled

andreas-kupries:
## Issue: 

https://github.com/rancher/rancher/issues/46078

## Problem

See issue

## Solution

The existing validator for user operations is extended with additional which apply if and only if the local auth provider is marked (per setting) as disabled.

## CheckList
  
- [x] Test
  
- [x] Docs


pkg/resources/management.cattle.io/v3/users/validator.go
```diff
@@ -36,21 +41,24 @@ type admitter struct {
...
-// Validator validates tokens.
+// Validator validates users.
 type Validator struct {
 	admitter admitter
 }
 
 // NewValidator returns a new Validator instance.
-func NewValidator(userAttributeCache controllerv3.UserAttributeCache, sar authorizationv1.SubjectAccessReviewInterface, defaultResolver validation.AuthorizationRuleResolver, userCache controllerv3.UserCache) *Validator {
+func NewValidator(userAttributeCache controllerv3.UserAttributeCache, sar authorizationv1.SubjectAccessReviewInterface, defaultResolver validation.AuthorizationRuleResolver, userCache controllerv3.UserCache, settingCache controllerv3.SettingCache) *Validator {
+	logrus.Debugf("ZZZZZZ [user-validation] SETUP")
```
crobby: Is this debug log intentional?   "ZZZZZZ" seems rather unhelpful

crobby (approved):
Seems reasonable, has tests/docs, just a nit on the one debug log

pkg/resources/management.cattle.io/v3/users/validator.go
andreas-kupries: No. That seems to have slipped through into the commit.  Should have stayed local. Will be removed.

andreas-kupries: Done.

pkg/resources/management.cattle.io/v3/users/validator.go
```diff
@@ -173,6 +207,56 @@ func (a *admitter) checkUsernameUniqueness(username string) (*admissionv1.Admiss
...
+	}
+	// user is not local (default admin or system). pass to regular checks.
+	logrus.Debugf("[user-validation] %s: continue for non-local user %q", operation, user.Name)
+	return nil, nil
+}
+
+func (a *admitter) isLocalAuthProviderDisabled() (bool, error) {
+	setting, err := a.settingCache.Get(disabledLocalAuthProviderSetting)
+	if err != nil {
+		return false, err
```
andreas-kupries, line 245: Check for `is not found` and treat this as `no error, exists with default value: false`

andreas-kupries: done with commit de7f52221b95462e7f25fa9297828bb7bbbf41dc

pkg/resources/management.cattle.io/v3/users/validator.go
```diff
@@ -82,15 +89,34 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
 	}
 
-	if request.Operation == admissionv1.Create && newUser.Username != "" {
-		if resp, err := a.checkUsernameUniqueness(newUser.Username); err != nil || resp != nil {
-			return resp, err
+	if request.Operation == admissionv1.Create {
+		response, err := a.isRejectedLocalUser("create", newUser)
+		if err != nil {
+			return nil, err
+		}
```
bigkevmcd: nitpick: this could be shortened to... (same for a couple more cases).

```go
if response != nil || err != nil {
    return response, err
}
```

bigkevmcd (approved):

bigkevmcd (approved):

pkg/resources/management.cattle.io/v3/users/validator.go
andreas-kupries: changed with commit b278c9786e47a1bb16825faca4aa5093f10218f2
note: works only if response and error cannot be non-nil at the same time.
if they can then the new code has different semantics as response is returned non-nil, together with the error, instead of as nil

pkg/resources/management.cattle.io/v3/users/validator.go
```diff
@@ -173,6 +198,59 @@ func (a *admitter) checkUsernameUniqueness(username string) (*admissionv1.Admiss
 	return nil, nil
 }
 
+func (a *admitter) isRejectedLocalUser(operation string, user *v3.User) (*admissionv1.AdmissionResponse, error) {
+	// Check state of `local` auth provider. we are only relevant when it is disabled
```
pmatseykanets:
```suggestion
	// Check state of `local` auth provider. It's only relevant when it is disabled.
```

pkg/resources/management.cattle.io/v3/users/validator.go
```diff
@@ -173,6 +198,59 @@ func (a *admitter) checkUsernameUniqueness(username string) (*admissionv1.Admiss
...
+	disabled, err := a.isLocalAuthProviderDisabled()
```
pmatseykanets: The check ONLY applies when an external provider is enabled not just the enabled setting.
See https://github.com/rancher/rancher/pull/54829#discussion_r3220651883

pkg/resources/management.cattle.io/v3/users/validator.go
andreas-kupries: Done with commit 8ab353b14d46caccbe735b0383fc9171e1e4b3ae

andreas-kupries:
Rewrite is done as per the discussion tuesday.

bigkevmcd (approved):
Looks fine to me.

JonCrowther (approved):

PR #1414, opened 2026-04-27
Replace push:tags release trigger with validated workflow_dispatch

tomleb:
# Issue rancher/rancher#54790

Replace the `on: push: tags: v*` trigger with a `workflow_dispatch` workflow that validates the version before creating the tag and GitHub release.

The `validate` job:
- Checks version format (`v<major>.<minor>.<patch>[-prerelease]`)
- Confirms the version prefix matches the branch's allowed minor (read from `VERSION.md` on the default branch)
- Enforces sequential patch ordering within the minor line
- Confirms the tag does not already exist

The `release` job (runs after `validate`):
- Creates and pushes the annotated tag
- Creates the GitHub release with auto-generated notes

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,20 +1,106 @@
-name: release
+name: Release
 
 on:
-  push:
-    tags:
-      - v*
   workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release (e.g. v0.5.1)'
+        required: true
+        type: string
 
-permissions:
-  contents: write
+permissions: {}
 
 env:
   REGISTRY: docker.io
   REPO: rancher
 
 jobs:
+  validate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          persist-credentials: false
+
+      - name: Validate version format
+        env:
+          VERSION: ${{ inputs.version }}
+        run: |
+          if ! [[ "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$ ]]; then
+            echo "::error::Version '$VERSION' must match v<major>.<minor>.<patch>[-prerelease]"
+            exit 1
+          fi
+
+      - name: Resolve allowed minor from VERSION.md
+        id: minor
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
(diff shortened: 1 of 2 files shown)


crobby (commented):
Is there a chance that release/tag fails, but the image build/push still succeeds and results in a "dangling" image published without a tag?  Should we make sure the tag/release succeeds before doing the build/push?

Also...should we document somewhere that "git tag" should no longer be done externally?

Also...should we include some sort of VERSION.md file validation since we're relying on the correctness there (probably applies to all of these workflow PRs).

tomleb:
@crobby 

> Is there a chance that release/tag fails, but the image build/push still succeeds and results in a "dangling" image published without a tag? Should we make sure the tag/release succeeds before doing the build/push?

That's a great question / great catch. I think it makes sense to wait for the tag to be created. So I suppose for simplicity, I'd just do the publish after both validate and tag job.

> Also...should we document somewhere that "git tag" should no longer be done externally?

Good idea, I'm just not sure where. Should that be in the README? (And I'd say if we do, let's do that in all repos).

> Also...should we include some sort of VERSION.md file validation since we're relying on the correctness there (probably applies to all of these workflow PRs).

You are on fire, yes that's a good point. I'm wondering if we should have something slightly more complicated where we have a json file (or yaml, wtv) that's easy to validate which then generates the VERSION.md :thinking: It seems easier to validate a json file with a schema than a markdown table...

crobby:
> @crobby
> 
> > Is there a chance that release/tag fails, but the image build/push still succeeds and results in a "dangling" image published without a tag? Should we make sure the tag/release succeeds before doing the build/push?
> 
> That's a great question / great catch. I think it makes sense to wait for the tag to be created. So I suppose for simplicity, I'd just do the publish after both validate and tag job.
> 
> > Also...should we document somewhere that "git tag" should no longer be done externally?
> 
> Good idea, I'm just not sure where. Should that be in the README? (And I'd say if we do, let's do that in all repos).
> 
> > Also...should we include some sort of VERSION.md file validation since we're relying on the correctness there (probably applies to all of these workflow PRs).
> 
> You are on fire, yes that's a good point. I'm wondering if we should have something slightly more complicated where we have a json file (or yaml, wtv) that's easy to validate which then generates the VERSION.md 🤔 It seems easier to validate a json file with a schema than a markdown table...

I'm fine with the "don't do git tag on your own" note being in README for all repos.

I do agree that a markdown table is maybe not the easiest thing to validate.  I'd be happy with anything that can be reliably produced and parsed.

tomleb:
@crobby I'm thinking of punting the move to VERSION.md from markdown to some other format to later. I think it's a good think to have validation but I think it's not necessary yet (famous last words). Partly out of laziness, partly because I want the new workflow working soon.

I have added a section for releases.

crobby (approved):
I think this looks good.

ericpromislow (approved):
LGTM

PR #1416, opened 2026-04-27
Add support for image pull secrets

HarrisonWAffel:
## Problem
The webhook chart does not accept image pull secrets, preventing it from being deployed into environments which only have access to an authenticated image registry (e.g. secure air-gap environments) and cannot configure registry credentials at the container runtime level.

## Solution
Add `global.cattle.imagePullSecrets` to `values.yaml` and use it when deploying the webhook. A template helper was added to allow users to pass either a simple list of strings, or complete object references (e.g. `name: <secret>`). 

## CheckList
- [x] Test
Updated the helm unit test to account for this new field and the potential ways values can be provided

- [x] Docs
There doesn't seem to be a chart README, so no documentation updates.

### DIFF
--- a/charts/rancher-webhook/values.yaml
+++ b/charts/rancher-webhook/values.yaml
@@ -6,6 +6,9 @@ image:
 global:
   cattle:
     systemDefaultRegistry: ""
+    # imagePullSecrets:
+    #   - name: secret1
+    imagePullSecrets: []
   hostNetwork: false
 
 mcm:
--- a/charts/rancher-webhook/values.yaml
+++ b/charts/rancher-webhook/values.yaml
@@ -6,8 +6,6 @@ image:
 global:
   cattle:
     systemDefaultRegistry: ""
-    # imagePullSecrets:
-    #   - name: secret1
     imagePullSecrets: []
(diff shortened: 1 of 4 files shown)


charts/rancher-webhook/values.yaml
```diff
@@ -6,6 +6,9 @@ image:
 global:
   cattle:
     systemDefaultRegistry: ""
+    # imagePullSecrets:
```
crobby: Are the commented out lines helpful?

crobby (approved):
Looks pretty straightforward, just a non-blocking question on the commented-out values.

charts/rancher-webhook/values.yaml
HarrisonWAffel: I thought it could be useful as a reference, it's also a pattern I've seen in other charts since the expected value of image pull secrets isn't always the same (string array vs array of local object references). But since Rancher automatically installs this chart I can drop it, no problem

thatmidwesterncoder (approved):

jakefhyde (approved):

crobby (approved):

PR #1423, opened 2026-04-28
[v0.10] Fix branching

tomleb:
CI runs against rancher main, not against rancher 2.14


crobby (approved):
Makes sense

ericpromislow (approved):
Looks correct

PR #1428, opened 2026-04-30
Split release workflow into Cut release and On release

tomleb:
# Issue rancher/rancher#54790

## Summary

Splits the release workflow into two `workflow_dispatch` workflows:

- **`.github/workflows/cut-release.yaml`** ("Cut release") — entry point. Validates the version against `VERSION.md`, creates and pushes the annotated tag, then dispatches the On release workflow at the new tag ref.
- **`.github/workflows/release.yaml`** ("On release") — `workflow_dispatch`. Builds, creates the GitHub release, and publishes the image. A guard job rejects any dispatch on a non-tag ref.

## Why

The publish-image action's cosign provenance check verifies a SAN of the form:

```
^https://github.com/<org>/webhook/.github/workflows/release.(yml|yaml)@refs/tags/<tag>$
```

The previous single workflow ran the whole pipeline from `workflow_dispatch` on a branch, producing a SAN of `refs/heads/main` instead of `refs/tags/<tag>` — verification failed:

```
ERROR: Failed to generate slsav1 provenance. Check whether the image is present in the Prime registry.
failed to verify "..rancher-webhook:v0.11.0-rc.3@sha256:...": no matching attestations:
expected SAN value to match regex "^https://github.com/.../webhook/.github/workflows/release.(yml|yaml)@refs/tags/v0.11.0-rc.3$",
got "https://github.com/.../webhook/.github/workflows/release.yaml@refs/heads/main"
```

So the publish step has to live in `release.yaml` and run from `refs/tags/v*`.

A `push: tags` trigger does not work here: GitHub blocks workflow runs triggered by `GITHUB_TOKEN` from starting other runs via push events, so the tag pushed by Cut release does not fire On release. `workflow_dispatch` is exempt from that restriction, which is why the second workflow uses `workflow_dispatch` and Cut release calls \`gh workflow run release.yaml --ref <tag>\`.

### DIFF
--- a/.github/workflows/cut-release.yaml
+++ b/.github/workflows/cut-release.yaml
@@ -1,0 +1,118 @@
+name: Cut release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release (e.g. v0.5.1)'
+        required: true
+        type: string
+
+permissions: {}
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+          persist-credentials: false
+
+      - name: Validate version format
+        env:
+          VERSION: ${{ inputs.version }}
+        run: |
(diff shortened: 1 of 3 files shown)


crobby:
If the release workflow does get triggered, I think this is good.  I'm not sure that it will though based on:  https://docs.github.com/en/actions/concepts/security/github_token#when-github_token-triggers-workflow-runs

crobby (approved):

PR #1445, opened 2026-05-08
Improve inconsistency in RTB validation webhooks

JonCrowther:
## Problem
In both the CRTB and PRTB webhooks, in the `validateCreateFields` functions we check if the RoleTemplate referenced exists. If it doesn't exist we don't allow the RTB to be created.

However, later on in the validation code, we fetch the RoleTemplate so we can check the rules. If the RoleTemplate isn't found, it returns with an "Allow" response. It's not possible to get to that point since the previous check has to pass. This can lead to false positives when scanning and is generally a little confusing.

## Solution
I removed the "Allow" response, instead just returning an error.

I also improved the PRTB check for RoleTemplate by returning a `field.Invalid` response instead of an error. This way it matches the CRTB validation.

## CheckList
  
- [ ] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/validator.go
+++ b/pkg/resources/management.cattle.io/v3/clusterroletemplatebinding/validator.go
@@ -127,9 +127,6 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 
 	roleTemplate, err := a.roleTemplateResolver.RoleTemplateCache().Get(crtb.RoleTemplateName)
 	if err != nil {
-		if apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


andreas-kupries (approved):

pmatseykanets (approved):

crobby (approved):
Seems good, should probably link the issue at some point.
I should have checked, but does this wind-up making a change that should be in the webhook docs?

JonCrowther:
> Seems good, should probably link the issue at some point. I should have checked, but does this wind-up making a change that should be in the webhook docs?

This does not change any behaviour since the check I changed isn't actually reachable. I linked a related issue

PR #1451, opened 2026-05-11
Make LDAP/AD authconfig UserIDAttribute and GroupIDAttribute fields immutable

pmatseykanets:
## Issue: https://github.com/rancher/rancher/issues/55642

Complimentary to https://github.com/rancher/rancher/pull/54912

## Problem

With [rancher/rancher#54912](https://github.com/rancher/rancher/pull/54912) Rancher now has `UserIDAttribute` and `GroupIDAttribute` fields in LDAP/AD authconfigs, letting admins choose a stable LDAP attribute (e.g. `sAMAccountName`) as the principal identifier instead of DN.

These fields must be immutable after the provider is enabled — changing them would silently break all RBAC bindings (CRTBs, PRTBs, GRBs) that reference the old principal name. Rancher's `testAndApply` enforces this for the UI/API flow, but `kubectl apply`, Terraform, and direct API mutations bypass it.

## Solution

The webhook now enforces two things for `UserIDAttribute` and `GroupIDAttribute`:

- Format validation: both fields are validated with `IsValidLdapAttr` (RFC 4512), same as every other LDAP attribute field.
- Immutability on enabled providers: when both the old and new config have `Enabled == true`, any change to either field is rejected with a `field.Invalid` error. To change these fields, the admin must disable the provider first, apply the change, then re-enable — which is the supported reconfiguration path.

## CheckList
  
- [x] Test
  
- [x] Docs


go.mod
```diff
@@ -4,6 +4,8 @@ go 1.25.0
 
 toolchain go1.25.9
 
+replace github.com/rancher/rancher/pkg/apis => github.com/pmatseykanets/rancher/pkg/apis v0.0.0-20260504145013-06d2cbb05095
```
pmatseykanets, line 7: This will be removed once the rancher PR is merged.

crobby (approved):
Seems to do what it aims to do.  lgtm

PR #1465, opened 2026-05-13
[main] Add admission validation for WebhookDeploymentCustomization

crobby:
Issue: https://github.com/rancher/rancher/issues/54090

## Problem

Invalid `WebhookDeploymentCustomization` values (zero replicas, malformed toleration keys, conflicting PDB fields) pass through the API and only fail at Helm install time on the downstream cluster, making errors difficult to diagnose.

## Solution

Add admission validators for `WebhookDeploymentCustomization` 

- `replicaCount` must be >= 1
- Toleration keys validated against k8s label name rules
- Affinity label selectors validated via apimachinery
- PDB: must be non-negative int or 0-100% string; `minAvailable` and `maxUnavailable` cannot both be non-zero

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -245,6 +245,22 @@ Both `minAvailable` and `maxUnavailable` must be a string which represents a non
 ^([0-9]|[1-9][0-9]|100)%$
 ```
 
+##### Feature: Webhook Deployment Customization
+
+The `WebhookDeploymentCustomization` field configures the rancher-webhook deployment on downstream clusters. The following sub-fields are validated:
+
+- `replicaCount`: If set, must be at least 1.
+- `appendTolerations`: Toleration keys are validated against the upstream apimachinery label name regex:
+  ```regex
+  ([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]
+  ```
+- `overrideAffinity`: Node affinity `nodeSelectorTerms` are validated via label name validation. Pod affinity and pod anti-affinity are validated via label selectors using the [apimachinery label selector validation](https://github.com/kubernetes/apimachinery/blob/02a41040d88da08de6765573ae2b1a51f424e1ca/pkg/apis/meta/v1/validation/validation.go#L56).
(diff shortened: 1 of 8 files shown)


ericpromislow (approved):
LGTM

pkg/resources/management.cattle.io/v3/cluster/validator.go
```diff
@@ -130,6 +131,18 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
 		return response, err
 	}
 
+	if wdc := newCluster.Spec.WebhookDeploymentCustomization; wdc != nil {
```
rohitsuse: Should this be extracted into a new function matching the pattern the validators above use (e.g. a.validateXXX() returning *admissionv1.AdmissionResponse)? The other validate functions check !response.Allowed to decide whether to bail out, but this block doesn't — and it never sets response.Allowed = false like they do.

docs.md
```diff
@@ -245,6 +245,22 @@ Both `minAvailable` and `maxUnavailable` must be a string which represents a non
 ^([0-9]|[1-9][0-9]|100)%$
 ```
 
+##### Feature: Webhook Deployment Customization
```
rohitsuse, line 248: https://github.com/rancher/webhook/pull/1465/files#diff-6a042219e4779f0e778edb839c177737f9d49be99d5032a644a5ae7cded83876R248
2. https://github.com/rancher/webhook/pull/1465/files#diff-6a042219e4779f0e778edb839c177737f9d49be99d5032a644a5ae7cded83876R828
3. https://github.com/rancher/webhook/pull/1465/files#diff-c7ecb1ad9303708619fe578ec48d06d532b659c85089963406ce7c703c2ec8ccR89

1 2 3 have the same content. Was this purposeful ?

crobby: Yes, this is intentional, we documents both management.cattle.io/v3.Cluster and provisioning.cattle.io/v1.Cluster separately (same validation applies to both API groups), and Cluster.md is the per-resource doc.

pkg/resources/management.cattle.io/v3/cluster/validator.go
crobby: Good point, I should refactor this a bit.

PR #1502, opened 2026-05-26
[main v0.12] Rework webhook TLS and RBAC: replace dynamiclistener, bake WebhookConfigurations

crobby:
Issue:  https://github.com/rancher/rancher-security/issues/1243

## Problem

The webhook currently uses dynamiclistener for TLS certificate management and ships with a `cluster-admin` ClusterRoleBinding.

## Solution

- Replace dynamiclistener + secretHandler with file-based TLS
- Bake WebhookConfigurations and needacert annotation into the chart (instead of runtime generation)
- Narrow RBAC from cluster-admin to an explicit ClusterRole with only required permissions
- Start TLS server in a goroutine so informers can start independently
- Make securityContext unconditional (only NET_BIND_SERVICE capability is conditional)
- Populate caBundle from TLS secret after helm upgrade

## Related PRs

- rancher/rancher#55037 — Rancher-side webhook config reworking

## Testing

- [x] Webhook starts and serves TLS with file-based certs
- [x] caBundle populated correctly after helm upgrade
- [x] RBAC scoped correctly — webhook can perform all required operations


charts/rancher-webhook/templates/rbac.yaml
```diff
@@ -1,12 +1,104 @@
...
+  resources: [roles]
+  verbs: [update]
+- apiGroups: ["rbac.authorization.k8s.io"]
+  resources: [rolebindings, clusterrolebindings]
+  verbs: [create]
+# The fleetworkspace mutator creates a per-workspace ClusterRole and binds it (plus the
+# fleet-admin ClusterRole) to the workspace creator. Those grants include permissions the
+# webhook ServiceAccount does not itself hold (fleet.cattle.io/*), so Kubernetes RBAC
+# escalation prevention requires escalate (to create the ClusterRole) and bind (to reference
+# it from a [Cluster]RoleBinding).
```
tomleb: Is `bind` still required? https://github.com/rancher/webhook/pull/1590 got rid of the mutator

charts/rancher-webhook/values.yaml
```diff
@@ -12,6 +12,11 @@ global:
 mcm:
   enabled: true
 
+# caBundle is populated at runtime by needacert (wrangler/pkg/needacert)
+# once the cattle-webhook-tls secret has been generated. Override only
+# when running this chart standalone without needacert.
+caBundle: ""
```
tomleb, line 19: Wondering if we should use this for the integration test. And make the need-a-cert conditional. That way we don't need the "hack" in the integration test script.

charts/rancher-webhook/templates/webhook.yaml
```diff
@@ -2,8 +2,1167 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: rancher.cattle.io
+webhooks:
```
tomleb, line 5: We should add the ability to generate this with `go generate`.

I had a POC that was working in https://github.com/tomleb/rancher-webhook/blob/webhookconfig-in-helmchart/:
- The script to do the codegen: https://github.com/tomleb/rancher-webhook/blob/webhookconfig-in-helmchart/scripts/update-chart-manifests
- The CLI that builds the *WebhookConfiguration to a file: https://github.com/tomleb/rancher-webhook/blob/webhookconfig-in-helmchart/main.go#L40-L50

Thoughts?

charts/rancher-webhook/templates/rbac.yaml
```diff
@@ -1,12 +1,104 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: ClusterRole
```
tomleb, line 2: Does the unit test / integration test make use of those RBAC resources? The CI should give an error if a webhook is trying to read / create / whatever something that it won't have access to in production.

tomleb (commented):
A couple of comments :)

charts/rancher-webhook/templates/rbac.yaml
crobby: I will double check....my changes predate that removal.

charts/rancher-webhook/values.yaml
crobby: That seems like a pretty reasonable thing to do.

charts/rancher-webhook/templates/rbac.yaml
crobby: Nice....mutator removal should let us remove this ugly bit of rbac.

charts/rancher-webhook/templates/webhook.yaml
crobby: Interesting thought.  I'll take a look.

charts/rancher-webhook/templates/rbac.yaml
crobby: I think our coverage in the integration tests is likely just partial.  I'll noodle on a creative way to get some more thorough tests in there.

charts/rancher-webhook/values.yaml
crobby: Reworked to make needacert conditional.

tomleb: That works, though my first thought is doesn't caBundle being set implies already needacert=false? Basically, do we need needacert.enabled at all? If caBundle is empty -> then needacert annotations.

tomleb: Just nits honestly

charts/rancher-webhook/templates/rbac.yaml
crobby: @tomleb I've added some new tests that will do testing based on our actual ClusterRole as well as the Validators that are running.  Should be covering "everything" now.

crobby: Ah...and the new tests actually caught a couple of omitted rbac entries!

tomleb: Hum... So I see you have added a test that checks for resources in MutatingWebhookConfiguration and ValidatingWebhookConfiguration. That's not quite what I meant but I suppose that can still be useful.. 

What I meant was more permissions like: webhook has some code path that creates various resources (eg: SAR, Secrets, etc). SARs will certainly not be in the MutatingWebhookConfiguration/ValidatingWebhookConfiguration, won't be picked up by the test. But we still need to ensure this runs fine when webhook does the create. So unit test will probably run the code with a "fake" k8s client -> no RBAC check -> we don't really know if that will succeed in production.

Does that make more sense? Tbh not sure if we really need to test that... But seemed like the right thing to do.

crobby: Yeah, I kind of like the test for all of the resources in Mutating/Validating webhook configuration.  That should prevent the case of adding a resource, but forgetting to add the rbac.

I've added a check specifically for SARs as well.

charts/rancher-webhook/templates/webhook.yaml
crobby: I think this is a good idea.  I will include it in another PR.

ericpromislow (approved):
Looks ok to me

charts/rancher-webhook/templates/webhook.yaml
crobby: Issue for the follow-up to this:  https://github.com/rancher/rancher/issues/57556

charts/rancher-webhook/values.yaml
crobby: needacert now just depends on caBundle being empty (or not)

PR #1507, opened 2026-06-02
[main] Allow update "env" setting with another "env" setting

joshmeranda:
## Problem

Some rancher setting like `agent-image` are likely to be changed by users via helm or directly with environment variables. Since the current validations prevent all updates to "env" sourced Settings, this is not possible without manually deleting those settings. This breaks the rancher/rancher provisioning tests and changes existing behavior.

## Solution

Allow changes to "env" sourced Settings when the new setting is also "env" sourced.

## CheckList
  
- [x] Test
  
- [x] Docs


crobby (approved):
Should probably link to the existing issue that this is related to or create a new issue to track this.  lgtm though

PR #1590, opened 2026-06-16
Align fleetworkspace webhook handling with controller flow

thardeck:
The `FleetWorkspace` mutating admission handler is removed from the webhook server registration, and the associated mutator implementation is deleted. This keeps webhook mutation behavior focused on request shaping and avoids creating workspace side-effect resources in the webhook process.

Webhook configuration reconciliation continues to rebuild the mutating webhook list from registered handlers at startup. A regression test is added to verify configuration updates replace webhook entries rather than accumulating stale entries.


tomleb (commented):
docs need update : https://github.com/thardeck/webhook/blob/791edb62d05e8ee31edcab229f9f987a0a91defd/pkg/resources/management.cattle.io/v3/fleetworkspace/FleetWorkspace.md

p-se (approved):
lgtm

crobby (commented):
I'm good with this once the docs are updated (update the md file and then go generate)

tomleb (approved):

PR #1600, opened 2026-06-18
Fix User validator

pmatseykanets:
## Problem

A nil labels.Selector passed to the cache List method causes a nil pointer dereference (#1410). The webhook panics on any admission request that reaches the local auth provider check.

## Solution

The selector is replaced with labels.Everything(), which lists all AuthConfig objects as intended.

## CheckList
  
- [ ] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/management.cattle.io/v3/users/validator.go
+++ b/pkg/resources/management.cattle.io/v3/users/validator.go
@@ -249,7 +249,7 @@ func (a *admitter) hiddenLocalAuthProvider() (bool, error) {
 	}
 
 	// Check for active external auth provider, query etcd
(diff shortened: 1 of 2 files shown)


tomleb (approved):
Should this have a unit test at least testing this functionality? Approving anyway.

pmatseykanets:
> Should this have a unit test at least testing this functionality?

You're right. Updating tests to explicitly check for `labels.Everything()` in mocked calls.

JonCrowther (approved):

JonCrowther (approved):

andreas-kupries (commented):
Yes. LGTM, and thank you. Stupid me.

PR #1601, opened 2026-06-18
Fix User validator for hide-local-auth-provider feature

pmatseykanets:
## Problem

There were a few issues identified since #1410:

- When the hide-local-auth-provider feature was active, the local user check ran before the manage-users bypass, so users with the manage-users verb could not delete or update local users.
- The AuthConfig informer was registered after the management factory's sync barrier ran, leaving a startup window where the informer cache was empty. During that window, the local user check saw no active external providers and skipped the guard silently. 
- The feature name in docs didn't match the code.

## Solution

- The manage-users check is now done before the local user check for update and delete operations, so the bypass applies to both. Create operations keep the check unconditional — the feature should always block creating new local users regardless of privilege.
- The informer is now pre-registered before the sync call.
- The feature name in the docs is updated to match the code.

## CheckList
  
- [ ] Test
  
- [ ] Docs

### DIFF
--- a/pkg/clients/clients.go
+++ b/pkg/clients/clients.go
@@ -54,6 +54,11 @@ func New(ctx context.Context, rest *rest.Config, mcmEnabled bool) (*Clients, err
 		return nil, err
 	}
 
+	// Pre-register informers used by validators before Start so they are included
+	// in the initial cache sync barrier. Without this, lazily-registered informers
+	// may not be synced when the HTTP server begins serving admission requests.
+	_ = mgmt.Management().V3().AuthConfig().Cache()
+
 	if err = mgmt.Start(ctx, 5); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


andreas-kupries (approved):
Looks very good to me.

bigkevmcd (approved):

JonCrowther (approved):

pkg/clients/clients.go
```diff
@@ -54,6 +54,11 @@ func New(ctx context.Context, rest *rest.Config, mcmEnabled bool) (*Clients, err
 		return nil, err
 	}
 
+	// Pre-register informers used by validators before Start so they are included
+	// in the initial cache sync barrier. Without this, lazily-registered informers
+	// may not be synced when the HTTP server begins serving admission requests.
+	_ = mgmt.Management().V3().AuthConfig().Cache()
```
tomleb, line 60: I'm surprised this isn't a problem with other caches 🤔 might be worth looking into, idk.

tomleb (approved):

pkg/clients/clients.go
pmatseykanets: It probably is.

PR #1656, opened 2026-07-09
sync-deps: never recommend dependency downgrades

crobby:
## Solution
Use golang's semver to compare the versions of deps.  Only recommend a bump if the version used Rancher is greater than the current version in webhook


tomleb (approved):
Nice

PR #1729, opened 2026-07-28
Set the creatorId annotation on dry-run provisioning Cluster creates

pmatseykanets:
## Problem

The mutating webhook for provisioning Clusters returned early on every dry-run request, before setting the creatorId annotation.
The validating webhook requires that annotation on dry-run creates, so every server-side dry-run create was denied while the identical real create succeeded.
For PSACT-configured clusters the validating webhook also checks the admission-config secret the mutator writes, which fails the same way on dry-run since the write never happens.

## Solution

The mutator's early return is replaced with a guard around only the PSACT secret handling, the one stage with a real side effect.
On dry-run, the object parse, the creatorId annotation, the dynamic-schema revert, and the JSON patch generation now run as they do on a real request, so the preview patch matches the eventual apply.
The validating webhook skips only its checks against that mutator-produced state (the secret and the cluster fields derived from it) on dry-run; its pure input checks — the kubernetes version floor and the template's existence — still run, so invalid input is denied in the preview exactly as it would be on a real request.
The patch changes only the in-flight request object, which a dry-run never persists, so the webhook's SideEffects: NoneOnDryRun registration remains accurate.

## CheckList
  
- [x] Test
  
- [x] Docs

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -829,6 +829,15 @@ If the action is an update, and the old cluster had a `nil` `.spec.rkeConfig`, a
 
 Prevent the creation of objects if the secret specified in `.spec.rkeConfig.etcd.s3.cloudCredentialName` does not exist.
 
+##### Pod Security Admission Configuration Template
+
+When `spec.defaultPodSecurityAdmissionConfigurationTemplateName` is set, the referenced
+`PodSecurityAdmissionConfigurationTemplate` must exist and the cluster's Kubernetes version must be 1.23 or above.
+The admission-configuration secret managed by the mutator, and the cluster fields derived from it, are also
+validated — except on server-side dry-run requests, where the mutator does not perform that secret management (see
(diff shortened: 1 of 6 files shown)


crobby (commented):
Anything here that might require a change/addition to the Cluster.md file (and the generated docs)?

pmatseykanets:
> Anything here that might require a change/addition to the Cluster.md file (and the generated docs)?

Thanks for nudging 😄 Though we never documented dry-run behavior before I think we should.
This PR changes validation/mutation contract here (even though to make it correct / align) it's worth documenting, which is done in c96d527

pmatseykanets:
@crobby Any other concerns?

jiaqiluo (approved):
LGTM! 
Thank you for adding the missing docs.

crobby (approved):

crobby:
> @crobby Any other concerns?

Nope, lgtm

PR #1741, opened 2026-08-04
Pin read-vault-secrets to a commit that SHA-pins hashicorp/vault-action

rohitsuse:
The `v3` tag of `rancher-eio/read-vault-secrets` points at `0da85151` (2025-08-12), whose `action.yaml` calls `hashicorp/vault-action@v3` by tag. `7282bf97` (their current `main`) SHA-pins it instead.

In `rancher/webhook-security`, which syncs these workflows from here, the tag form is rejected at job setup and the RCS release job dies before its first step:

```
Error: The action hashicorp/vault-action@v3 is not allowed in rancher/webhook-security
because all actions must be from a repository owned by your enterprise ...
```

The trailing `# v3` comment is dropped along with the bump. It is what made Renovate resolve the tag and pin us back to `0da85151` in the first place, and it is inaccurate for a commit seven months past that tag.

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -20,7 +20,7 @@ jobs:
     # The FOSSA token is shared between all repos in Rancher's GH org. It can be
     # used directly and there is no need to request specific access to EIO.
     - name: Read FOSSA token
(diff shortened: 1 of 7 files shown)


ericpromislow (approved):
Verified this pulls in a more recent instance of `rancher-eio/read-vault-secrets` as specified

PR #1751, opened 2026-08-10
[release/v0.9] Pin read-vault-secrets to a commit that SHA-pins hashicorp/vault-action

rohitsuse:
Backport of #1741: SHA-pin hashicorp/vault-action via read-vault-secrets.


ericpromislow (approved):
Straightforward

PR #1765, opened 2026-08-14
[main] bypass read only settings for rancher service account

joshmeranda:
## Problem

Some settings, like `cacerts`, are considered read only and all updates to them are rejected. When users want to update their rancher deployment to use different cacerts, webhook will reject the update requests sent by rancher during startup and cause rancher to fail.

## Solution

Allow updates updates coming directly from the Rancher serviceaccount.

## CheckList
- [x] Test


crobby (changes requested):
The code itself looks good to me.  We should probably update the validator docs to include this change in behavior.

tomleb (approved):

crobby (approved):

crobby:
If the CI fails again, it probably just needs a rebase to pick-up some CI fixes that were added recently.

PR #1789, opened 2026-08-18
[main] provisioning cluster: remove carriage returns from agentEnvVars values

vardhaman22:
## Issue: 

https://github.com/rancher/rancher/issues/50213

## Problem

When configuring a provisioning cluster's agent environment variables from a file on UI, values containing DOS/Windows line endings (\r\n) retain the trailing carriage return.
These carriage returns are propagated to the generated node registration command, resulting in malformed environment variable values
## Solution

Update the provisioning cluster mutating webhook to normalize agentEnvVars values during Create and Update operations.

The normalization removes trailing carriage return (\r) and newline (\n) characters while preserving any newline characters within the value.
## CheckList
  
- [x] Test
  
- [ ] Docs

### DIFF
--- a/pkg/resources/provisioning.cattle.io/v1/cluster/Cluster.md
+++ b/pkg/resources/provisioning.cattle.io/v1/cluster/Cluster.md
@@ -154,6 +154,10 @@ kube-apiserver argument that mount it; when the field is cleared, the secret is 
 dropped. On server-side dry-run requests the secret is neither written nor deleted, and the derived fields are not
 set or dropped — the corresponding validation of that state is skipped on dry-run as well (see Validation Checks).
 
+#### Agent Environment Variable Normalization
+
+Trailing carriage return (`\r`) and newline (`\n`) characters are stripped from the `Value` of each entry in `spec.agentEnvVars`.
+
 ### On Update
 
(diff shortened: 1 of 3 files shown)


crobby (approved):
Seems good.  Is there any chance of \r or \n in the middle of the value that would need stripping or is this expected to be sufficient?

Also, is this change worthy of being noted in the mutator's doc?

HarrisonWAffel (changes requested):
Agree with Chad, we'll need to update the documentation as well. IIUC this was preventing users from registering nodes entirely, so I don't know if there are any migration / existing cluster concerns here, but something to think about

vardhaman22:
> Seems good. Is there any chance of \r or \n in the middle of the value that would need stripping or is this expected to be sufficient?
> 
kept the behavior same except the line endings since the issue was reported explicitly for the line endings.

HarrisonWAffel (approved):

crobby (approved):

vardhaman22:
closing this based on the offline discussion.
It can cause unnecessary redeployment of cluster agent on existing clusters. 
The UI fix should be enough to solve the issue.

PR #1797, opened 2026-08-20
[backport v0.11 - rancher v2.15] reject user operations on rancher-managed resource quota and limit range resources

andreas-kupries:
## Issue: 

https://github.com/rancher/rancher/issues/56773

Companion PR r/rancher https://github.com/rancher/rancher/pull/56801

## Problem

The NRQR (Rancher-managed namespace `ResourceQuota`, identified by label `resourcequota.management.cattle.io/default-resource-quota: true`) was enforced only at namespace creation time. Any user with RBAC access to `resourcequotas` could subsequently edit the object directly, fully bypassing project-level enforcement.

The same is true for the LimitRange resource managed by Rancher.

## Solution

Added validating webhooks for  `ResourceQuota` and `LimitRange` objects.
The companion PR modifies Rancher to send its requests using the webhook bypass.
This means that the new validators see only user initiated requests, and only have to reject operations involving the marker label. IOW

  - Reject attempts to create resources bearing the marker label
  - Reject attempts to delete resources bearing the marker label
  - Reject attempts to change resources bearing the marker label
    - This includes changes attempting to strip the marker (demotion to unmanaged)
  - Reject attempts to add a marker to to un-marked resources (i.e. prevent promotion to managed)

PR #1798, opened 2026-08-20
[backport v0.10 - rancher v2.14] reject user operations on rancher-managed resource quota and limit range resources - #1797#1797

andreas-kupries:
## Issue: 

https://github.com/rancher/rancher/issues/56774

Companion PR r/rancher https://github.com/rancher/rancher/pull/56804

## Problem

The NRQR (Rancher-managed namespace `ResourceQuota`, identified by label `resourcequota.management.cattle.io/default-resource-quota: true`) was enforced only at namespace creation time. Any user with RBAC access to `resourcequotas` could subsequently edit the object directly, fully bypassing project-level enforcement.

The same is true for the LimitRange resource managed by Rancher.

## Solution

Added validating webhooks for  `ResourceQuota` and `LimitRange` objects.
The companion PR modifies Rancher to send its requests using the webhook bypass.
This means that the new validators see only user initiated requests, and only have to reject operations involving the marker label. IOW

  - Reject attempts to create resources bearing the marker label
  - Reject attempts to delete resources bearing the marker label
  - Reject attempts to change resources bearing the marker label
    - This includes changes attempting to strip the marker (demotion to unmanaged)
  - Reject attempts to add a marker to to un-marked resources (i.e. prevent promotion to managed)


bigkevmcd (approved):

snasovich (approved):
Approved based on it being the same as https://github.com/rancher/webhook/pull/1797

rohitsuse (approved):
LGTM from Frameworks persepective.

PR #1829, opened 2026-09-01
[main] Fix invalid semver for untagged dev builds

crobby:
## Summary
- `scripts/version` builds the dev/untagged version string as `0.0.0-<short-commit>`. When the short commit hash is all-digits with a leading zero (e.g. `0894256`), that prerelease identifier is invalid per semver (numeric identifiers can't have leading zeros), which fails Helm chart validation in `make package-helm`.
- Prefix the commit hash with `g` (matches `git describe` convention) so the identifier is always alphanumeric, sidestepping the numeric leading-zero rule.

Seen failing in CI for #1824: `Error: validation: chart.metadata.version "0.0.0-0894256" is invalid`.

## Test plan
- [x] CI passes package-helm / build steps

### DIFF
--- a/scripts/version
+++ b/scripts/version
@@ -11,7 +11,7 @@ GIT_TAG=${GIT_TAG:-$(git tag -l --contains HEAD | head -n 1)}
 if [[ -z "$DIRTY" && -n "$GIT_TAG" ]]; then
     VERSION=$GIT_TAG
 else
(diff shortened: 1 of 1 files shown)


mallardduck (approved):

tomleb (approved):
Ohhh, we may have the same issue with rancher/rancher then 🤔

PR #1848, opened 2026-09-10
fix: allow deletion of local namespace on downstream clusters

susesamu:
Issue: https://github.com/rancher/rancher/issues/54831


crobby (changes requested):
Actual changes look fine, but you should also update the Namespace.md file to include the new validation information.  After doing that, you can run a `go generate` to update the main webhook docs.

crobby (approved):
Thanks for the docs update

PR #1880, opened 2026-09-15
allow k8s api-server to update the status of managed resource quota resources

andreas-kupries:
## Issue: 

Ref https://github.com/rancher/rancher/issues/56773#issuecomment-5674364336

## Problem

Status update requests from the `system:apiserver` are rejected.

## Solution

The  `system:apiserver` is added to the whitelist.

## CheckList
  
- [ ] Test
  
- [ ] Docs


pkg/resources/core/v1/resourcequota/validator.go
```diff
@@ -113,11 +115,18 @@ func (a *admitter) Admit(request *admission.Request) (*admissionv1.AdmissionResp
...
-			// the status of quota resources, managed or not.
+			// Both kubernetes quota controller and api server are
+			// allowed to update the status of quota resources,
+			// managed or not.
 			if request.UserInfo.Username == kubernetesQuotaController {
 				return admission.ResponseAllowed(), nil
 			}
+			if request.UserInfo.Username == kubernetesAPIServer {
+				return admission.ResponseAllowed(), nil
+			}
```
snasovich: [Optional] Nit - combine 2 `if`s in one - or better yet introduce list of `statusUpdateAllowListUsers` or something along those lines so we can easily add to it.

snasovich (approved):
LGTM, one optional nit

pkg/resources/core/v1/resourcequota/validator.go
andreas-kupries: Fixed.

snasovich (approved):

gehrkefc (approved):

PR #1894, opened 2026-09-16
Validate the Active Directory bind mechanism

pmatseykanets:
## Issue: https://github.com/rancher/rancher/issues/52891 

Complementary to https://github.com/rancher/rancher/pull/56935

## Problem

The Active Directory authconfig accepted any value for bindMechanism, including ntlm without TLS. Rancher checks this on its own testAndApply path, which leaves a config written directly through the API unvalidated.

## Solution

An empty value, simple and ntlm are now accepted, and ntlm additionally requires tls or starttls. The bind settings are read from the raw object so the validation does not depend on the Rancher types carrying them.

## CheckList
  
- [ ] Test
  
- [ ] Docs


crobby (approved):
lg from frameworks perspective

prachidamle (approved):

